Issue #19, opened 2018-06-08
Trigger inputs


VaibhavPage:
**Is your feature request related to a problem? Please describe.**
Currently sensor lacks the way of passing payload obtained from signal to triggers.

**Describe the solution you'd like**

magaldima:
I've added a `LatestEvent` to the `NodeStatus` that contains the most recent `Event` for a certain signal node. This is done in #36 .

magaldima:
We still should add a field to the sensor spec to specify `arguments` or `parameters` that we are passing to the trigger. By default, I think we should be passing the latest Event which is now stored in the sensor status. These args or params should be in addition to the latestEvent. This may also involve creating a `Triggerer` interface similarly to the `Signaler` interface which takes in an `Event` and other params/args.

vicaire:
Will the whole payload necessarily be passed to the workflow? Or will this allow the user to extract a field from the payload? 

For instance, let us assume that the event is that a file was added to GCS (which triggers either a webhook notification, or a new message in a pub/sub queue). 

Could the solution let the user specify that the name of the GCS object should be extracted from the pub/sub message json content, and then passed to the Argo workflow as the value for the parameter "OBJECT_URI"?

magaldima:
Ideally, the entire payload shouldn't be passed to the workflow. However, the entire payload should be passed to a Trigger to allow extracting only certain desired content fields. In order to do this, you need to either specify a direct JSON field name or use some sort of regex matching in the sensor spec which implies you know the shape of the JSON content before you create the sensor (a totally reasonable assumption). 

The idea of then passing to the Argo Workflow is a bit trickier. Right now, triggers are fairly generic and can support any number of k8s resources. Passing a parameter to the workflow requires modifying the workflow in memory before creation. We should probably create a workflow specific trigger so that this is an easier problem to solve and we can integrate tighter with the existing `argo` platform. This will also more clearly define the RBAC roles for `argo-events`. Later, we can think about adding back functionality for different types of triggers.

vicaire:
magaldima@, the plan that you described looks good to me. Thanks!

magaldima:
this is more clearly defined/addressed in #50

PR #31, opened 2018-06-20
WIP - creating extensible signal plugin interface

### LINKED ISSUE #8: extensible signal interface
Users of Axis cannot be assumed to use the same signals as implemented in this project. Conversely, in order to maintain usefulness & extensibility, we would need to keep adding new signal types and sources, greatly adding to the complexity of the project.

I suggest a containerized interface for signals. Each signal executor is essentially a packaged image that implements a signal interface. Executors could send events defined as `CloudEvent` protos via `gRPC` to the host controller. These executors would still get deployed within the sensor controller and monitored during lifecycle of a sensor. The advantages being we can still leverage K8s jobs + pods for resiliency and scheduling and we can become extensible to be able to support any type of signal combination. There are two problems with this approach. 
1. How do we ensure the executor implements the signal interface and doesn't just execute any arbitrary code?
2. How do we define any extensible signal via a `Sensor` specification CRD? Will this entail hardcoded fields such as `url` and `key` and flexible fields such as maps to store additional information about that signal specification?

An alternative would be to define Axis as a library. The downside here is that the library itself would need to be tied very closely with Kubernetes and the Kube API.

magaldima:
This is a work in progress demonstrating an example of adding an extensible signal plugin interface.
closes #8 and #19


magaldima:
closes #19

VaibhavPage:
the codebase has definitely gotten cleaner. 😄

PR #33, opened 2018-06-21
Make executors imagePullPolicy configurable.

### LINKED ISSUE #32: Make job executors imagePullPolicy configurable
The sensor jobs executor has the imagePullPolicy of `PullIfNotPresent` hard-coded. This makes development difficult.

It will be good if the imagePullPolicy for the executor is also made configurable similar to the properties likes executor image name, resources, etc.

Continuing the current way would require deleting the executor image everytime from all the nodes in the cluster to ensure that the image is pulled again which is cumbersome.

shrinandj:
Testing Done:

1. Explicitly set the policy to "always" in the config-map. Verified that
   the executor pod created had the imagePullPolicy of "always".
2. Modified the imagePullPolicy to "IfNotPresent" and the same steps above.
3. Removed the `executorImagePullPolicy` from the configMap. Verified that
   the executor pod created had the imagePullPolicy of "always".

Closes #32.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -43,6 +43,10 @@ type SensorControllerConfig struct {
 	// ExecutorImage is the name of the image to run for the container inside the sensor executor jobs
 	ExecutorImage string `json:"executorImage"`
 
+	// ExecutorImagePullPolicy is the imagePullPolicy for the the executor. If this is empty,
+	// the imagePullPolicy is "Always".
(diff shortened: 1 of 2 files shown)


shrinandj:
I am using two K8S clusters. One is the K8s cluster that comes with the docker daemon on Mac and another is a K8S cluster on AWS. I push images to docker hub. On both, I was not seeing the updated image being used and started looking into how to fix it.

PR #36, opened 2018-06-26
creating extensible signal plugin interface

### LINKED ISSUE #8: extensible signal interface

magaldima:
# Signal plugins

## Introduction 
This is a draft design for incorporating a plugin-based design for signals. The difficulty in this approach is finding the right fit between configurations and extensibility.

## Initial Use Cases
1. User wishes to listen to events from outside the currently default builtin sensor functionality without having to change the code or how sensors operate.
2. User wishes to leverage an existing platform from their stack as a source way to trigger workflows.
3. Especially relevant for `Stream` signals as we want to use an internal messaging system as a signal source. 

## Plugin Principles
1. Functionality should first be added for signals which are used the most. 
2. Everything that is pluggable should be kept modular. (i.e separate binary)
3. Where possible, checks will be added to the sensor-controller to ensure the integrity of the plugins. (i.e. plugins follow a specific interface)

## Design
- `Signaler` interface is defined within the `shared` package. All signals must implement this interface.
```
// Signaler is the interface for signaling
type Signaler interface {
	Start(*v1alpha1.Signal) (<-chan *v1alpha1.Event, error)
	Stop() error
}
``` 
- Use [go-plugin](https://github.com/hashicorp/go-plugin) to register/create all signals. 
- Plugin must have a `main()` function and use `go-plugin` to serve up the plugin structure/functionality via RPC or gRPC. (One plugin per binary!)
- Binaries MUST be included in the sensor-controller's `Dockerfile` as:
```
COPY dist/nats-plugin /
ENV SIGNAL_PLUGIN=/nats-plugin
```

## Further Work
1. Finding the right middle ground between configuration and extensibility. The right approach for this may be that some signals are by default built-in such as calendar, resource, and webhooks. These types of signal sources are not subject to change due to different implementations. On the other hand, message streams and external file dependencies are subject to the user's specific stack and technologies. Thus, these benefit from a plugin-based architecture. Currently, the `Streams` definition makes it easy to use a custom stream, however we also lose configuration structure and have to ensure the correct fields are set later on in the plugin when we start the signal.
2. Including plugins without having to modify the `Dockerfile` and without taking down the sensor-controller. This could be accomplished by attaching a PVC to the sensor-controller as a location where we could copy plugin binary files. The controller has to know when these become available in order to create them. We could add configuration to the controller's `configmap` in order to track the available plugin binaries. This means we could run into scenarios where the configuration differs from the actual binaries available...
3. Building plugins outside the existing configuration scope. This includes new signals altogether to the existing `Streams`, `Artifacts`, `Resources`, `Calendar`, and `Webhooks`. An example for this would be: how do we implement a new custom `DB` signal without changing the `types.go` file? In this case, a user should be able to define their own configurations AND their own binary. If so, these configurations would exist outside the scope of the `Sensor` resource thus making the API non-declarative. 

closes #8


shrinandj:
I'm still get up to speed with this framework and specifics; so apologies if some of my questions are "too basic".

I understand that the need for having an "extensible signal plugin interface" so that newer signals can be added seamlessly. However, I'm not sure if a new executable per signal would be ideal. If users want to support new signals, they'll need to take the `argo-events` docker image, add their own binaries to that and roll out their own images. Everytime there's a new docker image of `argo-events`, they'll need to roll-out their new docker image too with the new `argo-events` docker image as the base.

Is this understanding correct?

magaldima:
@shrinandj Yes, that is the gist of this change. I know it doesn't seem ideal, but we're exploring options for being able to use internal services as signal sources (specifically for `Streams`) and also to be able to make it easy for others to do the same. If you have any suggestions or ideas for this, they would be very much appreciated!

shrinandj:
I don't have much experience with `Streams` and specific use-cases you have. We can checkin the current approach and revisit it with better options.

Issue #34, opened 2018-06-21
Add argo-ci and pr check


magaldima:
**Is your feature request related to a problem? Please describe.**
Since this project is now under the Argo Project, we should follow suite with the other projects and add an `.argo-ci/ci.yaml` file and setup the appropriate build/test checks.

**Describe the solution you'd like**
Create the CI process for Argo-Events.

**Describe alternatives you've considered**
I have a working `.travis.yml` file in the repo and was previously using Travis-CI test/builds to verify PR checks. I would need access to the Travis CI for Argo.

magaldima:
@edlee2121 @shrinandj Can you guys tackle this issue? Right now, we just need it to run the project tests.

shrinandj:
Sure. I'll do this.

shrinandj:
I have access to the Kubernetes cluster where Argo's current builds happen.

Currently creating the workflow for building `argo-events`. Once I have the workflow executing correctly, I'll create the appropriate webhooks in Github to send the request to K8s cluster for building everytime a PR is created.

shrinandj:
I created a workflow for running the CI for argo-events. However, I'm running into a problem where the workflow does not complete even though the pod execution has completed. Filed an issue on the main argo repo for it.

https://github.com/argoproj/argo/issues/898

shrinandj:
CI for pull-requests has been setup.

PR #42, opened 2018-07-05
Add workflow for CI.

shrinandj:
This commit creates the workflow that will be run on every PR for argo-events.

Testing Done:

1. Ran workflow independently on a cluster. It succeeded modulo appropriate RBAC role
(See https://github.com/argoproj/argo/issues/898 for details).

Refs #34.

### DIFF
--- a/.argo-ci/ci.yaml
+++ b/.argo-ci/ci.yaml
@@ -1,0 +1,73 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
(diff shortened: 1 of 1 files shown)


magaldima:
@shrinandj @jessesuen what's the argoproj policy around signing commits? I have it as part of the argo-events policy, but seems like the existing argo policy doesn't enforce it. Tell me if I should remove this check.

shrinandj:
We don't need to sign individual commits. Can you please remove the check?

Issue #46, opened 2018-07-18
Code generation with ./hack/update-codegen.sh fails


shrinandj:
**Describe the bug**

I was trying to add some additional fields to a struct in `types.go`. I added the code and tried to run `./hack/update-codegen.sh` to generate new code. However, that failed with the following error:
```
$ ./hack/update-codegen.sh
bash: ./vendor/k8s.io/code-generator/generate-groups.sh: No such file or directory
```

**Expected behavior**
`./hack/update-codegen.sh` should execute without any errors.

shrinandj:
I manually created the file `./vendor/k8s.io/code-generator/generate-groups.sh` and copied the [contents](https://raw.githubusercontent.com/kubernetes/code-generator/master/generate-groups.sh) from here into it. However, that failed with the following errors:
```
$ ./hack/update-codegen.sh
+ set -o errexit
+ set -o nounset
+ set -o pipefail
+ '[' 6 -lt 4 ']'
+ '[' deepcopy,client,informer,lister == --help ']'
+ GENS=deepcopy,client,informer,lister
+ OUTPUT_PKG=github.com/argoproj/argo-events/pkg/client
+ APIS_PKG=github.com/argoproj/argo-events/pkg/apis
+ GROUPS_WITH_VERSIONS=sensor:v1alpha1
+ shift 4
++ dirname ./vendor/k8s.io/code-generator/generate-groups.sh
+ cd ./vendor/k8s.io/code-generator
+ go install ./cmd/defaulter-gen ./cmd/client-gen ./cmd/lister-gen ./cmd/informer-gen ./cmd/deepcopy-gen
+ FQ_APIS=()
+ for GVs in ${GROUPS_WITH_VERSIONS}
+ IFS=:
+ read G Vs
+ for V in ${Vs//,/ }
+ FQ_APIS+=(${APIS_PKG}/${G}/${V})
+ '[' deepcopy,client,informer,lister = all ']'
+ grep -qw deepcopy
+ echo 'Generating deepcopy funcs'
Generating deepcopy funcs
++ codegen::join , github.com/argoproj/argo-events/pkg/apis/sensor/v1alpha1
++ local IFS=,
++ shift
++ echo github.com/argoproj/argo-events/pkg/apis/sensor/v1alpha1
[...]

PR #48, opened 2018-07-19
Upgrade to code-gen v1.10.

shrinandj:
However, there were some manual changes required to the fake clientset generated code. The newly generated
code also included a call to object.Watch() but there was no corresponding implementation for it. This code
was manually deleted. It should be safe since this was in the fake clientset code, the build succeeds and
the unit tests complete successfully.

Testing Done:

1. make test passes
2. Built and deployed the sensor-controller. Ran the example sensor. It completed successfully.

Refs #46


Gopkg.toml
```diff
@@ -30,7 +30,7 @@ required = [
 
 [[constraint]]
   name = "k8s.io/code-generator"
-  branch = "release-1.9"
+  branch = "release-1.10"
```
magaldima: Do we also need to update the `k8s.io/apimachinery` and `k8s.io/api` dependencies?

shrinandj: All of these could've been changed. But eventually they are dependent on github.com/argoproj/argo which is still at `release-1.9`. That cause the code generation to fail if we bump up the version of `k8s.io/apimachinery` and `k8s.io/api`

shrinandj:
The build failed because the docker image used for building argo-events doesn't include `protobufs`. I tried installing protobuf as part of the build workflow itself. However, that failed because installing protobuf requires unzipping a... well zip file. But the docker image doesn't include `unzip` either :( ...

Then, I tried creating an `argo-events` specific build image. But that failed because some of the instructions for installing protobuf on linux are vague.

I am going to try and fix that. Will keep this PR updated about my progress.

shrinandj:
I'll rebase this PR on top of https://github.com/argoproj/argo-events/pull/51 so that it uses the newer image for builds.

shrinandj:
Re-did some basic testing.
- Built the `sensor-controller`
- Ran the sample sensor. It completed successfully.

magaldima:
lgtm!

Issue #17, opened 2018-06-07
Long running sensors


VaibhavPage:
**Is your feature request related to a problem? Please describe.**
When a signal resolves, we resolve the state of sensor (consider it contains only one signal). If the sensor is repeatable, we reinitialize the sensor after the trigger is complete. But in case of signals like webhook, we want to keep HTTP server running and want to stop the signal only if it runs past certain predefined time or user/some process stops the signal.

**Describe the solution you'd like**
Add "type" to signal which specifies whether a signal is resolved when a single event occurs or it lives as long as specified timeout occurs. Timeout will be optional field, as we may want to keep the sensor running. 

example-

1) 
```
signals:
  - name: signal1
     mqtt:
        url: tcp://localhost:1883
        type: singular
        topic: hello
    name: signal2
```

2)
```
signals:
  - name: signal2
    webhook:
        port: 9000
        endpoint: "/app"
        method: "POST"
        type: continuous
        timeout: 10000s
```

kzadorozhny:
I see the problem with this approach.

HTTP sensor should be available for the consumers. Even periodically restarting HTTP service will hurt availability. 

To make HTTP sensors highly available, they should be provisioned as ReplicaSet resource owned by the sensor controller.

Same design applies to most sensors. Most of them can benefit from running continuously. Or if we look at the problem from the opposite side, none of sensors jobs need to be restarted after each completion.

In situations, when there are more than one dependency, those could coexist in the same process just fine.

magaldima:
@kzadorozhny thanks for the feedback. I have a couple of questions concerning this approach:
1. Is there a 1-1 mapping between Sensors and ReplicaSet resources? When I create a new sensor, does this result in the creation of a new ReplicaSet?
2. If answer to (1) is No and you have a continuously running ReplicaSet (ex: HTTP) sensor service, how do you configure multiple sensors (ex: HTTP) to be handled by the same service? How do you dynamically new signal implementations (ex: HTTP routes) to the service while it is running?
3. If answer to (1) is Yes, do you think having a separate ReplicaSet + multiple Pods running for a single sensor may be overkill? I think this implementation has a very high overhead for running a single sensor.

magaldima:
Closing this because #49 was merged. This PR essentially made signals as deployments. there is still some work to be done around webhooks, but I'll open another more specific issue for this.

PR #49, opened 2018-07-19
implementing signals as microservice deployments

magaldima:
This PR introduces a major shift in the architecture behind the `argo-events` signals. I know this PR has many changes, but I'll do my best to explain what is changing and why. I'll start with why.

**Why?**
1. Flexibility & Extensibility of signal sources. Users can build their own `micro` signal services without changing the `argo-events` code base. Users can deploy their own combination of signal support depending on their needs.
2. Zero-downtime for signals that should be always present e.g. webhooks (this is because the signals are now deployments and users can configure the deployments with >1 replicas).
3. Stateless signals simplifies signal code and allows 2.

**What is this?**
To get a better idea of what this change introduces, it would be helpful to understand the history of the `argo-events` signal functionality. 

Signals started out (as of the `v0.5-alpha1` release) as separate goroutines run inside of individual "executor" pods. For each active sensor, there was a running pod. 

A couple weeks ago, I merged in a change to make certain signals (artifact, calendar, resource, webhook) run as separate Goroutines within the sensor-controller pod while the stream signals (nats, kafka, mqtt, amqp) run as separate processing "plugged-in" to the controller binary via [go-plugin](https://github.com/hashicorp/go-plugin). All of the signals implemented the same interface, so calling every signal happened the same way.

Now, this PR introduces separate signal deployments which register themselves as [micro](https://github.com/micro/go-micro) services to any other `micro` clients. These services were enhanced in order to become stateless. In our case, the sensor-controller becomes the sole `micro` client and listens to signals via `gRPC` `Listen()`.


magaldima:
closes #17

shrinandj:
Is it fair to say that; after this change:

- Any K8s pod can register itself with `argo-events` as a signal. It will that pods job to figure out if that signal is fired. Whenever it does, that pod will inform the `sensor-controller` that it has fired and the `sensor-controller` should execute the trigger. The sensor pod should also determine when to resolve the signal (i.e. mark it completed and take some finalizing actions. e.g. stop the webserver).

magaldima:
@shrinandj just to clarify your understanding:

1. Yes, any k8s pod can register itself as a micro service that the `argo-events` sensor controller CAN use as a signal service. This depends on the name of the registered service and the type of the signal.

2. The `sensor-controller` is the `Listener` client so it initiates a bi-directional stream with a stateless process on a signal pod. The controller listens on the stream for events while the signal pod sends events on the stream. When the controller receives an event that satisfies it's sensor requirements, it sends a signal on the stream to terminate the signal. The signal pod then finishes any process associated with that req context. So to correct you, the controller determines when to resolve the signal, while the signal pod takes the necessary actions e.g. stopping the webserver.

magaldima:
@spk83 When a signal pod dies, all the bi-directional streams become disconnected from the sensor controller. When this happens, the controller's goroutines listening on the streams receive an error notifying them that the context has failed or the connection was closed. The signal nodes are then updated with an error. After the next processing loop of the affected sensor, the controller will attempt to re-establish a connection stream with the signal pod (up to the number of retries (3) for a sensor in an error phase). Therefore, if the signal pod is in a crash loop, the controller would try 3 times to make a connection, and after failing would not re-queue the sensor. The sensor resource and the signal node would be in an error phase.

shrinandj:
After this commit, do each of the signal docker images need to be deployed in the cluster individually?

PR #61, opened 2018-07-25
add-resource-params, fix filter, testing on minikube, updated quickstart

### LINKED ISSUE #50: argo workflow trigger with params
`Argo-events` was created in order to become the event framework for `Argo`. While we support creating argo `workflow` as part of a trigger, they are encompassed in the more general `ResourceObjects` which makes it difficult to couple valuable features like `Argo` params with the event framework. 

As discussed in #19 we should have an explicit argo workflow trigger so that we can easily pass arguments/params to the workflow.

**Additional context**
This may help to solve the problem of #41 depending on how the workflows are included in the sensor spec.

magaldima:
closes #50


examples/calendar-sensor.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         group: argoproj.io
         version: v1alpha1
         kind: Workflow
-        artifactLocation:
+        source:
```
shrinandj: +1 for changing this. Argo workflows themselves have a notion of artifacts and overloading that term was making it confusing.

shrinandj: In a following change, we should also rename structs from `S3Artifact` to `S3Source` or something like that.

magaldima:
TESTING EVIDENCE (from Minikube):

I created this webhook sensor:
```
kind: Sensor
metadata:
  name: webhook-with-resource-param
  labels:
    sensors.argoproj.io/controller-instanceid: axis
spec:
  signals:
    - name: webhook
      webhook:
        endpoint: /hello
        port: 7070
        method: POST
  triggers:
    - name: argo-workflow
      resource:
        namespace: default
        group: argoproj.io
        version: v1alpha1
        kind: Workflow
        # The parameters from the workflow are overridden by the webhook's message
        parameters:
          - src:
              signal: webhook
              path: message
              value: hello default
            dest: spec.arguments.parameters.0.value
        source:
          inline: |
              kind: Workflow
              metadata:
                name: arguments-via-webhook-event
              spec:
                entrypoint: whalesay
                arguments:
                  parameters:
                  - name: message
                    # this is the value that should be overridden
                    value: hello world
                templates:
                - name: whalesay
                  inputs:
                    parameters:
                    - name: message
                  container:
                    image: docker/whalesay:latest
                    command: [cowsay]
                    args: ["{{inputs.parameters.message}}"]
```

Then I sent an HTTP POST:
```
{
	"message": "this is my first webhook"
}
```

the logs of the argo workflow:
```
 __________________________
< this is my first webhook >
 --------------------------
    \
     \
      \
                    ##        .
              ## ## ##       ==
           ## ## ## ##      ===
       /""""""""""""""""___/ ===
  ~~~ {~~ ~~~~ ~~~ ~~~~ ~~ ~ /  ===- ~~~
       \______ o          __/
        \    \        __/
          \____\______/
```

examples/calendar-sensor.yaml
magaldima: agreed

magaldima:
Don't approve yet.. I gotta re-run dep ensure because of the new dependency on `tidwall/sjson`

magaldima:
okay all good now.

Issue #41, opened 2018-06-29
Argo-events should not require workflows to be in an S3 bucket


shrinandj:
**Is your feature request related to a problem? Please describe.**
Currently, `argo-events` requires and `artifactLocation` in the Trigger. This `artifactLocation` has the details of the workflow to run when the trigger is fired. This is cumbersome to setup (even if it is with minio). 

**Describe the solution you'd like**
It will be better if there is no need for the S3 bucket. Ideally, the workflow to run could be inlined with the Sensor yaml. If not, there could be other options of have the workflow yaml in a config-map.

**Describe alternatives you've considered**
None

shrinandj:
**Requirements:**

ArtifactLocation is a yaml map. The only key currently allowed being "s3":

```
artifactLocation: 
  s3: 
    accessKey: 
      key: accesskey
      name: artifacts-minio
    bucket: workflows
    endpoint: "artifacts-minio.default:9000"
    insecure: true
    key: hello-world.yaml
    secretKey: 
      key: secretkey
      name: artifacts-minio
```

Users should be able to specify various types of locations for the K8s resources in the trigger.

Possible options:

**Inline:**
It should be possible for the user to have the entire yaml directly embedded in the sensor yaml itself.

```
artifactLocation: 
  inline: |
      apiVersion: argoproj.io/v1alpha1
      kind: Workflow
      metadata: 
        generateName: hello-world-
      spec: 
        entrypoint: whalesay
        templates: 
          - 
            container: 
              args: 
                - "hello world"
              command: 
                - cowsay
              image: "docker/whalesay:latest"
            name: whalesay
```

**file:** 
Assuming that the sensor controller is run where it has access to the workflow yaml file, having a file option would be great.

Users could have the workflow yamls in a config map that is mounted at some path in the sensor-controller as a volume.

Users could have the workflow yamls in a dedicated volume that is mounted in the sensor-controller.

```
artifactLocation: 
  file: 
    path: /my-argo-yamls/my-workflow.yaml
```

**url:**
Argo submitting workflows using a url. So does kubernetes. It would be good if users can also specify that the artifactLocation is a url.

```
artifactLocation: 
  url: 
    path: https://raw.githubusercontent.com/my-argo-yamls/my-workflow.yaml
```

**Design:**

_Changes to types.go_

Currently, "Trigger" -> "ResourceObject" -> "ArtifactLocation".

ArtifactLocation only has one member of type S3Artifact.
```
[...]

PR #65, opened 2018-07-25
Support file trigger sources.

shrinandj:
With this change, sensor yamls can have a trigger in which the workflow
yaml is specified as a file path. This will enable cases where workflow
yamls are in config-maps and/or in some persistent volumes.

Refs #41


magaldima:
This looks good! I'm happy to merge in as is, but we should update the documentation to explain how to use this feature as the provided example probably doesn't work with the given `sensor-controller` deployment.

shrinandj:
I'll update the documentation in a follow-up PR. I added a comment in the example which says that the file should be accessible to the sensor-controller.

PR #66, opened 2018-07-26
Support URL trigger sources

shrinandj:
Sensor yamls can now specify the workflow location as a url path with this change.

Refs #41


pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -448,8 +448,9 @@ type URI struct {
 
 // ArtifactLocation describes the source location for an external artifact
 type ArtifactLocation struct {
-	S3     *S3Artifact `json:"s3,omitempty" protobuf:"bytes,1,opt,name=s3"`
-	Inline string      `json:"inline,omitempty" protobuf:"bytes,2,opt,name=inline"`
+	S3     *S3Artifact  `json:"s3,omitempty" protobuf:"bytes,1,opt,name=s3"`
+	Inline string       `json:"inline,omitempty" protobuf:"bytes,2,opt,name=inline"`
+	URL    *URLArtifact `json:"url,omitempty" protobuf:"bytes,4,opt,name=url"`
```
shrinandj: The offset is intentionally set to 4. 3 will be used for file artifacts (https://github.com/argoproj/argo-events/pull/65).

store/url.go
```diff
@@ -0,0 +1,49 @@
...
+func NewURLReader(urlArtifact *v1alpha1.URLArtifact) (ArtifactReader, error) {
+	if urlArtifact == nil {
+		panic("URLArtifact cannot be empty")
+	}
+	return &URLReader{urlArtifact}, nil
+}
+
+func (reader *URLReader) Read() ([]byte, error) {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
magaldima: Do you think we should be able to support providing options/configuration to the `http.Client` through the `UrlArtifact`?

PR #67, opened 2018-07-26
Add verifycert as an argument to URL trigger source

shrinandj:
Testing Done:

1. Created a sensor with verifycert explicitly set to false. It succeeded.
2. Created a sensor without verifycert. It succeeded.

Refs #41


magaldima:
the tests passed without changing the unit test... weird. I'll approve for now, but if i notice weird behavior for this, I'll raise a PR to keep the webhook unit_tests contained locally

PR #68, opened 2018-07-26
removing port from webhook and revising signal impl

### LINKED ISSUE #59: Remove port from WebhookSignal
`WebhookSignal`s currently allow users to specify a port to listen on, however this wouldn't be respected because the ClusterIP Service (for the webhook signal deployment) needs to be configured at deployment time and therefore is fixed at runtime. 

We should not let the user choose a port to listen on. I envision this would present a problem for the webhook signal unit tests so we'd likely have to modify the webhook signal implementation. I'm leaning toward making the webhook struct contain a reference to the http server which would be started on calling `New()`. This would further entail creating a custom http `Mux` that would allow us to not only register new endpoints, but also deregister endpoints. We'll have to look into this further as the `DefaultHttpMux` purposely chose not to do this because https://groups.google.com/forum/#!topic/golang-dev/kgN2TiUQf3M

magaldima:
closes #59


signals/webhook/webhook.go
```diff
@@ -38,27 +37,67 @@ const (
...
+		// Good practice to enforce timeouts to avoid Slowloris attacks
+		WriteTimeout: time.Second * 5,
+		ReadTimeout:  time.Second * 5,
+		IdleTimeout:  time.Second * 30,
+	}
+	// Start http server
+	go func() {
+		log.Printf("starting http server listening on: %s", srv.Addr)
+		err := srv.ListenAndServe()
+		if err == http.ErrServerClosed {
```
magaldima: should all errors be panicked since the server should never be shutdown under normal operations?

signals/webhook/webhook.go
```diff
@@ -38,27 +37,67 @@ const (
...
+		Addr: fmt.Sprintf(":%v", port),
+		// Good practice to enforce timeouts to avoid Slowloris attacks
+		WriteTimeout: time.Second * 5,
+		ReadTimeout:  time.Second * 5,
+		IdleTimeout:  time.Second * 30,
+	}
+	// Start http server
+	go func() {
+		log.Printf("starting http server listening on: %s", srv.Addr)
+		err := srv.ListenAndServe()
```
shrinandj: In a follow-up commit, we should support TLS.

signals/webhook/webhook.go
shrinandj: One situation where it might be needed to shutdown the server would be as part of restarting it. E.g. if some config option has been changed and we need to restart the server. Not sure if we do that today...

signals/webhook/webhook.go
```diff
@@ -38,27 +37,67 @@ const (
...
+// the connections will be re-initialized by the SignalClient
+type webhook struct {
+	srv *http.Server
+	sync.RWMutex
+	inactiveEndpoints map[string]struct{}
+}
+
+// New creates a new webhook listener for the specified port
+func New(port int) sdk.Listener {
+	srv := &http.Server{
```
shrinandj: Question about webhook signal in-general. 

The sensor yaml for webhooks will have the following:
```
    - name: webhook
      webhook:
        endpoint: /app
        method: POST
```

As a user, how can one reach the `/app` endpoint? Is there a well-defined Kubernetes service which can be used for accessing this from within the cluster?

signals/webhook/webhook.go
magaldima: good idea

signals/webhook/webhook.go
magaldima: the `hack/k8s/manifests/services/webhook.yaml` has a deployment and a load balancer service for this signal. does this answer your question?

signals/webhook/webhook.go
magaldima: in that case, someone can just delete the pod(s) for this signal, however I don't foresee a need to do that today

signals/webhook/webhook.go
```diff
@@ -38,27 +37,67 @@ const (
...
 // Note: micro requires stateless operation so the Listen() method should not use the
 // receive struct to save or modify state.
-type webhook struct{}
+// webhooks are special in that they should only have one http Server since the port is fixed at runtime
+// this means that webhook signals are stateful, however losing this state is not a concern since
+// the connections will be re-initialized by the SignalClient
+type webhook struct {
+	srv *http.Server
+	sync.RWMutex
+	inactiveEndpoints map[string]struct{}
```
shrinandj: Can we use Golang's sync map instead? https://golang.org/pkg/sync/#Map

signals/webhook/webhook.go
```diff
@@ -38,27 +37,67 @@ const (
...
+}
+
+// New creates a new webhook listener for the specified port
+func New(port int) sdk.Listener {
+	srv := &http.Server{
+		Addr: fmt.Sprintf(":%v", port),
+		// Good practice to enforce timeouts to avoid Slowloris attacks
```
shrinandj: Learn something new everyday! TIL "Slowloris attacks" 👍 👍 :-)

signals/webhook/webhook.go
magaldima: so use cases are:
- when the entry for a given key is only ever written once but read many times, as in caches that only grow
- when multiple goroutines read, write, and overwrite entries for disjoint sets of keys

I think both of these may apply to some degree.. so it may be worth losing the type safety over this...

signals/webhook/webhook.go
magaldima: I'll allow supporting TLS in next commit

magaldima: If this is expected to run inside a kubernetes pod, what is the best place to get the cert and key file and/or certificates?

magaldima: I'm going to push this off, if this is really needed, let's raise another issue for this

shrinandj: Let's do another issue for it. Two possibilities:

1. Self-signed cert (the webhook signal pod will simply create it's own cert and private key)
2. User provided cert and key (maybe volume mounted in the webhook signal pod?)

shrinandj:
Not sure why the build is failing...

magaldima:
# github.com/argoproj/argo-events/signals/webhook/micro/client
signals/webhook/micro/client/main.go:38:8: unknown field 'Port' in struct literal of type v1alpha1.WebhookSignal

fixing now

PR #92, opened 2018-08-22
WIP: re-architecting around gateways and sensors

VaibhavPage:


magaldima:
Couple things:
1. can you provide a brief summary of the high-level API changes involved in this PR?
2. please update the readme's, quickstart and docs with this new information.
3. please update the `hack/k8s/manifests` with the new yaml files for gateway-deployment.
4. the `pkg/*-client` directories can be changed to just `pkg/clients/` with sub directories for `gateway` and `sensor` just like the `pkg/apis/` directory.
5. can we group the top level `gateway-controller` and `sensor-controller` dirs under a common `controllers` directory and rename the sub dirs without the `-controller` suffix?
6. we should remove all project dependencies and reference to `github.com/micro/go-micro` and `github.com/micro/go-plugins`
7. we should be consistent with the logger we use across the project. the gateway controller, transformer, and individual gateways use `github.com/rs/zerolog` while the sensor controller uses `github.com/sirupsen/logrus`. What's your reasoning for using this new logger?

Overall, this looks really cool, I'm excited to try this out.

common/util.go
```diff
@@ -50,5 +76,17 @@ func ServerResourceForGroupVersionKind(disco discovery.DiscoveryInterface, gvk s
 			return &r, nil
 		}
 	}
-	return nil, fmt.Errorf("Server is unable to handle %s", gvk)
+	return nil, fmt.Errorf("server is unable to handle %s", gvk)
+}
+
+// SendSuccessResponse sends http success response
+func SendSuccessResponse(writer *http.ResponseWriter) {
+	(*writer).WriteHeader(http.StatusOK)
```
magaldima: why do you need to dereference the `writer` here?

examples/gateways/artifact-gateway-configmap.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: artifact-gateway-configmap
+  namespace: argo-events
```
magaldima: do we need to specify the namespace here? in an example i would think ppl can deply to whatever namespace they choose to use.

examples/gateways/artifact-gateway-configmap.yaml
```diff
@@ -0,0 +1,21 @@
...
+data:
+  s3.myartifact: |-
```
magaldima: how did you choose this naming convention? I see `s3.myartifact` and others such as `calendar.myConfig` and `nats.myConfig`.. is this standardized across all gateways?

gateway-controller/transform/transformer.go
```diff
@@ -0,0 +1,146 @@
...
+	// Generate an event id
+	eventId := suuid.Must(suuid.NewV4(), nil)
+	payload, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		fmt.Errorf("failed to parse request payload. Err %+v", err)
+		return nil, err
+	}
+
+	// Create an CloudEvent
+	ce := &sv1alpha.Event{
```
magaldima: nice now that this is defined in one place!

gateways/core/artifact/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM centos:7
```
magaldima: any reason why you are using this image? can we not build from `scratch`

gateways/core/artifact/s3.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// Create a done channel to control 'ListenBucketNotification' go routine.
+	doneCh := make(chan struct{})
+
+	// Indicate to our routine to exit cleanly upon return.
+	defer close(doneCh)
+
+	// Listen for bucket notifications
+	for notificationInfo := range minioClient.ListenBucketNotification(artifact.S3EventConfig.Bucket, artifact.S3EventConfig.filter.Prefix, artifact.S3EventConfig.filter.Suffix, []string{
```
magaldima: I see that you're using `ListenBucketNotifiation`.. I previously used this and removed it from my code because it is currently only supported by `minio` s3 and is not part of the standard s3 api. Obviously this makes the code a lot cleaner as you no longer need a stream, but shouldn't we be able to support all s3 sources and not just minio?

gateways/core/calendar/calendar.go
```diff
@@ -0,0 +1,212 @@
...
+			payload, err := event.Marshal()
+			if err != nil {
+				return fmt.Errorf("failed to marshal event. Err: %+v", err)
+			} else {
+				// dispatch the event to gateway transformer
+				_, err = http.Post(fmt.Sprintf("http://localhost:%s", c.gatewayConfig.TransformerPort), "application/octet-stream", bytes.NewReader(payload))
+				if err != nil {
+					return fmt.Errorf("failed to dispatch the event. Err: %+v", err)
+				}
+				c.gatewayConfig.Log.Info().Msg("event dispatched to gateway transformer")
```
magaldima: what about creating `kubernetes.Events` for this? this would enable us to create a log of all the events dispatched...

hack/k8s/manifests/gateway-controller-configmap.yaml
```diff
@@ -0,0 +1,9 @@
+# The gateway-controller configmap includes configuration information for the gateway-controller
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gateway-controller-configmap
+  namespace: argo-events
```
magaldima: let's remove the namespace references from manifests..

hack/update-codegen.sh
```diff
@@ -21,7 +21,15 @@ set -o pipefail
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
 CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
 
+CODEGEN_PKG="../vendor/k8s.io/code-generator"
+
 bash -x ${CODEGEN_PKG}/generate-groups.sh "deepcopy,client,informer,lister" \
-  github.com/argoproj/argo-events/pkg/client github.com/argoproj/argo-events/pkg/apis \
+  github.com/argoproj/argo-events/pkg/sensor-client github.com/argoproj/argo-events/pkg/apis \
```
magaldima: we should be using `github.com/argoproj/argo-events/pkg/client` for both

pkg/apis/gateway/v1alpha1/types.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// Gateway is the definition of a gateway-controller resource
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type Gateway struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata" protobuf:"bytes,1,opt,name=metadata"`
+	Status            NodePhase   `json:"type" protobuf:"bytes,2,opt,name=type"`
```
magaldima: should this be of type `GatewayStatus` instead? make sure to edit the annotations to match the name as well!

pkg/apis/gateway/v1alpha1/types.go
```diff
@@ -0,0 +1,94 @@
...
+	Version string `json:"version" protobuf:"bytes,6,opt,name=version"`
+
+	// Service is the name of the service to expose the gateway
+	Service Service `json:"service,omitempty" protobuf:"bytes,7,opt,name=service"`
+
+	// Sensors are list of sensors to dispatch events to
+	Sensors []string `json:"sensors" protobuf:"bytes,8,opt,name=sensors"`
+
+	// ServiceAccountName is name of service account to run the gateway
+	ServiceAccountName string `json:"serviceAccountName" protobuf:"bytes,9,opt,name=service_account_name"`
```
magaldima: the protobuf naming scheme should be camelCase: `serviceAccountName`

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -25,18 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// SignalType is the type of the signal
-type SignalType string
```
magaldima: great to see this go!

Makefile
```diff
@@ -15,8 +15,8 @@ override LDFLAGS += \
   -X ${PACKAGE}.gitTreeState=${GIT_TREE_STATE}
 
 # docker image publishing options
-DOCKER_PUSH=false
-IMAGE_NAMESPACE=argoproj
+DOCKER_PUSH=true
+IMAGE_NAMESPACE=metalgearsolid
```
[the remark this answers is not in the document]
VaibhavPage: its my registry name :) . The changes such as registry name, adding namespaces to manifests etc were all part of local testing and I had planned to remove them before resolving the WIP status.

common/common.go
```diff
@@ -17,39 +17,113 @@ limitations under the License.
...
-	EnvVarNamespace = "SENSOR_NAMESPACE"
-
-	// EnvVarConfigMap is the name of the configmap to use for the controller
+	// EnvVarConfigMap is the name of the configmap to use for the sensor-controller
 	EnvVarConfigMap = "SENSOR_CONFIG_MAP"
 
-	// EnvVarKubeConfig is the path to the Kubernetes configuration
-	EnvVarKubeConfig = "KUBE_CONFIG"
+	// Sensor image is the image used to deploy sensor.
+	SensorImage = "metalgearsolid/sensor"
```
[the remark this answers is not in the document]
VaibhavPage: same as above

examples/gateways/artifact-gateway-configmap.yaml
VaibhavPage: They are just random names I came up to test gateways. I haven't thought of how naming convention should like. Also do we need a convention? coz when user creates a custom gateway, its upto he/she to interpret the configuration.

gateways/core/artifact/Dockerfile
VaibhavPage: Definitely. I changed it to centos so I could interact with the pod through shell for testing. I've planned to revert it to scratch.

hack/update-codegen.sh
VaibhavPage: Agreed.

pkg/apis/gateway/v1alpha1/types.go
VaibhavPage: yup.

gateways/core/calendar/calendar.go
VaibhavPage: Do you have something in mind like a gateway dispatches an event by creating kubernetes.Event with "type" of particular configuration of that gateway and sensors then listens to such event/s? 

This does make the ecosystem completely kubernetes native and remove need for a pub-sub system that sit between gateway and sensor. Is etcd storage a right choice to store events for production grade deployment of argo-events?

Another approach we can take is to make framework opinionated about internal pub-sub and use NATS streaming. Gateways will publish to NATS streaming cluster and sensors will subscribe to it. This makes gateways and sensors completely decoupled from each other. Some of the advantages NATS streaming has to offer are "Historical message replay by subject" , "At-least-once-delivery" and "Message/event persistence". We can also provide out-of-box support for indexing and monitoring using elastic search and prometheus if an user wishes to use them. 

Thoughts?

controllers/gateway/operator.go
```diff
@@ -40,20 +47,25 @@ func newGatewayOperationCtx(gw *v1alpha1.Gateway, controller *GatewayController)
...
 		return err
 	}
-	gatewayClient := goc.controller.gatewayClientset.ArgoprojV1alpha1().Gateways(goc.gw.Namespace)
 
 	// manages states of a gateway
-	switch goc.gw.Status {
+	switch goc.gw.Status.Phase {
 	case v1alpha1.NodePhaseNew:
+
+		goc.markGatewayPhase(v1alpha1.NodePhaseNew, "new")
```
magaldima: this seems redundant as the phase is already new

controllers/gateway/operator.go
```diff
@@ -40,20 +47,25 @@ func newGatewayOperationCtx(gw *v1alpha1.Gateway, controller *GatewayController)
...
+
 		// Update node phase to running
-		goc.gw.Status = v1alpha1.NodePhaseRunning
+		goc.gw.Status.Phase = v1alpha1.NodePhaseRunning
```
magaldima: un-needed, you update the phase at the end of operation and this update happens in the `markGatewayPhase` function anyway..

controllers/gateway/operator.go
```diff
@@ -40,20 +47,25 @@ func newGatewayOperationCtx(gw *v1alpha1.Gateway, controller *GatewayController)
...
 	if err != nil {
 		goc.log.Error().Err(err).Msg("gateway validation failed")
+		goc.markGatewayPhase(v1alpha1.NodePhaseError, "validation failed")
 		return err
 	}
-	gatewayClient := goc.controller.gatewayClientset.ArgoprojV1alpha1().Gateways(goc.gw.Namespace)
 
 	// manages states of a gateway
-	switch goc.gw.Status {
+	switch goc.gw.Status.Phase {
```
magaldima: thoughts on adding another phase `pending` since this may be relevant for situations after creation deployment + service but before creation of gateway deployment pod?

controllers/gateway/operator.go
```diff
@@ -76,22 +88,19 @@ func (goc *gwOperationCtx) operate() error {
...
-			goc.gw.Status = v1alpha1.NodePhaseError
-			goc.gw, err = gatewayClient.Update(goc.gw)
-			if err != nil {
-				err = goc.reapplyUpdate(gatewayClient)
-				if err != nil {
-					goc.log.Error().Err(err).Msg("failed to update gateway")
-					return err
-				}
-			}
+			goc.markGatewayPhase(v1alpha1.NodePhaseError, "failed to create transformer gateway configuration")
```
magaldima: it's useful to include the `err` in the message.

controllers/gateway/operator.go
```diff
@@ -159,22 +168,13 @@ func (goc *gwOperationCtx) operate() error {
 		_, err = goc.controller.kubeClientset.AppsV1().Deployments(goc.gw.Namespace).Create(gatewayDeployment)
 		if err != nil {
 			goc.log.Error().Err(err).Msg("failed gateway deployment")
-			goc.gw.Status = v1alpha1.NodePhaseError
+			goc.markGatewayPhase(v1alpha1.NodePhaseError, "failed gateway deployment")
```
magaldima: same deal with including `err` in the message

controllers/gateway/operator.go
```diff
@@ -159,22 +168,13 @@ func (goc *gwOperationCtx) operate() error {
...
 		} else {
-			goc.gw.Status = v1alpha1.NodePhaseRunning
+			// expose gateway if service is configured
 			if goc.gw.Spec.Service.Port != 0 {
 				goc.createGatewayService()
```
magaldima: why do we fail silently here?

gateways/core/artifact/s3.go
```diff
@@ -40,18 +40,24 @@ const (
 
 type s3 struct {
 	// gatewayConfig provides a generic configuration for a gateway
-	gatewayConfig *gateways.GatewayConfig
-	registeredArtifacts []S3Artifact
+	gatewayConfig       *gateways.GatewayConfig
+	// registeredArtifacts contains map of registered s3 artifacts
+	registeredArtifacts map[uint64]*S3Artifact
```
magaldima: what is the purpose of this map? it doesn't look like we're saving anything in this? also the gateway doesn't want to be storing state correct?

controllers/gateway/operator.go
VaibhavPage: oops. got carried away there marking gateway phases :)

controllers/gateway/operator.go
VaibhavPage: makes sense.

controllers/gateway/operator.go
VaibhavPage: :+1:

controllers/gateway/operator.go
VaibhavPage: As the service is not absolutely critical part of the gateway. An user can create the service manually in the event of service creation failure.

gateways/core/artifact/s3.go
VaibhavPage: A gateway can be configured with multiple configuration. Each configuration is run in different go routine. The idea is to use a single gateway among multiple use cases that differ only in configurations. 

We are watching gateway-configmap to dynamically add new configuration and the purpose of this map is to keep track of configurations of gateway. Whenever an user updates the configmap with new configuration, this map is used to first check whether the configuration already exists or is it a completely new configuration.

VaibhavPage:
It makes sense to add number of completions count in sensor #6

VaibhavPage:
I was looking for a fast JSON output logger and came across zerolog. From the benchmarking posted on their github, it performs better than uber/zap.

controllers/gateway/validate.go
```diff
@@ -0,0 +1,20 @@
...
+	if goc.gw.Spec.Image == "" {
+		return fmt.Errorf("gateway image is not specified")
+	}
+	if goc.gw.Spec.Type == "" {
+		return fmt.Errorf("gateway type is not specified")
+	}
+	if len(goc.gw.Spec.Sensors) <= 0 {
+		return fmt.Errorf("no associated sensor with gateway")
+	}
+	if goc.gw.Spec.ServiceAccountName == "" {
```
VaibhavPage: use default service account instead

controllers/gateway/operator.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	_, err := goc.controller.kubeClientset.CoreV1().Services(goc.gw.Namespace).Create(gatewayService)
+	// Fail silently
+	if err != nil {
+		goc.log.Error().Err(err).Msg("failed to create service for gateway deployment")
+	}
+}
+
+// persist the updates to the Sensor resource
```
VaibhavPage: update comment, remove sensor resource

controllers/sensor/controller.go
```diff
@@ -55,7 +53,7 @@ type SensorController struct {
 	ConfigMap string
 	// namespace for the config map
 	ConfigMapNS string
-	// Config is the sensor controller's configuration
+	// Config is the sensor sensor-controller's configuration
```
VaibhavPage: remove sensor from the comment

common/util.go
```diff
@@ -80,13 +80,13 @@ func ServerResourceForGroupVersionKind(disco discovery.DiscoveryInterface, gvk s
...
+func SendSuccessResponse(writer http.ResponseWriter) {
+	writer.WriteHeader(http.StatusOK)
+	writer.Write([]byte(SuccessResponse))
 }
 
 // SendErrorResponse sends http error response
-func SendErrorResponse(writer *http.ResponseWriter) {
-	(*writer).WriteHeader(http.StatusOK)
-	(*writer).Write([]byte(SuccessResponse))
+func SendErrorResponse(writer http.ResponseWriter) {
```
magaldima: you should still be able to pass a pointer here.

gateways/core/artifact/s3.go
magaldima: so what happens when a gateway goes down and loses this information? does it pick up the existing configurations from the configMap and restart the necessary go routines?

controllers/gateway/operator.go
```diff
@@ -224,7 +224,7 @@ func (goc *gwOperationCtx) createGatewayService() {
 			Ports: []corev1.ServicePort{
 				{
 					Port:       goc.gw.Spec.Service.Port,
-					TargetPort: intstr.FromInt(goc.gw.Spec.Service.TargetPort),
+					TargetPort: intstr.FromInt(int(goc.gw.Spec.Service.TargetPort)),
```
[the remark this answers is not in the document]
VaibhavPage: makes sense. I'll add it

common/util.go
VaibhavPage: it is a pointer.
 https://stackoverflow.com/questions/13255907/in-go-http-handlers-why-is-the-responsewriter-a-value-but-the-request-a-pointer

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.6-alpha
```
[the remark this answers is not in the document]
VaibhavPage: and its from v0.6 onwards, right? or are marking this as v0.5?

VaibhavPage:
As gateway can have multiple configurations, each configuration within a gateway is it's own source. Sensor can then wait for notifications from different configurations from either same or different gateways. The naming convention for signal will be `gateway-name/configuration-name`.

VaibhavPage:
`Note`: gateway will send all notifications from different configurations to sensor because we just mention `sensor-name` in gateway's description. Although sensor will just reject notifications from configurations it is not interested in, its an eyesore to look at the sensor logs. In future, once we add pub-sub implementation, a configuration from gateway will publish with topic name  `gateway-name/configuration-name` and sensor will subscribe to it. So, sensor will only receive notifications from configuration/s in a gateway/s it is interested in.

gateways/core/calendar/calendar.go
magaldima: my thought here wasn't about "production" grade messaging layer. it was more for a temporary reporting/accounting store. I don't think using Kubernetes Events here is a good idea, however i do believe an internal pub/sub interface/mechanism (with persistent and guarantees) would be. I do however think possible recording meta-events about the state of gateways/sensors through Kubernetes events would be interesting. for example, publishing an event when a gateway reaches a running state.. this seems to be the kind of behavior other default Kubernetes objects do during their lifecycle (e.g. pods -> pulling images, creating container, etc.)

pkg/apis/gateway/v1alpha1/types.go
```diff
@@ -61,14 +59,24 @@ type GatewaySpec struct {
...
-	ServiceAccountName string `json:"serviceAccountName" protobuf:"bytes,9,opt,name=serviceAccountName"`
+	ServiceAccountName string `json:"serviceAccountName,omitempty" protobuf:"bytes,9,opt,name=serviceAccountName"`
+
+	// Affinity for scheduling
+	Affinity *corev1.Affinity `json:"affinity,omitempty" protobuf:"bytes,10,opt,name=affinity"`
+
+	// ImagePullPolicy for pulling the image
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty" protobuf:"bytes,11,opt,name=imagePullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
+
+	// todo: do we need separate labels or just use same labels defined on gateway resource?
```
magaldima: i don't think we need separate labels. we get labels from the serviecSpec and from the gateway metadata. just need to be clear in the specification/declaration where each one applies.

pkg/apis/gateway/v1alpha1/types.go
```diff
@@ -42,17 +42,15 @@ type GatewayList struct {
 
 // GatewaySpec represents gateway specifications
 type GatewaySpec struct {
+
 	// Image is the gateway processor image
 	Image string `json:"image" protobuf:"bytes,1,opt,name=image"`
```
magaldima: what's your thought about using a `corev1.Container` definition instead of Image, ImagePullPolicy, Affinity and such?

gateways/gateway.go
```diff
@@ -79,10 +115,211 @@ func (gc *GatewayConfig) newConfigMapWatch(name string) *cache.ListWatch {
...
+	}
+	staleConfigKeys, newConfigKeys := gc.diffConfigurations(newConfigs)
+
+	gc.log.Info().Interface("stale-config-keys", staleConfigKeys).Msg("stale config")
+	gc.log.Info().Interface("new-config-keys", newConfigKeys).Msg("new config")
+
+	// run new configurations
+	for _, newConfigKey := range newConfigKeys {
+		newConfig := newConfigs[newConfigKey]
+		gc.registeredConfigs[newConfigKey] = newConfig
```
magaldima: might want a lock here since despite what we previously discussed. this function is getting executed from `ResourceEventHandlerFuncs` from the watch and we're mutating a shared map.

gateways/gateway.go
```diff
@@ -79,10 +115,211 @@ func (gc *GatewayConfig) newConfigMapWatch(name string) *cache.ListWatch {
...
+
+	// create a client for gateway server
+	client := gwProto.NewGatewayExecutorClient(conn)
+
+	eventStream, err := client.RunGateway(context.Background(), &gwProto.GatewayConfig{
+		Config: config.config,
+		Src:    config.src,
+	})
+
+	gc.log.Info().Str("config-key", config.src).Msg("connected with server and got a event stream")
```
magaldima: put this log after the error check

gateways/gateway.go
```diff
@@ -79,10 +115,211 @@ func (gc *GatewayConfig) newConfigMapWatch(name string) *cache.ListWatch {
...
+
+	if err != nil {
+		gc.log.Error().Str("config-key", config.src).Err(err).Msg("failed to get event stream")
+		return err
+	}
+
+	config.conn = conn
```
magaldima: we shouldn't need this. as discussed use a context instead.

gateways/gateway.go
```diff
@@ -79,10 +115,211 @@ func (gc *GatewayConfig) newConfigMapWatch(name string) *cache.ListWatch {
...
+
+	for {
```
magaldima: you can do `for event, err := range eventStream.Recv() {...}`

gateways/gateway.go
```diff
@@ -79,10 +115,211 @@ func (gc *GatewayConfig) newConfigMapWatch(name string) *cache.ListWatch {
...
+		event, err := eventStream.Recv()
+		if err == io.EOF {
+			gc.log.Info().Str("config-key", config.src).Msg("event stream stopped")
+			return nil
+		}
+		if err != nil {
+			gc.log.Warn().Str("config-key", config.src).Err(err).Msg("failed to receive events on stream")
+			return nil
```
magaldima: should return the err

gateways/gateway.go
```diff
@@ -79,10 +115,211 @@ func (gc *GatewayConfig) newConfigMapWatch(name string) *cache.ListWatch {
...
+			break
+		}
+		payload, err := utils.TransformerPayload(event.Data, config.src)
+		if err != nil {
+			gc.log.Error().Str("config-key", config.src).Err(err).Msg("failed to transform request body")
+			return err
+		}
+
+		gc.log.Info().Str("config-key", config.src).Msg("dispatching the event to gateway-transformer...")
+		_, err = http.Post(fmt.Sprintf("http://localhost:%s", gc.transformerPort), "application/octet-stream", bytes.NewReader(payload))
```
magaldima: let's do an err check and put the log (with past tense) after the check.

gateways/gateway.go
```diff
@@ -79,10 +115,211 @@ func (gc *GatewayConfig) newConfigMapWatch(name string) *cache.ListWatch {
...
+}
+
+// diffConfig diffs currently registered configurations and the configurations in the gateway configmap
+// retunrs staleConfig - configurations to be removed from gateway
+// newConfig - new configurations to run
+func (gc *gatewayConfig) diffConfigurations(newConfigs map[uint64]*configData) (staleConfigKeys []uint64, newConfigKeys []uint64) {
+	var currentConfigKeys []uint64
+	var updatedConfigKeys []uint64
+
+	for currentConfigKey, _ := range gc.registeredConfigs {
```
magaldima: need a read lock here

VaibhavPage:
I'll add documentation as soon as possible.

VaibhavPage:
Added K8 events for gateway.

VaibhavPage:
K8 events currently work as mechanism for gateway to update it's state and avoid synchronization within multiple go routines altogether and have multiple gateway resource watchers. Event watcher currently sit within gateway processor. I am thinking of moving it to Gateway controller. Event watcher in gateway processor was listening to k8 events that are specific to that gateway, but Gateway operator will listen to k8 events of type `Gateway`. 
I am also thinking of watching k8 events to maintain Sensor state at Sensor controller. This will remove necessity to watch sensor resource in sensor pod.

controllers/gateway/transform/transformer.go
```diff
@@ -98,46 +103,96 @@ func (toc *tOperationCtx) transform(r *http.Request) (*sv1alpha.Event, error) {
...
 		Payload: tp.Payload,
 	}
+
+	toc.log.Debug().Interface("cloud-event", ce).Msg("transformed event")
 	return ce, nil
 }
 
-// dispatches the event to configured sensor
+// getWatcherIP returns IP of service which backs given component.
+func (toc *tOperationCtx) getWatcherIP(name string) (string, string, error) {
```
magaldima: this method seems a bit odd to me.. shouldn't we be using DNS with service names instead of inspected for IP addresses?

docs/custom-gateway.md
```diff
@@ -143,13 +144,46 @@ For detailed implementation, check out [Calendar gRPC gateway](https://github.co
...
+| GATEWAY_PROCESSOR_SERVER_HTTP_PORT     | Gateway processor server HTTP server port |
+|  GATEWAY_PROCESSOR_CLIENT_HTTP_PORT | Gateway processor client HTTP server port  |
+|  GATEWAY_HTTP_CONFIG_START            | REST endpoint to post new configuration |
+|  GATEWAY_HTTP_CONFIG_STOP             | REST endpoint to make configuration to stop |
+|  GATEWAY_HTTP_CONFIG_EVENT             | REST endpoint to send events to |
+
+
+For detailed implementation, check out [Calendar HTTP gateway](https://github.com/argoproj/argo-events/tree/eventing/gateways/rest/calendar)
+
+###### Advantage of writing a gRPC or http gateway is that you can use other languages for implementation of gateway processor server.
```
magaldima: isn't this an advantage for all types of gateways? as long as the gateway processor server has a library to interface with the gateway client.. be that NATS, Kafka, HTTP. you just need a supported protocol + infra (if needed in case of streams)

common/k8-events.go
```diff
@@ -0,0 +1,65 @@
...
+	return &corev1.Event{
+		Reason: event.Reason,
+		Type:   event.Type,
+		Action: event.Action,
+		EventTime: metav1.MicroTime{
+			Time: time.Now(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: event.Namespace,
+			Name:      event.Name + "-" + RandomStringGenerator(),
```
magaldima: can you not just use `GenerateName` here instead?

controllers/sensor/signal-filter.go
```diff
@@ -25,38 +25,121 @@ import (
...
-			return (timeFilter.Start.Before(eventTime) || timeFilter.Start.Equal(eventTime)) && eventTime.Before(timeFilter.Stop)
+func (se *sensorExecutor) filterTime(timeFilter *v1alpha1.TimeFilter, eventTime *metav1.Time) (bool, error) {
+	if timeFilter != nil {
+		currentT := time.Now().UTC()
+		se.log.Info().Str("current-time", currentT.String()).Msg("current time")
+		currentTStr := fmt.Sprintf("%d-%s-%d", currentT.Year(), int(currentT.Month()), currentT.Day())
+
+		if timeFilter.Start != "" && timeFilter.Stop != "" {
+			se.log.Info().Str("start time format", currentTStr + " " + timeFilter.Start).Msg("start time format")
+			startTime, err := time.Parse("2006-01-02 15:04:05", currentTStr + " " + timeFilter.Start)
```
magaldima: let's make the format string `2006-01-02 15:04:05` a constant since we use it more than once and seeing this may be confusing for some.

common/k8-events.go
VaibhavPage: yup, changed it in my local. I'll push some new changes with test cases

controllers/sensor/signal-filter.go
VaibhavPage: done :+1:

VaibhavPage:
removing WIP status

gateways/gateway.go
VaibhavRit: done

VaibhavPage:
ready to merge

Issue #106, opened 2018-10-16
Correct ordering of containers in Gateway pod


VaibhavPage:
**Is your feature request related to a problem? Please describe.**
Currently we don't maintain a order of startup for containers in gateway pod. In a scenario where gateway-processor container is running and received an event and tried to dispatch it to gateway transformer container but if the transformer container is not running, we loose that event.

**Describe the solution you'd like**
Add readiness probe to gateway-transformer. Only when gateway-processor is able to reach the transformer, we should proceed with processing the events

#100

Issue #104, opened 2018-10-16
Add K8 configmaps as artifact location for triggers


VaibhavPage:
**Is your feature request related to a problem? Please describe.**
Provide an ability to user to read a trigger resource from K8 configmap.

**Describe the solution you'd like**
Configmaps provide a simple K8 native storage to store trigger resources.

VaibhavPage:
#100

Issue #98, opened 2018-10-09
Add file system watcher as core gateway


VaibhavPage:
Watching changes to files/directories and taking action based on the event is a common use case.

Suggested config -

```
type fileWatcherConfig struct {
	// Path of object to watch
	Path string
	// Type of file operations to watch
	// Refer https://github.com/fsnotify/fsnotify/blob/master/fsnotify.go for more information
	Type string
}
```

VaibhavPage:
PR #100

VaibhavPage:
#

PR #117, opened 2018-10-30
WIP:  File Watcher gateway as core gateway, notification filters to storage grid gateway, k8 configmap as trigger artifact location.

VaibhavPage:
- [x] Correct ordering of gateway container startup. Gateway transformer > processor
- [x] Add K8 configmaps as artifact location for triggers
- [x] Add filters to storage grid notifications
- [x] Add file watcher as core gateway

#106 #104 #98


magaldima:
please check the unit tests, I'm seeing 
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x98 pc=0x142c643]
```

for the `github.com/argoproj/argo-events/controllers/gateway` package `TestProcessNextItem` test.

controllers/gateway/transform/transformer.go
```diff
@@ -286,3 +286,9 @@ func (toc *tOperationCtx) TransformRequest(w http.ResponseWriter, r *http.Reques
 	}
 	common.SendSuccessResponse(w)
 }
+
+// ReadinessProbe is probe to check whether server is running or not
+func (toc *tOperationCtx) ReadinessProbe(w http.ResponseWriter, r *http.Request) {
+	toc.log.Info().Msg("transformer is up and running")
```
magaldima: do we need to log a message? we can remove to clean up some logging

gateways/core/webhook/webhook.go
```diff
@@ -200,13 +200,17 @@ func (wce *webhookConfigExecutor) StopConfig(config *gateways.ConfigContext) err
 }
 
 func main() {
-	_, err := gatewayConfig.WatchGatewayEvents(context.Background())
+	err := gatewayConfig.TransformerReadinessProbe()
```
magaldima: is there any way to reduce some of this boilerplate code (possibly centralize to a util method).. I'm seeing this code and corresponding changes for all gateway types...

gateways/custom/storage-grid/main.go
```diff
@@ -61,26 +66,70 @@ type storageGridEventConfig struct {
...
 	Events []string
-	Filter StorageGridFilter
+	Filter *Filter
 	// srv holds reference to http server
 	srv *http.Server
 	mux *http.ServeMux
 }
 
+// Filter represents filters to apply to bucket nofifications for specifying constraints on objects
+type Filter struct {
```
magaldima: been thinking more about this kind of stuff.. the user needs to know the structure of all these types when he/she builds the appropriate configmap for a gateway.. don't you think this belongs in the API package somewhere? But at the very least, how do we ensure this stuff is transparent, and validated against whenever a user creates a config?

VaibhavPage: I agree that this needs to be transparent. We can add `Validate` method to `ConfigExecutor` interface to make sure that configs are validated.

VaibhavPage: Also a separate `types.go` file containing structs might help with readability

PR #118, opened 2018-11-12
Fix links of github tree references

dtaniwaki:
I found the links to github source tree in the docs are not valid.


VaibhavPage:
Thanks @dtaniwaki

PR #124, opened 2018-11-20
Updated docs and changelog for 0.6 release

VaibhavPage:


docs/custom-gateway.md
```diff
@@ -37,22 +36,35 @@ User needs to implement following interface.
...
+
+	ErrChan chan error
+
+	ShutdownChan chan struct{}
+
 	// Active tracks configuration state as running or stopped
 	Active bool
+	// Cancel is called to cancel the context used by client to communicate with gRPC server.
+	// Use it only if gateway is implemented as gRPC server.
+	Cancel context.CancelFunc
```
magaldima: this is cool but should it be restricted to use with `gRPC` server.. we should be using it for everything!

PR #128, opened 2018-11-22
- adding sensor image and imagePullSecrets as optional fields to Sens…

twiechert:
…orSpec

- adding transformer image as optional field to GatewaySpec
- creating deployment through appv1beta2 since appv1 seems to have consistently problems with imagePullSecret field. see https://github.com/kubernetes/kubernetes/issues/57801


VaibhavPage:
can we close the pr?

PR #135, opened 2018-12-01
Fixing examples

VaibhavPage:


examples/gateways/artifact.yaml
```diff
@@ -9,12 +9,13 @@ spec:
...
     containers:
     - name: "artifact-events"
-      image: "argoproj/artifact-gateway"
+      image: "argoproj/artifact-gateway:v0.6"
       imagePullPolicy: "Always"
       command: ["/bin/artifact-gateway"]
     serviceAccountName: "argo-events-sa"
   configMap: "artifact-gateway-configmap"
   eventVersion: "1.0"
+  imageVersion: "v0.6"
```
magaldima: why need this version here as well as above?

VaibhavPage: Its for the internal gateway-transformer image version. It will be removed in v0.7

magaldima: ah okay so this is just to allow v0.6 to work.

PR #136, opened 2018-12-03
Custom gateways

VaibhavPage:
- [x] AWS SNS Gateway
- [x] AWS SQS Gateway
- [x] GCP PubSub Gateway


gateways/core/calendar/start.go
```diff
@@ -43,15 +47,22 @@ func (ce *CalendarConfigExecutor) StartConfig(config *gateways.ConfigContext) {
...
 	for {
 		select {
-		case <-config.StartChan:
-			ce.GatewayConfig.Log.Info().Str("config-name", config.Data.Src).Msg("configuration is running")
-			config.Active = true
+		case _, ok := <-config.StartChan:
+			if ok {
+				ce.GatewayConfig.Log.Info().Str("config-name", config.Data.Src).Msg("configuration is running")
+				config.Active = true
+			}
```
magaldima: do you need an `else` condition if there is a failure receiving from the `StartChan`?

examples/sensors/webhook.yaml
```diff
@@ -5,14 +5,10 @@ metadata:
   labels:
     sensors.argoproj.io/sensor-controller-instanceid: argo-events
 spec:
+  serviceAccountName: argo-events-sa
+  imageVersion: "v0.6"
```
magaldima: why can't we just use `latest` here? this way, every time we bump versions and cut a tag we don't have to change all the examples? as long as we only publish the tags to docker image repo we should be fine..

gateways/core/resource/start.go
```diff
@@ -45,15 +49,22 @@ func (ce *ResourceConfigExecutor) StartConfig(config *gateways.ConfigContext) {
...
-		case <-config.StartChan:
-			ce.GatewayConfig.Log.Info().Str("config-name", config.Data.Src).Msg("configuration is running.")
-			config.Active = true
-
-		case data := <-config.DataChan:
-			ce.GatewayConfig.DispatchEvent(&gateways.GatewayEvent{
-				Src:     config.Data.Src,
-				Payload: data,
-			})
+		case _, ok := <-config.StartChan:
```
magaldima: this code blocks seems to be repeated a lot in all the gateway implementations.. can this be abstracted to a common func?

gateways/custom/aws-sns/start.go
```diff
@@ -0,0 +1,227 @@
...
+
+package aws_sns
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"sync"
+
+	"github.com/AdRoll/goamz/sns"
```
magaldima: why are you choosing this versus the aws-sdk? see: https://github.com/aws/aws-sdk-go

gateways/custom/aws-sqs/start.go
```diff
@@ -0,0 +1,160 @@
...
+limitations under the License.
+*/
+
+package aws_sqs
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/AdRoll/goamz/aws"
```
magaldima: same question about lib here.

gateways/core/calendar/start.go
VaibhavPage: If there was any failure before receiving start signal on  `StartChan` configuration state will be marked as `Error`.

gateways/core/resource/start.go
VaibhavPage: yeah, I'll create a separate PR for this.

gateways/custom/aws-sns/start.go
VaibhavPage: For the gateway we just need the structs that for receiving SNS and SQS notifications, so I added `github.com/AdRoll/goamz/sns` cause its simple to use.

PR #141, opened 2018-12-18
Refactor storage grid gateway

VaibhavPage:
The functionality remains the same but the way routes are activated and deactivated is refactored

### DIFF
--- a/gateways/custom/storagegrid/start.go
+++ b/gateways/custom/storagegrid/start.go
@@ -17,18 +17,19 @@ limitations under the License.
 package storagegrid
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/argoproj/argo-events/common"
-	"github.com/argoproj/argo-events/gateways"
-	"github.com/joncalhoun/qson"
-	"github.com/satori/go.uuid"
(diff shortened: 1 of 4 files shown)


gateways/custom/storagegrid/start.go
```diff
@@ -81,204 +115,189 @@ func filterEvent(notification *storageGridNotification, sg *StorageGridEventConf
...
-				}
-				b, err := qson.ToJSON(parsedURL)
-				if err != nil {
-					config.Active = false
-					config.ErrChan <- err
-					return
-				}
+// routeDeactivateHandler handles routes that are not active
+func (rc *routeConfig) routeDeactivateHandler(writer http.ResponseWriter, request *http.Request) {
+	rc.configExecutor.Log.Info().Str("endpoint", rc.sgConfig.Endpoint).Str("http-method", request.Method).Msg("route is not active")
```
magaldima: do you need to delete from `activeServers` here?

VaibhavPage: nope, instead of deleting it from activeServers, the code assigns http handler that returns 404 for the route. This essentially makes route deactive.

Issue #99, opened 2018-10-09
Add support for dispatching internal events on NATS streaming and Kafka


VaibhavPage:
**Is your feature request related to a problem? Please describe.**
Currently, events from gateways to sensors are sent over HTTP and there is no guarantee that event dispatched will reach to destination watcher. 

**Describe the solution you'd like**
Give an option to user to dispatch internal events over NATS streaming and Kafka so that there will be delivery guarantees on events.

PR #156, opened 2019-01-23
Nats standard and streaming support

VaibhavPage:
#99


controllers/sensor/Dockerfile
```diff
@@ -1,3 +1,3 @@
 FROM centos:7
-COPY dist/sn-controller /bin/
-ENTRYPOINT [ "/bin/sn-controller" ]
\ No newline at end of file
+COPY dist/sensor-controller /bin/
```
magaldima: thanks for doing this

controllers/sensor/validate.go
```diff
@@ -39,6 +40,27 @@ func ValidateSensor(s *v1alpha1.Sensor) error {
...
+			return fmt.Errorf("http server port is not defined")
+		}
+	case pc.NATS:
+		if s.Spec.EventProtocol.Nats.URL == "" {
+			return fmt.Errorf("nats url is not defined")
+		}
+		if s.Spec.EventProtocol.Nats.Type == "" {
+			return fmt.Errorf("nats type is not defined. either Standard or Streaming type should be defined")
+		}
+		if s.Spec.EventProtocol.Nats.Type == pc.Streaming && s.Spec.EventProtocol.Nats.ClientId == "" {
```
magaldima: this is okay, but repeating `if` statements. consider nesting of conditions to simplify these

gateways/core/webhook/start.go
```diff
@@ -117,26 +122,36 @@ func (rc *routeConfig) startHttpServer() {
...
-	common.SendErrorResponse(writer)
+	activeEndpoints[rc.wConfig.Endpoint].dataCh <- body
+	response = "request successfully processed"
+	rc.eventSourceExecutor.Log.Info().Str("endpoint", rc.wConfig.Endpoint).Str("http-method", request.Method).Str("response", response).Msg("request payload parsed successfully")
+	common.SendSuccessResponse(writer, response)
 }
 
 // StartEventSource starts a event source
 func (ese *WebhookEventSourceExecutor) StartEventSource(eventSource *gateways.EventSource, eventStream gateways.Eventing_StartEventSourceServer) error {
-	defer gateways.Recover(eventSource.Name)
```
magaldima: we're not recovering from panic anymore?

gateways/custom/storagegrid/start.go
```diff
@@ -67,28 +70,28 @@ type routeConfig struct {
 	sgConfig            *storageGrid
 	eventSource         *gateways.EventSource
 	eventSourceExecutor *StorageGridEventSourceExecutor
-	dataCh              chan []byte
-	doneCh              chan struct{}
-	errCh               chan error
 	startCh             chan struct{}
 }
 
+type endpoint struct {
```
magaldima: it seems a lot of this refactor touches every single gateway implementation and this struct seems to be pretty generic enough to exist in each separate implementation. We need to think about making our lives easier and isolating common functionality while potentially moving all the different implementations out of tree. We can take some lessons from the CNI and CSI projects in K8s space... the goal being we have an interface that all implementations must implement in order to communicate. We need to take opinionated positions at some point so we don't get lost updating everything on every PR

pkg/apis/gateway/v1alpha1/types.go
```diff
@@ -170,3 +161,26 @@ type SensorNotificationWatcher struct {
 	// Name is name of the sensor
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
 }
+
+// Dispatch protocol contains configuration necessary to dispatch an event to sensor over different communication protocols
+type EventProtocol struct {
+	Type common.EventProtocolType `json:"type" protobuf:"bytes,1,opt,name=type"`
```
magaldima: these types should never exist outside the `api` package.. please move the `EventProtocolType` to this file or another file inside the api package.

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -69,23 +69,75 @@ type Sensor struct {
...
 	DeploySpec *corev1.PodSpec `json:"deploySpec" protobuf:"bytes,3,opt,name=deploySpec"`
+
+	// EventProtocol is the protocol through which sensor receives events from gateway
+	EventProtocol *EventProtocol `json:"eventProtocol" protobuf:"bytes,4,opt,name=eventProtocol"`
+}
+
+// EventProtocol contains configuration necessary to receieve an event from gateway over different communication protocols
+type EventProtocol struct {
+	// Type defines the type of protocol over which events will be receieved
+	Type common.EventProtocolType `json:"type" protobuf:"bytes,1,opt,name=type"`
```
magaldima: okay so you reference this across gateway and sensor APIs.. you should create another api package for the event stuff (even though this package doesn't create any CRDs)

gateways/custom/storagegrid/start.go
VaibhavPage: Agreed. I'll create an issue for this and we can discuss the options to handle this.

Issue #12, opened 2018-06-01
Add top level boolean logic operator for signals


VaibhavPage:
**Is your feature request related to a problem? Please describe.**
With current Sensor spec, user can only use AND operator for multiple signals and it lacks support for complex signals circuiting. 

**Describe the solution you'd like**
Use "--" for OR operator and "-" for AND operator.

Example-
```
signals:
   -- name: time
      calender:
        interval: 10s
    - name: minioS3
      artifact:
        s3:
          bucket: hello
          endpoint: artifacts-minio.default:9000
          insecure: true
          accessKey:
            key: accesskey
            name: artifacts-minio
          secretKey:
            key: secretkey
            name: artifacts-minio
          event: s3:ObjectCreated:Put
          arn:
            partition: minio
            service: sqs
            region: us-east-1
            accountID: "1"
            resource: nats
        stream:
          nats:
            url: nats://example-nats-cluster:4222
            subject: bucketevents
   -- name: worklow-1
       resource:
         namespace: default
         group: "argoproj.io"
         version: "v1alpha1"
         kind: "Workflow"
         filter:
           prefix: scripts-bash
           labels:
             workflows.argoproj.io/phase: Succeeded 
      
```

VaibhavRit:
I feel like this feature is essential to make framework easy to use. Basically, without support for OR operator, you end up writing multiple sensors that are triggering same workflow. e.g. consider your gateway is generating 4 different configuration, and for each event we want to create a workflow, then adding these 4 signals in one sensor will cause sensor to wait on all 4 signals, definitely what we wanted.

VaibhavPage:
I feel like this feature is essential to make framework easy to use. Basically, without support for OR operator, you end up writing multiple sensors that are triggering same workflow. e.g. consider your gateway is generating 4 different configuration, and for each event we want to create a workflow, then adding these 4 signals in one sensor will cause sensor to wait on all 4 signals, definitely what we wanted.

PR #162, opened 2019-02-02
Boolean circuit for event dependencies

VaibhavPage:
#12


sensors/trigger.go
```diff
@@ -30,60 +31,126 @@ import (
...
-					"trigger failure", sec.sensor.Name, sec.sensor.Namespace, sec.controllerInstanceID, sensor.Kind, labels); err != nil {
-					sec.log.Error().Err(err).Msg("failed to create K8s event to escalate trigger failure")
-				}
-				continue
+	return false, nil
+}
+
+// canExecuteTrigger determines whether a trigger is executable based on condition set on trigger
+func (sec *sensorExecutionCtx) canExecuteTrigger(trigger v1alpha1.Trigger) bool {
+	if trigger.When == nil {
```
magaldima: so does this make this change non-breaking? for existing sensors defined without groups or `when` this will still allow the trigger to fire?

sensors/trigger.go
```diff
@@ -30,60 +31,126 @@ import (
...
+	sec.sensor.Status.CompletionCount = sec.sensor.Status.CompletionCount + 1
+
+	// create K8s event to mark the trigger round completion
+	labels[common.LabelEventType] = string(common.OperationSuccessEventType)
+	if err := common.GenerateK8sEvent(sec.kubeClient, fmt.Sprintf("completion count:%d", sec.sensor.Status.CompletionCount), common.OperationSuccessEventType,
+		"triggers execution round completion", sec.sensor.Name, sec.sensor.Namespace, sec.controllerInstanceID, sensor.Kind, labels); err != nil {
+		sec.log.Error().Err(err).Msg("failed to create K8s event to log trigger execution round completion")
+	}
+
+	// Mark all dependency nodes as active
```
magaldima: so every trigger execution resets the sensor nodes to `active` so that this sensor can be triggered again.?

PR #170, opened 2019-02-17
Git as trigger source

VaibhavPage:


pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -358,6 +360,39 @@ type URLArtifact struct {
...
+// GitArtifact contains information about an artifact stored in git
+type GitArtifact struct {
+	// Git URL
+	URL string `json:"url" protobuf:"bytes,1,name=url"`
+	// Directory to clone the repository. We clone complete directory because GitArtifact is not limited to any specific Git service providers.
+	// Hence we don't use any specific git provider client.
+	CloneDirectory string `json:"cloneDirectory" protobuf:"bytes,2,name=cloneDirectory"`
+	// Creds contain reference to git username and password
+	// +optional
+	Creds *GitCreds `json:"creds,omitempty" protobuf:"bytes,3,opt,name=creds"`
```
magaldima: I've against having the credentials stored in the sensor manifest objects for security purposes. I'd rather use a reference to a secret here.

magaldima: ahh, nevermind I see that the `GitCred` is actually `corev1.SecretKeySelector`. all good

store/git.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (g *GitArtifactReader) readFromRepository(r *git.Repository) ([]byte, error) {
+	w, err := r.Worktree()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get working tree. err: %+v", err)
+	}
+
+	pullOpts := &git.PullOptions{
+		RemoteName:        "origin",
```
magaldima: do you think it's okay to hardcode the remote name? what about if you want to read from a fork `upstream` or a different remote?

store/git.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if tag != "" {
+		tag, err := r.Tag(tag)
+		if err != nil {
+			return nil, fmt.Errorf("tag %s not found. err: %+v", tag, err)
+		}
+		refName := tag.Name()
+		return &refName, nil
+	}
+	return nil, nil
```
magaldima: do you really want to return `nil` error here? what happens if the `refName` is nil?

store/git.go
VaibhavPage: updated.

store/git.go
VaibhavPage: updated

PR #172, opened 2019-02-18
Enhance github and gitlab gateway

VaibhavPage:
1. Integrate http server within github and gitlab gateway, so that user doesn't need to spin up a separate webhook gateway to consume events.
2. Extract common webhook validate functionality


gateways/common/webhook.go
```diff
@@ -224,3 +226,24 @@ func ProcessRoute(rc *RouteConfig, helper *WebhookHelper, eventStream gateways.E
...
+
+func ValidateWebhook(endpoint, port string) error {
+	if endpoint == "" {
+		return fmt.Errorf("endpoint can't be empty")
+	}
+	if port == "" {
+		return fmt.Errorf("port can't be empty")
+	}
+
+	if !strings.HasPrefix(endpoint, "/") {
```
magaldima: do you think this is a fatal error? can't we easily handle this case by forming a correct endpoint by prepending a slash if necessary?

magaldima: there also might be standard libraries that parse both `endpoint` and `port` together to form an address.. you can look at leveraging those.

gateways/community/github/start.go
```diff
@@ -18,113 +18,219 @@ package github
...
+	}
+
+	actual := make([]byte, 20)
+	hex.Decode(actual, []byte(signature[5:]))
+
+	return hmac.Equal(signBody(secret, body), actual)
+}
+
+func validatePayload(secret []byte, headers http.Header, body []byte) error {
+	signature := headers.Get("x-hub-signature")
```
magaldima: these strings that start with `x-` can probably be made `consts`

Issue #183, opened 2019-02-25
resource gateway: panic: interface conversion: runtime.Object is nil, not *unstructured.Unstructured


hideto0710:
**Describe the bug**

I got errors below.

*resource-events*
```
starting gateway server
[2m2019-02-25T08:55:33Z[0m | info | msg: operating on event source | event-source-name:foo
[2m2019-02-25T08:55:33Z[0m | info | msg: starting to watch to resource notifications | event-source-name:foo
[2m2019-02-25T08:57:18Z[0m | warn | msg: object to watch is nil | event-source-name:foo
panic: interface conversion: runtime.Object is nil, not *unstructured.Unstructured goroutine 31 [running]: github.com/argoproj/argo-events/gateways/core/resource.(*ResourceEventSourceExecutor).listenEvents.func1(0xc420305330, 0xc4200acd20, 0xc4200a0320, 0x13604e0, 0xc420451980, 0xc420485d80, 0xc42008c3c0, 0xc4203aa1e0, 0xc4200ac780, 0xc42008c360) /Users/vpage/go/src/github.com/argoproj/argo-events/gateways/core/resource/start.go:92 +0x4f0 created by github.com/argoproj/argo-events/gateways/core/resource.(*ResourceEventSourceExecutor).listenEvents /Users/vpage/go/src/github.com/argoproj/argo-events/gateways/core/resource/start.go:79 +0x251
```
And then restarted, but not `starting to watch to resource` .

*gateway-client*
```
[2m2019-02-25T08:57:18Z[0m | error | msg: failed to receive event from stream | [31merror=[0m[31m"rpc error: code = Unavailable desc = transport is closing"[0m event-source-name:foo gateway-name:sample-resource-gateway gateway-namespace:argo-events
[2m2019-02-25T08:57:18Z[0m | info | msg: received a gateway state update notification | gateway-name:sample-resource-gateway gateway-namespace:argo-events
[2m2019-02-25T08:57:18Z[0m | info | msg: marking node phase | gateway-name:sample-resource-gateway gateway-namespace:argo-events node-id:1819460354 node-name:foo phase:Error
[2m2019-02-25T08:57:18Z[0m | info | msg: phase marked | gateway-name:sample-resource-gateway gateway-namespace:argo-events new-phase:Error node-id:1819460354 node-name:foo old-phase:Running
```

**To Reproduce**
Follow [this guide](https://github.com/argoproj/argo-events/blob/master/gateways/core/resource/install.md).

[...]

PR #185, opened 2019-02-26
Fix resource gateway

VaibhavPage:
#183


examples/sensors/resource.yaml
```diff
@@ -8,15 +8,15 @@ spec:
   deploySpec:
     containers:
       - name: "sensor"
-        image: "argoproj/sensor"
+        image: "argoproj/sensor:v0.7.4"
```
magaldima: Should this be v0.8 to be in sync with makefile?

PR #187, opened 2019-02-28
feature(): support event type in resource gateway

VaibhavPage:


VaibhavPage:
A while back, I thought of implementing a Custom Resource called `EventSource`. `eventSources`.  Basically like following,  

```
apiVersion: argoproj/v1alpha1
kind: EventSource
metadata:
    name: webhook-event-source
spec:
    eventSources:
        webhook:
            # map of config key and value.
            "index":
                port: 12000
                endpoint: "/endpoint"
            "foo":
                port: 12000
                endpoint: "/foo"
```


Although this is better than storing event sources in Configmaps, the code changes will touch almost every file that deals with gateways. So I kind of hold on it.

VaibhavPage:
@magaldima @dtaniwaki  let me know your thoughts.

dtaniwaki:
Could you tell me the advantage of defining `EventSource` over utilizing `ConfigMap` in terms of the functionality?
Maybe, we should create an issue to discuss it because it's out of the scope of this PR?

VaibhavPage:
in terms of functionality, no advantage. I would say it is an organizational change. Basically organizing the event sources in a spec so that they become part of a standard api, instead of strings as they are in configmaps. This is something we can think about after v1.0 is released.

dtaniwaki:
I see. Thanks for the explanation!
I think organizing thing is always good.

Issue #72, opened 2018-07-26
webhook signals should support TLS


magaldima:
**Is your feature request related to a problem? Please describe.**
We should support secure webhooks.

**Describe the solution you'd like**
@shrinandj outlined in #68 the following suggestions:
- Self-signed cert (the webhook signal pod will simply create it's own cert and private key)
- User provided cert and key (maybe volume mounted in the webhook signal pod?)

PR #206, opened 2019-03-06
feature(): making webhook secure

VaibhavPage:
#72


ntwrkguru:
>       # make sure to create the secret with two keys- cert and private

Is this a standard kubernetes tls secret? Or just a normal secret with `--from-file` parameters?

PR #208, opened 2019-03-07
Child resource recreation

dtaniwaki:
I implemented the following features.

- recreating pod and service for gateways
- recreating pod and service (HTTP eventProtocol) for sensors
- updating changed pod and service spec for gateways
- updating changed pod and service (HTTP eventProtocol) spec for sensors

This recreation is not graceful.

I tried to make it DRY as much as possible, but the implementation of the gateway and sensor are not the same and I wasn't able to make it perfect
Especially, I realized `deploySpec` is `Pod` in gateways while it's `PodSpec` in sensors. Is it expected?


VaibhavPage:
Actually we should change the PodSpec in sensor to Pod to keep it consistent with gateway

dtaniwaki:
@VaibhavPage Could you review this PR? Thanks!

VaibhavPage:
On it

dtaniwaki:
The CI failed in `dep ensure -vendor-only` which I didn't touch at all.

VaibhavPage:
Restarted the build.

dtaniwaki:
I'm fixing the CI failure. It's a race condition.

controllers/common/informer_test.go
```diff
@@ -0,0 +1,42 @@
...
+		SharedInformerFactory: informerFactory,
+		Queue: workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+	}
+}
+
+func TestNewPodInformer(t *testing.T) {
+	convey.Convey("Given an informer factory", t, func() {
+		factory := getInformerFactory()
+		convey.Convey("Get a new gateway pod informer and make sure its not nil", func() {
+			i := factory.NewPodInformer()
```
magaldima: Can you enhance these tests to actual create a pod and then delete and ensure the informer queues the proper owner?

dtaniwaki: Nice idea! I'll add it.

controllers/gateway/resource.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, err
+	}
+	if len(pods) == 0 {
+		return nil, nil
+	}
+	return pods[0], nil
+}
+
+// createGatewayPod creates a pod of gateway
+func (goc *gwOperationCtx) createGatewayPod() (*corev1.Pod, error) {
```
VaibhavPage: maybe change the function defn to take a pod spec as input

controllers/sensor/operator.go
```diff
@@ -90,126 +100,157 @@ func (soc *sOperationCtx) operate() error {
...
+		soc.log.Info().Str("svc-name", svc.Name).Msg("sensor service is created")
+	}
+
+	// if we get here - we know the signals are running
+	soc.log.Info().Msg("marking sensor as active")
+	soc.markSensorPhase(v1alpha1.NodePhaseActive, false, "listening for events")
+	return nil
+}
+
+func (soc *sOperationCtx) updateSensorResources() error {
```
VaibhavPage: Similar to gateway update method, I am proposing https://gist.github.com/VaibhavPage/0ba7d55df65487d590cf550075797936

controllers/sensor/resource.go
```diff
@@ -0,0 +1,130 @@
...
+
+// newSensorPod returns a new pod of sensor
+func (soc *sOperationCtx) newSensorPod() (*corev1.Pod, error) {
+	podSpec := soc.s.Spec.DeploySpec.DeepCopy()
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      soc.s.Name,
+			Namespace: soc.s.Namespace,
+		},
+		Spec: *podSpec,
```
VaibhavPage: lets change this to `Pod`

controllers/gateway/controller.go
```diff
@@ -65,20 +73,23 @@ type GatewayController struct {
...
+	rateLimiter := workqueue.NewItemExponentialFailureRateLimiter(rateLimiterBaseDelay, rateLimiterMaxDelay)
 	return &GatewayController{
 		ConfigMap:        configMap,
 		Namespace:        namespace,
 		kubeConfig:       rest,
 		log:              common.GetLoggerContext(common.LoggerConf()).Str("controller-namespace", namespace).Logger(),
 		kubeClientset:    kubernetes.NewForConfigOrDie(rest),
 		gatewayClientset: clientset.NewForConfigOrDie(rest),
-		queue:            workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+		queue:            workqueue.NewRateLimitingQueue(rateLimiter),
```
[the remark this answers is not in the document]
dtaniwaki: I think the default rate limiter is too fast to retry on failure cases. When I created a gateway with an invalid pod, it retried many times within a few seconds and generated hundreds of log lines. I think it's not reasonable to retry in milliseconds when a pod is invalid.

controllers/gateway/controller.go
```diff
@@ -115,7 +126,7 @@ func (c *GatewayController) processNextItem() bool {
 			common.LabelGatewayName: gateway.Name,
 			common.LabelEventType:   string(common.EscalationEventType),
 		}
-		if err := common.GenerateK8sEvent(c.kubeClientset, fmt.Sprintf("controller failed to operate on gateway %s, err: %+v", gateway.Name, err), common.StateChangeEventType,
+		if err := common.GenerateK8sEvent(c.kubeClientset, fmt.Sprintf("controller failed to operate on gateway %s", gateway.Name), common.StateChangeEventType,
```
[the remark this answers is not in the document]
dtaniwaki: Yeah, the limit is 128 characters which is too short to put error string in certain situations.

controllers/gateway/operator.go
```diff
@@ -86,76 +97,164 @@ func (goc *gwOperationCtx) operate() error {
...
+		return err
+	}
+
+	deleted := false
+	created := false
+
+	pod, err := goc.getGatewayPod()
+	if err != nil {
+		goc.log.Error().Err(err).Str("gateway-name", goc.gw.Name).Msg("failed to get pod for gateway")
+		goc.markGatewayPhase(v1alpha1.NodePhaseError, fmt.Sprintf("failed to get gateway pod. err: %s", err))
```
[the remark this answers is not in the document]
dtaniwaki: Do you mean something like `goc.markGatewayPhase(v1alpha1.NodePhaseError, fmt.Sprintf("failed to get gateway pod. err: %s", err.Error()))` to log wrapped error as well?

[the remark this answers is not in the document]
VaibhavPage: yeah

controllers/gateway/operator.go
```diff
@@ -86,76 +97,164 @@ func (goc *gwOperationCtx) operate() error {
...
+	err := Validate(goc.gw)
+	if err != nil {
+		goc.log.Error().Err(err).Str("gateway-name", goc.gw.Name).Msg("gateway validation failed")
+		if goc.gw.Status.Phase != v1alpha1.NodePhaseError {
+			goc.markGatewayPhase(v1alpha1.NodePhaseError, "validation failed")
+		}
+		return err
+	}
+
+	deleted := false
```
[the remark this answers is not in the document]
dtaniwaki: Thanks! I understand the idea of your proposal and it will make the flow more clear. I'll use your code with a few tweak and will update the PR soon.

controllers/sensor/resource.go
dtaniwaki: How about `PodTemplateSpec` as `Deployment` does? We don't need `Status` and `TypeMeta`.

VaibhavPage: works.

dtaniwaki: There's not `ServiceTemplateSpec` though. We need to define it with `ServiceSpec` and `ObjectMeta` by ourselves.

VaibhavPage: I guess in case of service, it would be easy if we just stick to complete `Service`. Is your argument from a view that it won't be consistent with deploySpec which is now `PodTemplateSpec` instead of complete `Pod`?

dtaniwaki: Yeah, and also I personally don't want to include unnecessary `Status` and `TypeMeta` to avoid unnecessary handling: e.g. Need to remove such fields from the spec before actually creating a service.

VaibhavPage: Ok. We can add ‘ServiceTemplateSpec’ which will contain just the serviceSpec and object meta

dtaniwaki:
I haven't updated codegen because I can't make it.

VaibhavPage:
I see. Are you able to run `sh update-codegen.sh` within `hack` folder?

dtaniwaki:
I updated the code based on the feedback, and updated codegen with only `make clientgen` and `make openapi-gen` with the changes in https://github.com/argoproj/argo-events/pull/217, also updated sensor examples as we changed the `deploySpec` type.
@VaibhavPage Would you review it again?

dtaniwaki:
I’m fixing the CI failure.

dtaniwaki:
Finally! I fixed the tests.

PR #213, opened 2019-03-08
Improve file gateway

dtaniwaki:
I made the payload of file and HDFS gateways json and usable in sensors.


dtaniwaki:
Let me change the name of the package. Directory event should be handled in this package. Maybe fsevent is a better package name.

PR #221, opened 2019-03-11
gateway/github: Enterprise support and some fixes

discordianfish:
Add baseURL and uploadURL options: This allows using the gateway with github enterprise.
Also some further fixes to make this work with GHE. Can't say if this was working with regular GH without them.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,7 +3,7 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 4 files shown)


discordianfish:
This works so far. Ran into another issue but that's independent, so the PR is ready to get merged.
Had to update go-github to includes the Name parameter when creating the webhook. Seems to be required for GHE and is reported to sometimes be required with public github too.

PR #224, opened 2019-03-12
Handle empty namespace on trigger resources

### LINKED ISSUE #214: Trigger resources without namespace can not be created
**Describe the bug**

I found omitting namespace in trigger resources causes an error in this line.
https://github.com/argoproj/argo-events/blob/a68118076fdf3e54ab63864661b25f4048327e3f/sensors/trigger.go#L230

Here's the actual error log.

```
2019-03-08T11:01:48Z | error | msg: trigger failed to execute |  error="failed to create resource object. err: an empty namespace may not be set during creation" sensor-name:hdfs-sensor trigger-name:hdfs-workflow-trigger
```

**To Reproduce**

Omit `namespace` in your trigger resources.

**Expected behavior**

https://github.com/argoproj/argo-events/blob/ac66a69bb09847b5369b4aeb74738ec43d13b6bf/pkg/apis/sensor/v1alpha1/types.go#L259
This line describes the default namespace is the namespace of the service account. However, it seems the default namespace is not applied.

I think the namespace of the service account is hard to imagine. Should we use the namespace of the sensor, or `default` namespace?

**Screenshots**

N/A

 - OS: macOS
 - Browser: N/A
 - Version: N/A

**Additional context**

N/A

dtaniwaki:
I set sensor's namespace as the default namespace of trigger resources, and added the tests.

I found OpenAPI spec is not updated when I executed `make openapi-gen`, so the PR also includes OpenAPI spec update for some other PRs.

Fixes #214

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1553,6 +1553,7 @@
     "github.com/nats-io/go-nats",
     "github.com/nats-io/go-nats-streaming",
     "github.com/nlopes/slack/slackevents",
+    "github.com/pkg/errors",
     "github.com/robfig/cron",
     "github.com/rs/zerolog",
     "github.com/satori/go.uuid",
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1291,12 +1291,13 @@
 
 [[projects]]
   branch = "release-7.0"
-  digest = "1:6e37d35d68d68b1368b803617d346633830eb13b50120f68ead2c029c05eeea0"
+  digest = "1:7336dcb25cac9940ac84ea6c59c4c36d1b6b6ca4a3aba38d6d8c74bc53699a40"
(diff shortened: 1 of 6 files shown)


dtaniwaki:
I think we should verify codegen in CI to avoid unexpected diff in another PR.

VaibhavPage:
Agreed.

VaibhavPage:
Getting

Failures:
  * /home/travis/gopath/src/github.com/argoproj/argo-events/sensors/trigger_test.go 
  Line 264:
  Expected: nil
  Actual:   'pods "my-pod" not found'
  * /home/travis/gopath/src/github.com/argoproj/argo-events/sensors/trigger_test.go 
  Line 277:
  Expected: nil
  Actual:   'pods "my-pod" not found'

PR #225, opened 2019-03-12
Update docs

VaibhavPage:
1. Update docs. Add installation steps for each gateway
2. Update examples
3. Update minio gateway so that it listens to multiple events


gateways/community/aws-sns/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+### Why is there webhook in the gateway?
+Because one of the ways you can receive notifications from SNS is over http. So the gateway runs a http server internally.
+Once you create a new gateway configmap or add a new event source entry in the configmap, the gateway will register the url of the server on AWS.
+All notifications for that topic will then be dispatched by SNS over to the endpoint specified in event source.
+
+### Example
+
+```yaml
```
magaldima: is there anyway to inject this from the examples file directly? that way you remove the necessity of keeping both these sources in sync.. less work maintaining this and more consistency for user experience.

gateways/community/aws-sns/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Setup
+
+**1. Install Gateway**
+
+We are installing gateway before creating configmap. Because we need to have the gateway pod running and a service backed by the pod, so 
+that we can get the URL for the service. 
+
+```bash
+kubectl -n argo-events create -f https://raw.githubusercontent.com/argoproj/argo-events/master/examples/gateways/aws-sns.yaml
```
magaldima: should these commands follow some kind of semantic versioning for the components? this points to `master` branch but probably references the `v0.8.3` tag in many places across images and components. This mandates that all examples, docs get updated everytime we have a release.. very burdensome

Makefile
```diff
@@ -17,7 +17,7 @@ override LDFLAGS += \
 #  docker image publishing options
 DOCKER_PUSH?=true
 IMAGE_NAMESPACE?=argoproj
-IMAGE_TAG?=v0.8
+IMAGE_TAG?=v0.8.3
```
dtaniwaki: I'd prefer `latest` as the default value so we don't have to update this value every time when we release new versions. What do you think?

Makefile
```diff
@@ -263,6 +263,16 @@ sqs-image: sqs-linux
 	docker build -t $(IMAGE_PREFIX)aws-sqs-gateway:$(IMAGE_TAG) -f ./gateways/community/aws-sqs/Dockerfile .
 	@if [ "$(DOCKER_PUSH)" = "true" ] ; then  docker push $(IMAGE_PREFIX)aws-sqs-gateway:$(IMAGE_TAG) ; fi
 
+slack:
```
dtaniwaki: This is not the problem of this PR, but how about putting Makefile in each gateway and include them by something like `include gateways/community/slack/Makefile` in the main `Makefile`?

VaibhavPage: I will create a separate pr for this.

gateways/community/aws-sns/README.md
VaibhavPage: I don't that's possible. Maybe the better way is just to reference them instead of actually posting these.

gateways/community/aws-sns/README.md
VaibhavPage: I am going to remove these `kubectl create` commands and instead just reference the respective files in the example folder.

PR #227, opened 2019-03-12
manage sensor status nodes

VaibhavPage:
1. update sensor status nodes when dependencies are updated
2. rename `deploySpec` to `template`
3. rename `serviceSpec` to `service`
4. update examples


VaibhavPage:
@magaldima @dtaniwaki can you review the pr?

controllers/sensor/operator.go
```diff
@@ -120,15 +120,15 @@ func (soc *sOperationCtx) createSensorResources() error {
 	err := ValidateSensor(soc.s)
 	if err != nil {
 		soc.log.Error().Err(err).Msg("failed to validate sensor")
-		errors.Wrap(err, "failed to validate sensor")
```
dtaniwaki: Sorry, it was my mistake!

sensors/event-handler.go
```diff
@@ -141,6 +141,30 @@ func (sec *sensorExecutionCtx) processUpdateNotification(ew *updateNotification)
 			}
 		}
 
+		// delete old status nodes if any
+	statusNodes:
+		for _, statusNode := range sec.sensor.Status.Nodes {
```
dtaniwaki: If I understand this logic correctly, it looks it goes to infinite loop once it finds an existing node.
Could you extract this logic to a method and write simple unit tests to make sure the behavior?

docs/sensor-protocol.md
```diff
@@ -334,7 +334,7 @@ SensorSpec represents desired sensor state
 | ----- | ---- | ----- | ----------- |
 | dependencies | [EventDependency](#github.com.argoproj.argo_events.pkg.apis.sensor.v1alpha1.EventDependency) | repeated | Dependencies is a list of the events that this sensor is dependent on. |
 | triggers | [Trigger](#github.com.argoproj.argo_events.pkg.apis.sensor.v1alpha1.Trigger) | repeated | Triggers is a list of the things that this sensor evokes. These are the outputs from this sensor. |
-| deploySpec | [k8s.io.api.core.v1.PodSpec](#k8s.io.api.core.v1.PodSpec) | optional | DeploySpec contains sensor pod specification. For more information, read https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#pod-v1-core |
+| deploySpec | [k8s.io.api.core.v1.PodSpec](#k8s.io.api.core.v1.PodSpec) | optional | Template contains sensor pod specification. For more information, read https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#pod-v1-core |
```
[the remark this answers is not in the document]
VaibhavPage: yeah, it needs to be `Template`. I'll update it

controllers/sensor/operator.go
VaibhavPage: no worries :)

sensors/event-handler.go
VaibhavPage: Sure.

VaibhavPage: Added the test. Because this is a `continue` stmt instead of `goto`, it won't go into infinite loop.

dtaniwaki: Ah, that's why! I read it wrong. but writing some tests is still good, thanks!

PR #228, opened 2019-03-13
Fix operator tests

dtaniwaki:
I'm trying to fix the tests, which fail in CI but succeed in my local environment.


VaibhavPage:
Looks like CI  succeeded.

PR #229, opened 2019-03-13
Fix goroutine error in test

dtaniwaki:
[The CI](https://travis-ci.org/argoproj/argo-events/builds/505797464) failed because goroutine is still waiting for the channel when the context is gone.

```
panic: Fail in goroutine after TestEventSources has completed [recovered]
	panic: Fail in goroutine after  has completed
```

I hope this fix solves the problem.


dtaniwaki:
This fix is a little bit wrong. I'll update the code now.

PR #230, opened 2019-03-13
Support applying parameters for complete trigger spec

VaibhavPage:
This feature will allow replacing anything in trigger template on the fly. It will be very helpful in cases where triggger workflow is stored on S3 or Git and user wants to figure out the url, bucket or repo name based on the event payload. Also to make it easier for user, this change would differentiate parameters for trigger template and for workflow template.


dtaniwaki:
I'm not sure when we use `WorkflowParameters` over `TemplateParameter`. Could you provide some examples in the PR?

VaibhavPage:
Sure. I'll update the PR with the examples

VaibhavPage:
For now consider this example , 
```


# The event payload for this example =
#
#  {
#   "name": "trigger-wf-1",
#   "namespace": "argo-events",
#   "bucket": "mybucket",
#   "key": "hello.yaml",
#   "image": "docker/busybox",
#  }
#
# The example refers to workflow stored on a S3 bucket. With normal sensor spec, you would hard-code the S3 bucket name, object key and other things in trigger.
# But user would like to have dynamic trigger spec where depending on the data in the payload.
# The example data payload contains information like bucket, key, name, namespace which is used to replace the placeholders in trigger spec.
# The image in payload will be used to replace the image in the workflow.
#
kind: Sensor
metadata:
  name: webhook-with-context-sensor
  labels:
spec:
  deploySpec:
    spec:
      containers:
        - name: "sensor"
          image: "argoproj/sensor"
          imagePullPolicy: Always
  dependencies:
    - name: "webhook-gateway:foo"
  eventProtocol:
    type: "HTTP"
    http:
      port: "9300"
  triggers:
      # Apply parameter at template
      templateParameters:
        - src:
            event: "webhook-gateway:foo"
            path: name
          dest: name
        - src:
            event: "webhook-gateway:foo"
            path: namespace
          dest: resource.namespace
        - src:
            event: "webhook-gateway:foo"
            path: bucket
          dest: resource.source.s3.bucket.name
        - src:
            event: "webhook-gateway:foo"
            path: key
          dest: resource.source.s3.bucket.key
      # Apply parameter for workflow
      workflowParameters:
        - src:
            event: "webhook-gateway:foo"
            path: image
          dest: spec.templates.0.container.image
      template:
        name: THIS_WILL_BE_REPLACED_BY => name in the payload
        resource:
          namespace: THIS_WILL_BE_REPLACED_BY => namespace in the payload
          group: argoproj.io
          version: v1alpha1
          kind: Workflow
          source:
            s3:
              bucket:
                name: THIS_WILL_BE_REPLACED_BY => bucket in the payload
                key: THIS_WILL_BE_REPLACED_BY => key in the payload
              endpoint: minio-service.argo-events:9000
              insecure: true
              accessKey:
                key: accesskey
                name: artifacts-minio
              secretKey:
                key: secretkey
                name: artifacts-minio

```

dtaniwaki:
I think I understand which is applied where. So, if is the workflow parameter applied only for workflow? What happened if a user wants to trigger a pod?

VaibhavPage:
You can apply `workflowParameters` on any kubernetes resource. I now realize `workflowParameters` may not be the apt name :sweat_smile: , may it should be renamed to `actionParameters` or something more intuitive.

dtaniwaki:
Yeah, The name `WorkflowParameter` may confuse users.
How about `ResourceParameter`?

VaibhavPage:
`ResourceParameter` is already the name of the struct. https://github.com/argoproj/argo-events/blob/be258c6085528c1945b43a1cb2e8f86a58389d86/pkg/apis/sensor/v1alpha1/types.go#L202:6

VaibhavPage:
Or maybe we can rename `ResourceParameter` struct to `Parameter` and have `WorkflowParameter` renamed to `ResourceParameter`

dtaniwaki:
Sounds good. Maybe, `TriggerParameter` for the struct name because it's a parameter of `Trigger`?

VaibhavPage:
I am planning to get rid of `ResourceObj` and `resource` key within the trigger, because we are not triggering anything except a resource. Also will replace custiom gvk struct with k8s gvk struct

VaibhavPage:
@magaldima @dtaniwaki updated the pr. please review

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -238,8 +238,8 @@ type ResourceParameter struct {
 	Dest string `json:"dest" protobuf:"bytes,2,name=dest"`
 }
 
-// ResourceParameterSource defines the source for a resource parameter from a event event
-type ResourceParameterSource struct {
+// ParameterSource defines the source for a parameter from a event event
+type ParameterSource struct {
```
dtaniwaki: Should it be `TriggerParameterSource` if we follow the name convention around here?

VaibhavPage: Yup, I will update it.

hack/generate-proto.sh
```diff
@@ -37,8 +37,8 @@ APIMACHINERY_PKGS=(
 
 dist/go-to-protobuf \
     --logtostderr \
+    --only-idl=true \
     --go-header-file=${PROJECT_ROOT}/hack/custom-boilerplate.go.txt \
     --packages=$(IFS=, ; echo "${PACKAGES[*]}") \
     --apimachinery-packages=$(IFS=, ; echo "${APIMACHINERY_PKGS[*]}") \
-    --proto-import=${PROJECT_ROOT}/vendor
```
dtaniwaki: Is the trailing backslash expected?

PR #241, opened 2019-03-19
Configure hook before processing route

discordianfish:
This prevents possible races between PostActive and RouteActiveHandler


VaibhavPage:
This pr is stale due #243 .Closing it.

PR #243, opened 2019-03-20
Refactor webhook route config

VaibhavPage:
1. Get rid of `Configs` map from RouteConfig. Make RouteConfig as an interface.
2. Fix github gateway issue introduced by #221 

This PR does not contain any breaking changes to spec.


VaibhavPage:
@magaldima @dtaniwaki please review the pr.

dtaniwaki:
> Get rid of Configs map from RouteConfig. Make RouteConfig as an interface.

@VaibhavPage Could you tell me the motivation of this interface change to start reviewing?

> Fix github gateway issue introduced by #221

By the way, I think this fix should've been separated for easier review and future reference. I prefer keeping PRs as small and independent as possible.

VaibhavPage:
Yeah, PRs have been big for a while, I will try to break these in smaller pieces in future 😄 

In webhook like gateways, we had `RouteConfig` which was a map basically used to hold gateway specific interfaces that are separate from common webhook fields. So to extract the values out of RouteConfig, we had cast these (possibility for nil pointer) interfaces into concrete structs and was needed to maintain label keys so that we can store these interfaces against them. 

Making RouteConfig as interface now mandates all webhook like gateways to implement it and the implementing struct can contain the extra fields which now can be accessed directly in the methods like `RouteHandler`.

Makefile
```diff
@@ -35,9 +35,9 @@ endif
 
 # Build the project images
 .DELETE_ON_ERROR:
-all: sensor-linux sensor-controller-linux gateway-controller-linux gateway-client-linux webhook-linux calendar-linux resource-linux artifact-linux file-linux nats-linux kafka-linux amqp-linux mqtt-linux storage-grid-linux github-linux hdfs-linux gitlab-linux sns-linux sqs-linux pubsub-linux trello-linux
+all: sensor-linux sensor-controller-linux gateway-controller-linux gateway-client-linux webhook-linux calendar-linux resource-linux artifact-linux file-linux nats-linux kafka-linux amqp-linux mqtt-linux storage-grid-linux github-linux hdfs-linux gitlab-linux sns-linux sqs-linux pubsub-linux
```
dtaniwaki: Is removing of the trello gateway intended for this PR?

gateways/common/webhook.go
```diff
@@ -73,16 +73,20 @@ type activeServer struct {
...
+// Route contains common information for a route
+type Route struct {
+	Webhook     *Webhook
+	Logger      *zerolog.Logger
+	StartCh     chan struct{}
+	EventSource *gateways.EventSource
+}
+
+// RouteConfig is an interface to manage  the configuration for a route
+type RouteConfig interface {
```
dtaniwaki: Is there any way to easily identify this `RouteConfig` is an interface and the other `RouteConfig`s defined in gateways are struct? Maybe rename this `RouteConfigInterface`?

Makefile
VaibhavPage: Yeah. The library for trello doesn't have solid support for webhooks. e.g. it lets you create webhook but doesn't let you delete it which is a problem when you want to stop an event source. Also, it doesn't take `id` at the time of creation which could have been used later to check if the hook is already created, especially in the case where gateway pod has restarted. 

It would be in vain to refactor trello gateway for RouteConfig interface. So I got rid of it.

If the trello library matures and support to delete hooks, we can add it later.

gateways/common/webhook.go
VaibhavPage: Not sure if we want `Interface` in the name, how about `RouteManager`

dtaniwaki: Sound good 👍

Issue #244, opened 2019-03-20
Consistent versioned examples


magaldima:
**Is your feature request related to a problem? Please describe.**
Discrepancies between the examples and certain release versions are causing lots of issues for people. This makes it extremely difficult to develop when there are breaking changes and is especially cumbersome because the gateway configs live in an unstructured `ConfigMap` instead of a versioned API. 

**Describe the solution you'd like**
My proposal is to build an automated test suite that scans over all folders in the examples to ensure they adhere to the requirements for each specific gateway/sensor configuration. In addition, I would like to move all the `community` gateways out of tree into a separate repo along with the change to make all the configurations live as a versioned `EventSource` CRD. This must be a requirement to achieve stability.

**Describe alternatives you've considered**
Let's start with a simple script that scans the examples and validates them against some structured API. This can be for all objects involved: Gateways, Sensor, and the gateway config maps. 

**Additional context**

VaibhavPage:
I agree with your proposal. We already have a validation test that basically scans the example folders and validate gateway and sensor spec/configuration. But what we lack is test for gateway configmaps as there is no single validate method (This is where EventSource CRD will be helpful). 

Moving out community gateways into a separate repo (maybe something called `argo-events-gateways`) should be a priority as well. 

My proposal is to get #225 merged with suggested changes, drop the v0.9 tag so that it will resolve any outstanding patch confusions and work towards `EventSource CRD` for v1.0

VaibhavPage:
@spk83 @dtaniwaki thoughts?

dtaniwaki:
I agree with it too. I’m going to implement e2e tests for basic cases and will make it usable for community gateway tests, maybe by using the examples.

VaibhavPage:
#251

PR #249, opened 2019-03-21
E2e tests

dtaniwaki:
I implemented E2E tests to check whether the basic case still works in all new PRs. 

The e2e tests create the following 2 types of namespaces and clean them up after test execution.
- Namespace for controllers.
- Namespace for actual test gateways and sensors.

`hack/e2e/run-e2e.sh` creates a new namespace for controllers with randomly generated name and use the namespace to execute actual e2e test cases, which create another namespace per execution. If you want to reuse the namespace for controllers, you can manually execute `hack/e2e/setup-e2e.sh` and run your e2e tests in the namespace.

I also introduced [`kind`](https://github.com/kubernetes-sigs/kind) to run this e2e test a kubernetes cluster in Docker including Travis CI. As the kubernetes version is set by env, we can also run the e2e tests in multiple versions of Kubernetes clusters if we want.

As a test case execution is isolated by namespace, I believe it's easy to add more test cases like [the one we discussed about examples](https://github.com/argoproj/argo-events/issues/244).

Please try it out by just executing `make kind-e2e` after installing kind by `go get sigs.k8s.io/kind` and building images by `make build-e2e-images`.


dtaniwaki:
I finally get it worked in CI!

hack/e2e/kind-run-e2e.sh
```diff
@@ -0,0 +1,30 @@
...
+IMAGE_PREFIX=${IMAGE_PREFIX:-argoproj/}
+IMAGE_TAG=${IMAGE_TAG:-latest}
+
+kind create cluster --name $CLUSTER_NAME --image $KUBERNETES_VERSION
+export KUBECONFIG="$(kind get kubeconfig-path --name=$CLUSTER_NAME)"
+kubectl cluster-info
+
+function cleanup {
+  if [[ -z "$KEEP_CLUSTER" ]]; then
+    echo "* Cleaning up the e2e cluter..."
```
magaldima: Not a big deal but mispelling here.. I think you want “cluster”

PR #251, opened 2019-03-22
WIP: Event source crd 

VaibhavPage:
Just started with a draft version of EventSource CRD

The CRD basically contains `Source` and `Type`.

`Source` basically contains the actual EventSource for different gateways.
`Type` is the type of EventSource that the CR will contain.


gateways/event-sources.go
```diff
@@ -246,8 +262,8 @@ func (gc *GatewayConfig) stopEventSources(configs []string) {
 }
 
 // manageEventSources syncs registered event sources and updated gateway configmap
-func (gc *GatewayConfig) manageEventSources(cm *corev1.ConfigMap) error {
-	eventSources, err := gc.createInternalEventSources(cm)
+func (gc *GatewayConfig) manageEventSources(es *esv1alpha1.EventSource) error {
```
magaldima: so from my understanding, CRDs are meant to represent infrastructure on k8s in the form of more primitive APIs.. it seems like this `EventSource` only has relevance/context when paired with a `Gateway`. Each event source doesn't create system state via resources, it does so inside of the gateway implementation, thus introducing the possibility that desired state != declared state. What about if the event sources themselves were actually pods? i think we can take a note from Knative here on this one.. check out [this](https://github.com/knative/docs/tree/master/docs/eventing/sources) and pay special attention to the [container event sources](https://github.com/knative/eventing-sources/blob/master/pkg/apis/sources/v1alpha1/containersource_types.go). They use the concept of a `sink` inside the CRD to refer a some object that resolves to a domain name to use as the target for the cloudEvents client.

magaldima: also following the above paradigm allows us to outsource the various event sources out of tree! in this implementation PR, you can't do that since they are defined as event source types in the API!

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -0,0 +1,410 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+// SourceType is the type of event source
+type SourceType string
+
+// Possible values for event source
+// These must match the field names from `Source`
+const (
+	SourceTypeWebhook     SourceType = "webbook"
```
magaldima: not sure this is a good idea.. see above comment. this makes it very hard to generalize this object/API. with these specific types, one can't create an anonymous event source without breaking the API!

gateways/eventing.proto
```diff
@@ -25,7 +25,7 @@ message EventSource {
     // The event source name.
     string name = 2;
     // The event source configuration value.
-    string data = 3;
```
dtaniwaki: Are we making a breaking change in the protocol?

gateways/config.go
```diff
@@ -92,7 +95,7 @@ type EventSourceData struct {
 	// Src contains name of the event source
 	Src string `json:"src"`
 	// Config contains the event source
-	Config string `json:"config"`
```
dtaniwaki: Is this change necessary? We can just set a marshalled string when it's constructed. Or maybe we can create a converting method from `EventSourceData` to `EvenSource` proto and marshal to the string there.

VaibhavPage:
Agree with you guys that declaring all event sources in a single CR is not the correct way. 
@magaldima I looked at how Knative uses `ContainerSource` but it looks like it ties the `ContainerSource` to specific sink. What if I want to define a `Webhook` CR that will be used by multiple gateways, each sending events to different targets. The CR should only define the configuration and let gateway-client take care of sending the events to sensors. 
Another way is to create a CR for each event source. The gateway spec will include fields such as `Group`, `Version` and `Resource` to identify the resource and use `dynamicinformer` to listen to addition, updates and deletion. The CR spec can live in anywhere. 
We can think of adding `Sink`s to gateway spec later which `gateway-client` can use.

magaldima:
> What if I want to define a `Webhook` CR that will be used by multiple gateways, each sending events to different targets. The CR should only define the configuration and let gateway-client take care of sending the events to sensors.

It sounds like these EventSources, instead of being implemented as pure CloudEvent clients, they should be implemented as CloudEvents `publishers` whereas the Gateways should be implemented as CloudEvents `subscribers`. If this is the case, then I would argue over the possible introduction of a 4th CRD: `Channel` that abstracts the underlying pub/sub mechanism. Given this proposal:
- there can be multiple channels (independent pub/sub systems)
- there can be a many-1 mapping from `EventSource` -> `Channel`
- there can be a 1-many mapping from `Channel` -> `Gateway`

My only question with this approach is: how does the gateway's spec change? Are gateways still going to be tied to a particular type of event (i.e. webhook, sns, pubsub)? Probably not - that's the purpose of this work. So if they have no type, does having multiple gateway's make sense here? I guess, maybe for a weak multi-tenant solution..

VaibhavPage:
So basically gateway servers now are what CloudEvents publishers will be and gateway client the CloudEvents subscriber. So I get that, if you move gateway-server/CloudEvents publisher out of the gateway, the gateway will become generic with only the client left.

From my understanding, we will have a separate CRD for CloudEvents publisher that will contain the specific `EventSource` configurations.  And if a CloudEvents subscriber wants to listen to events, a user will create a `Channel`. Now, I am guessing CloudEvents publisher will listen to `Channels` because it is the one who is going to sink events to `Channel`. And that's where 1-to-many mapping of event source comes into the picture. Later if gateway client wants to opt out of events, the user will delete the `Channel` and `CloudEvents publisher` removes that channel as a sink from that `event source`. 

And we can use the same `Channel` for interactions between `gateway client/CloudEvents subscriber` and `Sensor`. Although I am inclined to use a default `Channel` backed by HTTP for communication between `Publisher` and `Subscriber` because, in case of Stream Publishers, it won't make sense to use `Nats` or `Kafka` backed `Channel` 😄 .

dtaniwaki:
Having Channel CRD sounds good. However, I’m afraid that it makes harder for users if we need more CRs to run only one event driven workflow. Is it possible to generate a Channel CR via a Gateway CR so we don’t have to create this resource by default?

VaibhavPage:
tbh, looking at it, this whole thing started to feel a bit overengineered. I think we already have pieces in places in terms of publisher and subscriber as gateway-client and gateway-server. All we really need is CRDs to express different `EventSources`, which is very straightforward to achieve. Do we really need a standalone `Publisher` and then different `Channels`, figuring out who watches what, the mappings and how to manage them etc. Not sure the value proposition here.

dtaniwaki:
Maybe this is out of this topic, but do we really need a gateway server pod in `Gateway` CR? If it communicates via a specific protocol, I think we can move out the gateway server from the CR and let a user create a gateway server by their own method such as `Deployment`, `Statefulset`, or etc. Then, we don't have to care about what event source is used. As for our core gateway servers, we can utilize `EventSource` CR if needed.

VaibhavPage:
We actually don't need gateway-server to be part of the `Gateway` CR. As you said, it can live separately. This will make things easy to manage. The `Gateway` spec will additionally contain the address of the server to connect to, and this was the `Channel` bit. We actually don't need to have a `Channel` which is backed by some protocol, we can just be opinionated about protocol and use gRPC.  I also think we should add `Group`, `Kind` and `Version` in `Gateway` spec. This is for `EventSource` CR. This will let gateway-client listen to whatever CR user has created for `EventSource`. The gateway-client will blindly watch the `EventSource` CR and pass any changes to gateway server.

Issue #242, opened 2019-03-19
Support Backoff Feature for Calendar Schedule/Interval Job


keaxu-exp:
**Is your feature request related to a problem? Please describe.**
Feature Request

**Describe the solution you'd like**
Backoff for scheduling. 
For instance, if a workflow takes 2 mins to run, but scheduled every minute. The second run is waiting until first run complete.

Configurable execution policy. 
e.g. Users are able to configure that if first run failed, when it's time to exec second run, is that gonna be drop on the floor, or run it anyway.

VaibhavPage:
#60

VaibhavPage:
I am thinking of adding following fields to a trigger 
```
# optional
policy:
    # define a linear or exponential backoff
    backoff:
        steps:
       duration:
       jitter:
      factor:
   # Becasue we support triggering any K8s resource, the only way to determine a resource is in either success or failure stage is to look at the labels on the resource.
    criteria:
        success:
            map[string]string
        failure:
            map[string]string
    # determines whether sensor should transition to error state if the backoff times out and yet the resource neither transitioned into success or failure.
    errorOnBackoffTimeout: false
```

And in sensor spec, 

`continueOnFailedRound: false
`

will be set to determine whether the sensor should transition into error if previous round of triggers were unsuccessful

keaxu-exp:
two questions:
- what does "jitter" do ?
- since "duration" and "factor" show there, as I understand correctly, this is exceptional backoff ? 
so if duration is 1s and factor is 2, then the backoff time will be 
```
1s 2s 4s 8s 16s .... util exhausted threshold ?
```

VaibhavPage:
1. `jitter` is basically the randomness. Have a read at https://stackoverflow.com/questions/46939285/why-is-random-jitter-applied-to-back-off-strategies
You don't need to set this at all and it is not required in this case. It is rather there for fun.

2. The backoff can be either linear or exponential. If you set factor to 1, then it is linear else it is exponential.

davidboren:
This might be a very dumb question but I just want to make sure I understand.  In the backoff scenario lined out by @keaxu-exp above, where the first workflow is still running and the second is waiting on that resource, does that translate to referencing the label AND status on the running workflow in the failure section like so?
```yaml
        failure:
            "some-workflow-label": "Running"
```
Or am I missing some other obvious pattern?

VaibhavPage:
[...]

PR #256, opened 2019-03-29
add trigger backoff and execution policy

VaibhavPage:
#242


examples/sensors/trigger-backoff.yaml
```diff
@@ -0,0 +1,111 @@
...
+        # Because the sensor is able to trigger any K8s resource, it determines the resource state by looking at the resource's labels.
+        policy:
+          # Backoff before checking the resource labels
+          backoff:
+            # Duration is the duration in nanoseconds
+            duration: 1000000000 # 1 second
+            # Duration is multiplied by factor each iteration
+            factor: 2
+            # The amount of jitter applied each iteration
+            jitter: 0.1
```
keaxu-exp: what 'jitter' actually do ? not fully understand here.

examples/sensors/trigger-backoff.yaml
```diff
@@ -0,0 +1,111 @@
...
+        name: trigger-1
+        # Policy to configure backoff and execution criteria for the trigger
+        # Because the sensor is able to trigger any K8s resource, it determines the resource state by looking at the resource's labels.
+        policy:
+          # Backoff before checking the resource labels
+          backoff:
+            # Duration is the duration in nanoseconds
+            duration: 1000000000 # 1 second
+            # Duration is multiplied by factor each iteration
+            factor: 2
```
keaxu-exp: So it's exponential backoff, am I understand correct ?

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -255,6 +265,42 @@ type TriggerParameterSource struct {
...
+
+	// The amount of jitter applied each iteration
+	Jitter float64 `json:"jitter" protobuf:"bytes,3,opt,name=jitter"`
+
+	// Exit with error after this many steps
+	Steps int `json:"steps" protobuf:"bytes,4,opt,name=steps"`
+}
+
+// TriggerPolicyCriteria defines the criteria to decide if a resource is in success or failure state.
+type TriggerPolicyCriteria struct {
```
magaldima: how are these used? it seems like both are labels? maybe rename to to include that fact? also, what is the use case for having this?

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -266,15 +312,18 @@ type SensorStatus struct {
...
 
 	// Nodes is a mapping between a node ID and the node's status
 	// it records the states for the FSM of this sensor.
 	Nodes map[string]NodeStatus `json:"nodes,omitempty" protobuf:"bytes,5,rep,name=nodes"`
+
+	// TriggerCycleCount is the count of sensor's trigger cycle runs.
+	TriggerCycleCount int32 `json:"triggerCycleCount,omitempty" protobuf:"varint,6,opt,name=triggerCycleCount"`
+
+	// TriggerCycleState is the status from last cycle of triggers execution.
+	TriggerCycleStatus TriggerCycleState `json:"triggerCycleStatus" protobuf:"bytes,7,opt,name=triggerCycleStatus"`
```
magaldima: can we also store the `lastCycleTime`?

sensors/trigger.go
```diff
@@ -213,15 +230,78 @@ func (sec *sensorExecutionCtx) executeTrigger(trigger v1alpha1.Trigger) error {
...
+		Jitter:   trigger.Policy.Backoff.Jitter,
+	}, func() (bool, error) {
+		obj, err := resourceInterface.Get(name, metav1.GetOptions{})
+		if err != nil {
+			sec.log.WithError(err).WithField("resource-name", obj.GetName()).Error("failed to get triggered resource")
+			return false, nil
+		}
+
+		labels := obj.GetLabels()
+		if labels == nil {
```
magaldima: what about if the policy criteria is empty and i just care about obj creation?

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -255,6 +265,42 @@ type TriggerParameterSource struct {
...
+// TriggerPolicy dictates the policy for the trigger retries
+type TriggerPolicy struct {
+	// Backoff before checking resource state
+	Backoff Backoff `json:"backoff" protobuf:"bytes,1,opt,name=backoff"`
+
+	// Criteria to check resource state
+	Criteria *TriggerPolicyCriteria `json:"criteria" protobuf:"bytes,2,opt,name=criteria"`
+
+	// ErrorOnBackoffTimeout determines whether sensor should transition to error state if the backoff times out and yet the resource neither transitioned into success or failure.
+	ErrorOnBackoffTimeout bool `json:"errorOnBackoffTimeout" protobuf:"bytes,3,opt,name=errorOnBackoffTimeout"`
```
dtaniwaki: Is there any case that NO error on timeout is desired?

pkg/apis/sensor/v1alpha1/types.go
VaibhavPage: Basically, the labels defined in this struct are used to figure out if the resource has reached success state or it failed. We are using labels because the sensor triggers an unstructured object, so it is not possible to simply get the particular field that is supposed to reflect the state of the resource as you would normally do in a structured K8s object. In case of workflows, the controller sets the labels like `workflows.argoproj.io/phase: Succeeded`  or `workflows.argoproj.io/phase: Failed` to mark workflow as successful or failed.

I can rename the struct to `TriggerStateLabels`.  

The use case is that if the sensor has multiple triggers to fire but the execution of each trigger must depend on the success or failure of the previous trigger.

pkg/apis/sensor/v1alpha1/types.go
VaibhavPage: 👍

sensors/trigger.go
VaibhavPage: That is the default behavior. If you don't provide the policy, the sensor will just create the trigger object and move to next one if any

pkg/apis/sensor/v1alpha1/types.go
VaibhavPage: I guess it is useful when a failed trigger doesn't necessarily mean that next one can't proceed. e.g. If you have a trigger that basically writes to a S3 bucket and the second trigger expects the data to be available in the bucket but also has a fallback mechanism. In that case, you would want to set back off on the first trigger to wait till it goes in success state. If after backoff, it is not possible to figure out the state of the triggered resource or it failed, you still want second trigger object to get created

PR #285, opened 2019-05-17
Setup lint

dtaniwaki:
I added linter to keep the quality and readability of the code.

### DIFF
--- a/gateways/core/resource/config.go
+++ b/gateways/core/resource/config.go
@@ -39,8 +39,6 @@ type resource struct {
 	Namespace string `json:"namespace"`
 	// Filter is applied on the metadata of the resource
 	Filter *ResourceFilter `json:"filter,omitempty"`
-	// Version of the source
-	Version string `json:"version"`
 	// Group of the resource
 	metav1.GroupVersionKind `json:",inline"`
 	// Type is the event type. Refer https://github.com/kubernetes/apimachinery/blob/dcb391cde5ca0298013d43336817d20b74650702/pkg/watch/watch.go#L43
--- a/test/e2e/common/client.go
+++ b/test/e2e/common/client.go
@@ -91,7 +91,7 @@ func (clpl *E2EClient) ForwardServicePort(tmpNamespace, podName string, localPor
 	}
 
 	path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", tmpNamespace, podName)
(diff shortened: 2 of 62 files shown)


dtaniwaki:
Although I fixed most of the lint error in the current code, I wasn't able to fix the following 2 error because I don't know what the expected behavior is.

```
sensors/trigger.go:86:4: SA4004: the surrounding loop is unconditionally terminated (staticcheck)
                        return false
                        ^
sensors/trigger.go:94:4: SA4004: the surrounding loop is unconditionally terminated (staticcheck)
                        return true
                        ^
```

Could someone give me a help?

test/e2e/common/client.go
```diff
@@ -121,7 +121,6 @@ L:
 			break L
 		case <-readyChan:
 			break L
-		default:
```
dtaniwaki: Removing this `default` changes the behavior. I think the current code doesn't wait 10 seconds because of this `default`.

gateways/core/resource/config.go
```diff
@@ -39,8 +39,6 @@ type resource struct {
 	Namespace string `json:"namespace"`
 	// Filter is applied on the metadata of the resource
 	Filter *ResourceFilter `json:"filter,omitempty"`
-	// Version of the source
-	Version string `json:"version"`
```
dtaniwaki: This conflicts with `Version` of inline `GroupVersionKind`. I think we can remove it.

test/e2e/common/client.go
dominikh: Removing the default does change the behavior, but the existing behavior is arguably incorrect. The first case statement can never match, as `default` will always be ready before 10 seconds have passed. Also, since the presence of the default branch causes the loop to spin, you will keep creating new timers in `time.After` rapidly for as long as readyChan isn't ready, so this is arguably a memory leak, too.

Of course, once the `default` branch is removed, the loop will only ever execute exactly once, as both cases will break the loop.

dtaniwaki: I thnk we can just put the select without the loop.

VaibhavPage:
You may want to rebase for 
```
sensors/trigger.go:86:4: SA4004: the surrounding loop is unconditionally terminated (staticcheck)
                        return false
                        ^
sensors/trigger.go:94:4: SA4004: the surrounding loop is unconditionally terminated (staticcheck)
                        return true
```

VaibhavPage:
closing this PR, we will open up another PR to introduce proper linting.

PR #295, opened 2019-06-13
Ability to use JSON objects in the userPayload field for calendar events

elyzov-plesk:
At now, I am able to pass only string in userPayload field, but if I want to bind userPayload data to different places, sensor should receive it as object.

So, it would be great be able to use JSON objects in the userPayload field.

### DIFF
--- a/gateways/core/calendar/config.go
+++ b/gateways/core/calendar/config.go
@@ -17,6 +17,7 @@ limitations under the License.
 package calendar
 
 import (
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


jaygorrell:
Can you please give an example of how to send json data with this? I've tried everything but it's not getting picked up. For example, I expected to access `userPayload.command` as a `resourceParameters` `path` using this:
```
userPayload: '{"command":"e2e-tests"}'
```

Issue #315, opened 2019-07-26
Support append/prepend operations for trigger parameters


marxarelli:
**Is your feature request related to a problem? Please describe.**
I'm looking to use Argo Events to respond to [patchset-created events](https://gerrit-review.googlesource.com/Documentation/cmd-stream-events.html#_patchset_created) generated by Gerrit code review, creating a workflow that runs CI tasks based on a workflow file stored in the patchset's project repo. Gerrit does not include the full clone URL for the project in the event data and so it must be constructed from a prefix and the project name. However, this isn't possible given the fact that `templateParameters` can only overwrite values, not prepend/append existing ones.

**Describe the solution you'd like**
I would love a `TriggerParameter.Operation` field of some sort that supported keywords such as `overwrite` (default), `prepend`, and `append`. (Or something similar that achieves the desired functionality.) I'm down to author a pull request if that's helpful but I thought I'd ask first to judge how it might be received.

**Describe alternatives you've considered**
I've thought about writing a Gerrit plugin to include additional information in the event payload but then I'd have to write more Java and that's no fun.

**Additional context**
Full example of what I'm hoping for:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: gerrit-webhook-sensor
  labels:
    sensors.argoproj.io/sensor-controller-instanceid: argo-events
    argo-events-sensor-version: v0.10
spec:
  template:
    spec:
      containers:
        - name: "sensor"
          image: "argoproj/sensor"
          imagePullPolicy: Always
      serviceAccountName: argo-events-sa
  dependencies:
    - name: "gerrit-webhook-gateway:gerrit-events"
      filters:
        context:
          # source: { host: gerrit.git.wmflabs.org }
          contentType: application/json
        data:
          - path: type
            type: string
            value: [ patchset-created ]
  eventProtocol:
    type: "HTTP"
    http:
      port: "9300"
  triggers:
    - template:
        name: gerrit-webhook-workflow-trigger
        group: argoproj.io
        version: v1alpha1
        kind: Workflow
        source:
          git:
            url: https://gerrit.git.wmflabs.org/r/
            cloneDirectory: /src
            filePath: .pipeline/argo-workflow.yaml
            branch: REPLACED_WITH_PATCHSET_REF
      templateParameters:
        - src:
            event: "gerrit-webhook-gateway:gerrit-events"
            path: patchSet.ref
          dest: source.git.branch
        - src:
[...]

PR #319, opened 2019-07-26
Support different trigger parameter operations (#315)

marxarelli:
Implemented an additional `operation` field for trigger parameters that
dictates what is to be done with the current value at `dest`. Possible
keyword values are:

 - `overwrite` Overwrite the current value (default/prior behavior)
 - `prepend` Prepend the value at `dest` with the value at `source.path`
 - `append` Append the value at `dest` with the value at `source.path`


marxarelli:
Thanks for the merge!

VaibhavPage:
#315

PR #322, opened 2019-07-31
Support fetching/checkouts of non-branch/-tag git refs

marxarelli:
Provides a `source.git.ref` field for specifying a specific non-branch
non-tag git ref at which to perform a checkout. Providing one will
result in a shallow (depth 1) clone of the repo and shallow subsequent
fetches.

The motivation for this change is to support sourcing of trigger
templates from Gerrit patchsets. Gerrit uses non-branch/-tag refs
extensively and specifically in this case stores change patchsets in
`refs/changes/[slice]/[change]/[patchset]`. While arbitrary refs are
kind of an obscure git feature, there very well may be other git based
systems utilizing them.

### DIFF
--- a/pkg/apis/sensor/v1alpha1/types.go
+++ b/pkg/apis/sensor/v1alpha1/types.go
@@ -466,6 +466,11 @@ type GitArtifact struct {
 	// +optional
 	Tag string `json:"tag,omitempty" protobuf:"bytes,8,opt,name=tag"`
 
+	// Ref to use to pull trigger resource. Will result in a shallow clone and
+	// fetch.
+	// +optional
+	Ref string `json:"ref,omitempty" protobuf:"bytes,8,opt,name=ref"`
+
 	// Remote to manage set of tracked repositories. Defaults to "origin".
 	// Refer https://git-scm.com/docs/git-remote
 	// +optional
--- a/pkg/apis/sensor/v1alpha1/types.go
+++ b/pkg/apis/sensor/v1alpha1/types.go
@@ -469,12 +469,12 @@ type GitArtifact struct {
 	// Ref to use to pull trigger resource. Will result in a shallow clone and
 	// fetch.
 	// +optional
-	Ref string `json:"ref,omitempty" protobuf:"bytes,8,opt,name=ref"`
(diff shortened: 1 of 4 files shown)


marxarelli:
While I was implementing this change, I was a little confused about these `fetchRefSpecs` in `store/git.go`:

```go
var (
	fetchRefSpec = []config.RefSpec{
		"refs/*:refs/*",
		"HEAD:refs/heads/HEAD",
	}
)
```

Why such an aggressive fetch of `refs/*`? Doesn't seem necessary if one is wanting a branch/tag (the default fetch git sets up for a remote should cover those) and could potentially pull down lots of extra data. (We have a repo where +100M unnecessary data would be pulled down from non-default refs.)

The second one seems to confuse git by creating an actual branch named `HEAD` which conflicts with the symbolic ref of the same name. If I connect to the sensor pod and do a `git reflog` on the cloned repo I get something like "ambiguous reference HEAD." I can create an issue with more info if that helps.

Anyway, since I don't know the history behind those refspecs, I didn't want to touch it, and simply included a conditional exemption for when `source.git.ref` is supplied. In that case, it really shouldn't be necessary to have anything but the specific ref.

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -466,6 +466,11 @@ type GitArtifact struct {
 	// +optional
 	Tag string `json:"tag,omitempty" protobuf:"bytes,8,opt,name=tag"`
 
+	// Ref to use to pull trigger resource. Will result in a shallow clone and
+	// fetch.
+	// +optional
+	Ref string `json:"ref,omitempty" protobuf:"bytes,8,opt,name=ref"`
```
VaibhavPage: the index for the field in protobuf should be 9

marxarelli:
Thanks for the merge!

mgoodness:
How can I use this? Built from `master`, pushed & deployed new images, but `source.git.ref` isn't recognized as a valid field.

Diff when applying my Sensor via Argo CD:

Screen Shot 2019-08-29 at 1 53 47 PM

PR #324, opened 2019-08-02
Feature/support slack interaction actions

### LINKED ISSUE #308: Support Slack Interactive messages
Slack's new Block messaging composition allows adding interactive elements to messages. A user interacting with these elements causes Slack to trigger [a POST request](https://api.slack.com/messaging/interactivity/enabling) to a configured endpoint. This request can be the basis of triggering a workflow.

When a user interacts with an interactive message in Slack (say, clicks a button), a workflow should be triggered.

This can be achieved with a http webhook gateway, but Slack requires that interactive message actions be acknowledged within 3 seconds. A new gateway or retrofitting the existing Slack gateway would quite easily address this usecase.

logicfox:
- Resolves #308 
- Adds signature verification if Signing Secret is provided as a Kubernetes secret reference in the Gateway configmap.

Adds support for Slack Interactions - action events generated when a user interacts with a [block elements and interactive components](https://api.slack.com/messaging/interactivity#components). These events are sent as Http POST requests of MIME type `application/x-www-form-urlencoded`. The payload of the request is extracted and send to the active webhook endpoint, similar to how Events API inner event payloads are handled.


VaibhavPage:
Can you take a look at the failed pipeline?

logicfox:
I'm not sure how I can fix this data race in tests. I can't reproduce it on my laptop. https://travis-ci.org/argoproj/argo-events/builds/567166693#L923

akemner:
this seems to be exactly what I'd want to do but is there any docs or examples for getting this working?

spkane:
@logicfox @VaibhavPage Are there any docs or a single usage example for this, to help people try it out? Versus just relying on trying to parse the code and hoping for the best?

logicfox:
@spkane There's unfortunately no docs other than the examples. Have you taken a peek into that? The gateway receives the request from Slack and forwards it to the sensors as a payload.

spkane:
@logicfox We found a bug that we reported here: https://github.com/argoproj/argo-events/issues/726, it took us awhile to get around that. I haven't seen examples of using slack interactively with argo-events yet, but will continue to dig into it.

spkane:
@logicfox Is the Slack trigger capable of sending interactive messages? From what I can tell this code makes it possible to respond to an interaction from a payload that is received from an interactive slack message, but I see no way to get argo-events to send a message that utilizes the Block Kit. The Slack trigger appears to only expect a channel and a text message. Am I missing something here?

kavishdahekar:
> @logicfox Is the Slack trigger capable of sending interactive messages? From what I can tell this code makes it possible to respond to an interaction from a payload that is received from an interactive slack message, but I see no way to get argo-events to send a message that utilizes the Block Kit. The Slack trigger appears to only expect a channel and a text message. Am I missing something here?

Was anybody able to figure this out? Trying to send a json encoded block message in the "message" results in the json being sent in the slack message as is, without being parsed.

PR #366, opened 2019-09-28
getting_started.md: Change/fix manual port-forward cmd

StianOvrevage:
Changes `kubectl port-forward` to point to service instead of pod.

Reason:
I was not able to make this work using port-forwarding to the pod as per the docs. I also don't understand enough of the architecture to troubleshoot effectively. However, just port-forwarding to the service worked flawlessly.


whynowy:
This is not needed, `port-forward` to the POD should be good.

Issue #363, opened 2019-09-26
JSON parsing for SQS messages


catalinvr:
**Describe the bug**
Hello team,

I started with SQS event source for Argo events. Everything that is on example page works as expected.
Now I want to send a JSON message to SQS from my app and on Workflow side to access a value from that message and send as args to my task.

Thank you

**To Reproduce**

1. Define the SQS message like this:
```
{"foo":"bar"}
```

2. Set the container args like is below: 
```
  triggers:
    - template:
        name: sqs-workflow
        group: argoproj.io
        version: v1alpha1
        kind: Workflow
        source:
          inline: |
            apiVersion: argoproj.io/v1alpha1
            kind: Workflow
            namespace: argo
            metadata:
              generateName: rp-sqs-workflow-
            spec:
              entrypoint: whalesay
              arguments:
                parameters:
                - name: message
                  # this is the value that should be overridden
                  value: hello world
              templates:
              - name: whalesay
                inputs:
                  parameters:
                  - name: message
                container:
                  image: docker/whalesay:latest
                  command: [cowsay]
                  args: ["{{inputs.parameters.message.foo}}"]
      resourceParameters:
        - src:
            event: "rp-sqs-gateway:notification-1"
          dest: spec.arguments.parameters.0.value
``` 
3. deploy the sensor
4. send the SQS message
5. check that "bar" was displayed on workflow logs

**Expected behavior**
"bar" was displayed on workflow logs.

VaibhavPage:
You can define 

```
resourceParameters:
  - src:
       event: "rp-sqs-gateway:notification-1"
       path: "foo"
   dest: spec.arguments.parameters.0.value
       
```

And define `args` for `container` as args: ["{{inputs.parameters.message}}"]

catalinvr:
Hi @VaibhavPage 
Thanks for helping me with this. I'll test it this week.
Also, I'll check that I can add a PR on documentation.

Thank you

PR #371, opened 2019-10-07
#363 update the SQS gateway documentation

catalinvr:
Hi team, 

I would like to update the SQS gateway documentation based on #363
Please can you check that everything is fine?

Thank you.


catalinvr:
Hi @VaibhavPage please can you review the previous build. I fails with the following error:
`Error parsing coverage: open profile.cov: no such file or directory`

Issue #353, opened 2019-09-14
AMQP Gateway rate limit


fayak:
**Is your feature request related to a problem? Please describe.**
In the project I'm using argo in, I often get a batch of ~800events sent by an application I can't easily modify. The problem is that our Kubernetes cluster isn't able to handle all of these events simultaneously, but the Argo AMQP Gateway doesn't have to logic to determine the maximum number of events it may grab. Therefore, it fetches all messages available on rabbitmq, and indirectly kills our cluster with heavy load.

I thought of limiting the number of pod in the cluster with Kubernetes, but all exceeding events leads to a failure.

**Describe the solution you'd like**
To this problem, I reckon 2 available solutions :
- Have an option in Argo to limit the number of simultaneous running workflow, and therefore have the gateway to not pull events from rabbitmq if it overflows this limit
- Have an option in the AMQP gateway to define a rate limit and enforce it. It may look like something as "max number of event pulled by x unit of time"

**Describe alternatives you've considered**
The options I've thought of were;
- max number of pods -> it make all exceeding workflows fail
- parallelism option -> It doesn't exist at the "namespace" level
- rate limit through rabbit -> Rabbit doesn't seems to have rate limits on exchanges
- rate limit in the sending application -> In our deployment it is really hard to add such feature
- have a third service pulling from an exchange used by the sending app, and pushing them to a second exchange where argo pulls events -> lol

**Additional context**

/

VaibhavPage:
I may be wrong, but for option 1, you can take a look at `parallelism` level in Argo workflow. But I think it makes sense to have a rate limiter put in the gateway.

fayak:
@vaibhavPage yes I've looked at parallelism, but, correct me if I'm wrong, it doesn't prevent the maximum simultaneous execution of workflows, but only steps (pods?) within a workflow

VaibhavPage:
My bad, you are right.

fayak:
Closing it as #378 got merged

PR #378, opened 2019-10-15
Add rateLimit option to amqp gateway (#353)

Makunouchi:


gateways/core/stream/amqp/start.go
```diff
@@ -44,6 +45,31 @@ func (ese *AMQPEventSourceExecutor) StartEventSource(eventSource *gateways.Event
 	return gateways.HandleEventsFromEventSource(eventSource.Name, eventStream, dataCh, errorCh, doneCh, ese.Log)
 }
 
+func getLimitedDelivery(ch *amqplib.Channel, a *amqp, delivery chan amqplib.Delivery, queue string) {
+	for {
+		var elapsedTime int64
+		startTime := time.Now().Unix()
+
+		for i := uint32(0); i < a.RateLimit; i++ {
```
fayak: So If I get it right, if I send 100 messages, with a rate limit of 10msg,
I'll get 10msg at `00:00:00`, nothing for 59s, then another 10msg at `00:01:00`, etc, rather than a message at `00:00:00`, another one at `00:00:06`, etc

Am I correct ?

fayak: I feel like this is not wrong, but with such scenario, I'll be better to have an extra configuration option to set the time window instead of the default 60s

gateways/core/stream/amqp/config_test.go
```diff
@@ -26,6 +26,7 @@ url: amqp://amqp.argo-events:5672
 exchangeName: fooExchangeName
 exchangeType: fanout
 routingKey: fooRoutingKey
+rateLimit: 50
```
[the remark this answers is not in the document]
Makunouchi: That was a mistake, I removed it

gateways/core/stream/amqp/start.go
Makunouchi: To avoid this, I introduced a new option `ratePeriod` which allow the user to set its own time window.

gateways/core/stream/amqp/config.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package amqp
 
 import (
+	"errors"
```
VaibhavPage: Would you mind changing it to `github.com/pkg/errors` to keep the codebase consistent

PR #390, opened 2019-10-24
fix uuid with correct return value

XiangChen666:
current code will have error when doing go build without module


XiangChen666:
using google uuid

PR #410, opened 2019-11-26
Fixing whitespace in github events source example

sudermanjr:
Yaml was just a little off


sudermanjr:
Looks like this go fixed separately.

PR #415, opened 2019-11-28
feat: AWS SNS gateway support https endpoint

clementlecorre:

I would like support for the https protocol for the aws sns gateway.

For the moment it is possible to configure only in http
```
error="rpc error: code = Unknown desc = failed to send subscribe request. err: InvalidParameter: Invalid parameter: Endpoint must match the specified protocol\n\tstatus code: 400"
```
AWS Support :
The protocol you want to use. Supported protocols include:
- http – delivery of JSON-encoded message via HTTP POST
- https – delivery of JSON-encoded message via HTTPS POST

Add http or https in event source config  : https://github.com/argoproj/argo-events/blob/master/examples/event-sources/aws-sns.yaml

I have build docker img for testing: `habx/aws-sns-gateway:v0.11`


retnuh:
It would be nice to detect this based on the endpoint rather than making the user configure it every time...

delwaterman:
+1 on this. Any updates on when this will get merged?

nfickas:
Is this feature currently available?

PR #421, opened 2019-12-10
Update README.md

saradhis:
Add Intuit to the list of users.


sarabala1979:
@saradhis can you add it in alphabet order?

saradhis:
> @saradhis can you add it in alphabet order?

Updated.

PR #439, opened 2019-12-16
feat: Labels and annotations in service exposing sensor

### LINKED ISSUE #438: Enable passing labels and annotations to service exposing sensor
Our k8s setup requires us to add some labels to the service which the sensor exposes to communicate with the gateway for HTTP protocol. Otherwise it fails in our admission controller. It'll be nice to allow users to pass labels and annotations as part of the sensor spec

Modify the `sensor` spec, specifically the [`eventProtocol (http)`](https://github.com/argoproj/argo-events/blob/master/examples/sensors/calendar.yaml#L18) part to allow labels and annotations. Along these lines maybe - 
```
eventProtocol:
    type: "HTTP"
    http:
      labels:
        label-key: "label-value"
      annotations:
        annotation-key: "annotation-value"
      port: "9300"
```

Alternatively, we can add a `metadata` field to the `eventProtocol` key. Or we can duplicate the `sensor` level labels and annotations to the service

Shreyam:
fixes #438


Shreyam:
Closing as committed with wrong user name/email. Will recreate

PR #444, opened 2019-12-19
fix: Adding sensor name to NATS queue group name.

jshields-te:
The current queue group name approach only uses the event source name to
create the group name. This approach will cause the NATS events coming
from a single event source to get load balanced across *all* sensors.

This commit updates the queue group name to be composed of the sensor
name, the event source name, then `-queue`. This should allow the NATS
events to be properly load balanced within a specific sensor and allows
a fan-out of events to multiple sensors.


VaibhavPage:
The NATS receiver logic has been changed, closing this PR.

PR #489, opened 2020-02-12
Match example event source

zachaller:


zachaller:
This has been fixed by changing the dep instead.

PR #490, opened 2020-02-12
Fix kafka client nil reference

### LINKED ISSUE #485: Kafka Gateway is not working
**Describe the bug**
A simple Kafka gateway I setup using this definition is not working at all;

```
kind: Gateway
metadata:
  name: kafka-gateway
  labels:
    gateways.argoproj.io/gateway-controller-instanceid: argo-events
spec:
  replica: 1
  type: kafka
  eventSourceRef:
    name: kafka-event-source
  template:
    metadata:
      name: kafka-gateway
      labels:
        gateway-name: kafka-gateway
    spec:
      containers:
        - name: gateway-client
          image: argoproj/gateway-client:v0.12
          imagePullPolicy: Always
          command: ["/bin/gateway-client"]
        - name: kafka-events
          image: argoproj/kafka-gateway:v0.12
          imagePullPolicy: Never
          command: ["/bin/kafka-gateway"]
  subscribers:
    http:
      - "http://kafka-sensor.argo-events.svc:9300/"
```

The kafka-gateway logs is showing the following;

```
INFO[2020-02-08 00:01:15] started processing the event source...        event-source=kafka-event-source
INFO[2020-02-08 00:01:15] parsing the event source...                   event-source=kafka-event-source
INFO[2020-02-08 00:01:15] connecting to Kafka cluster...                event-source=kafka-event-source
```

That means that the process is getting blocked somewhere when trying to allocate the sarama consumer, without showing any error.

**Additional context**
The issue happens because kafkaEventSource is not properly allocated. Is created as a (nil) pointer, and yaml.Unmarshal cannot allocate it either. This can be fixed doing the following change on gateways/server/kafka/start.go;

From;

```
	var kafkaEventSource *v1alpha1.KafkaEventSource
	if err := yaml.Unmarshal(eventSource.Value, kafkaEventSource); err != nil {
		errorCh <- err
		return
	}
```

To;

```
	var kafkaEventSource v1alpha1.KafkaEventSource
	if err := yaml.Unmarshal(eventSource.Value, &kafkaEventSource); err != nil {
		errorCh <- err
		return
	}
```

If I have some time I will just work on the PR, otherwise you have already the solution here.

zachaller:
Fixes #485 ran into issues with kafka tested the changes suggested in 485 thought I would do a pr even though you mentioned add it in 486

### DIFF
--- a/gateways/server/kafka/start.go
+++ b/gateways/server/kafka/start.go
@@ -63,8 +63,8 @@ func (listener *EventListener) listenEvents(eventSource *gateways.EventSource, d
 	logger := listener.Logger.WithField(common.LabelEventSource, eventSource.Name)
 
 	logger.Infoln("parsing the event source...")
-	var kafkaEventSource *v1alpha1.KafkaEventSource
-	if err := yaml.Unmarshal(eventSource.Value, kafkaEventSource); err != nil {
+	var kafkaEventSource v1alpha1.KafkaEventSource
+	if err := yaml.Unmarshal(eventSource.Value, &kafkaEventSource); err != nil {
 		errorCh <- err
 		return
 	}


zachaller:
Closed I actually see its already in 486

PR #524, opened 2020-03-07
feat: update biobox presentation link to slideshare

bx-julian:


bx-julian:
closing because I used wrong account and force pushing just makes the commit have multiple authors from github's perspective

Issue #520, opened 2020-03-03
data.body from redis and aws-sqs are different: sqs data.body base64 encoded


miyanaga:
**Describe the bug**
A sensor receives raw payload from redis gateway but base64 encoded from aws-sqs gateway.

**To Reproduce**
1. Create a sensor like this:
```
apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: the-sensor
  labels:
    sensors.argoproj.io/sensor-controller-instanceid: argo-events
spec:
  template:
    spec:
      containers:
        - name: sensor
          image: argoproj/sensor:v0.13.0-rc
          imagePullPolicy: Always
      serviceAccountName: argo-events-sa
  subscription:
    http:
      port: 9300
  dependencies:
    - name: test-dep
      gatewayName: the-gateway
      eventName: example
  triggers:
    - template:
        name: workflow-trigger
        k8s:
          group: argoproj.io
          version: v1alpha1
          resource: workflows
          operation: create
          source:
            resource:
              apiVersion: argoproj.io/v1alpha1
              kind: Workflow
              metadata:
                generateName: event-job-
                namespace: default
              spec:
                entrypoint: whalesay
                arguments:
                  parameters:
                    - name: message
                      value: hello world
                templates:
                  - name: whalesay
                    inputs:
                      parameters:
                        - name: message
                    container:
                      image: docker/whalesay:latest
                      command: [cowsay]
                      args: ["{{inputs.parameters.message}}"]
          parameters:
            - src:
                dependencyName: test-dep
                dataKey: body
              dest: spec.arguments.parameters.0.value
```

If the-gateway is redis and publish like this:
```
PUBLISH the_channel "hello from redis"
```
cowsay: "hello from redis"

But the-gateway is aws-sqs and send message like this:
```
aws sqs send-message --queue-url="https://queue-url" --message-body="hello from sqs"
```
cawsay: "aGVsbG8gZnJvbSBzcXM="

"aGVsbG8gZnJvbSBzcXM=" is base64 encoded "hello from sqs"

**Expected behavior**
I think the body should not to be encoded.

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04
 - Version: v0.13.0-rc

antoniomo:
As far as I have tested, this also severely hampers the usability of the SQS events, as you can't really use triggers on the body payload.

miyanaga:
I think to decode base64 "Body" at here.

[...]

PR #542, opened 2020-03-15
feat: support json body as an option [POC]

antoniomo:
This is based on the discussion here: https://github.com/argoproj/argo-events/issues/520

This PR is to discuss how to proceed, and focuses on SQS, but ideally the same could be applied to other event sources and I would be happy to patch those as well, but I don't have an easy setup to test them.

In the specific case of SQS, the event body could be either JSON, XML, or a raw string. As such, blindly converting https://github.com/argoproj/argo-events/blob/master/pkg/apis/events/event-data.go#L77 to a `*json.RawMessage` would make `argo-events` unusable for the other possible payloads.

In this POC, I make it an `interface{}` and then add the https://github.com/argoproj/argo-events/commit/72bb77b31b40a077f16e09110ffe8f7cd7bcc163#diff-ddd95d9ab8f8f0146aac7d54b792d791R254-R257 field to the event-source. The idea is that if we know this SQS source will return JSON, we can treat it accordingly and allow for triggers and so on. If this field is not specified or false, we can't make any assumptions and the `Body` is treated as raw `[]byte`.

This way of handling "is the body JSON or not?" would work for other non-SQS event-sources where JSON is common, but not the only possibility.

All feedback welcome. Some questions I have myself:

- Is this how we want to handle this? Alternatives are welcome!
- Should the field `JSONBody bool` be instead a BodyType string or similar, to allow future expansion? <- I think this is advisable, I didn't add it to the POC to make it as simple as possible, for discussion


VaibhavPage:
I am fine with JSONBody bool as this should only apply to gateways for messaging queues where event structure is unknown. @antoniomo can you also apply the same change for AMQP, MQTT and NATS gateway?

antoniomo:
Added those. Can you verify also the `protobuf:"bytes,X,..."` tags on the structs here https://github.com/argoproj/argo-events/blob/master/pkg/apis/eventsources/v1alpha1/types.go ? Making sure those are right.

Issue #541, opened 2020-03-14
Support go templating for trigger parameters


tinyzimmer:
**Is your feature request related to a problem? Please describe.**

I'm toying with the idea of using `argo-events` and `argo-workflows` for CI. However, with many developers utilizing the console, I would need a clear and concise way for them to see which workflows are theirs. When using the github event hook, some key payload parameters contain special characters (/) that cannot be used inside names or labels of workflow objects. This sucks because it'd be awesome if the developer could view jobs by a label containing the ref of the branch that's building.

**Describe the solution you'd like**

I'd like to be able to optionally pass the event/context payloads through a go template (with sprig functions) before inserting them into the workflow yaml definition. In what I've tried locally the trigger definition becomes like this (`contextTemplate` is also implemented):

```yaml
  dependencies:
    - name: event-payload
      gatewayName: github-gateway
      eventName: my-repo
  triggers:
    - template:
        name: github-workflow-trigger
        k8s:
          group: argoproj.io
          version: v1alpha1
          resource: workflows
          operation: create
          source:
            resource:
              apiVersion: argoproj.io/v1alpha1
              kind: Workflow
              metadata:
                generateName: my-repo-
                labels:
                  ref:
                  commitSHA:
              spec:
                entrypoint: cicd
                arguments:
                  parameters:
                  - name: payload
                templates:
                - name: cicd
                  steps:
                  - - name: run-cicd
                      templateRef:
                        name: cicd-workflow-templates
                        template: cicd-template
                      arguments:
                        parameters:
                          # ...
          parameters:
            - src:
                dependencyName: event-payload
                dataKey: head_commit.id
              dest: metadata.labels.commitSHA
            - src:
                dependencyName: event-payload
                dataTemplate: '{{ .Event.ref | replace "/" "-" }}'
              dest: metadata.labels.ref
            - src:
                dependencyName: event-payload
                dataTemplate: '{{ .Event.ref | replace "refs/heads/" "" | replace "/" "-" }}-{{ trunc 7 .Event.head_commit.id }}-'
              dest: metadata.generateName
              operation: append
```

**Describe alternatives you've considered**

Not a whole lot at the moment. Other than perhaps building a custom container that does formatting on the incoming event before passing it to the gateways/sensors. 

**Additional context**

I've implemented this on a local copy of the code, and can open a PR for further discussion if we want.

VaibhavPage:
a PR would be awesome.

PR #543, opened 2020-03-15
feat: Support dataTemplate and contextTemplate for Trigger Parameters

tinyzimmer:
This PR relates to the issue I opened:  #541 

One thing I am definitely open to maybe doing is maybe a number on `ResolveParamValue`. With the two additions it feels like it's getting a bit unruly. 

I can add some coverage as well. So consider this just a WIP to get feedback on the approach.

As a side note, I am having trouble running `make codegen`

```
./hack/update-api-docs.sh
I0315 18:03:19.811707    6828 main.go:127] parsing go packages in directory github.com/argoproj/argo-events/pkg/apis/eventsources/v1alpha1
F0315 18:03:21.224675    6828 main.go:133] no API packages found in github.com/argoproj/argo-events/pkg/apis/eventsources/v1alpha1
```

But that doesn't seem right...


tinyzimmer:
> 1. You may need to run `make codegen` and commit updated files for deep copy and api docs.

I feel like an idiot, but I'm still struggling with this. On my mac it gives the error I showed above once it's done *allegedly* running deep copy funcs (that dont produce any changes).

Tried on Linux, and it ran all the way through, but again didn't actually generate any deep-copy code, and for the html/md files it just converted every kind to it's full path...e.g.

```diff
-<code>context</code></br> <em> Argo Events common.EventContext </em>
+<code>event</code></br> <em>
+github.com/argoproj/argo-events/pkg/apis/common.Event </em>
```

This is beginning to make me doubt my sanity. 

> 2. Can you also add an example sensor YAML file under `examples/sensors` folder that demonstrates the templating functionality?

I will add an example if not tonight, then sometime in the morning.

tinyzimmer:
> Changes look good to me. I'll fix the `make codegen` part.

I'd be very curious what the issue was in case I want to contribute more in the future 😄

VaibhavPage:
As you have introduced string fields within an existing struct, the deep-copy script won't update anything. But, `make codegen` should at least update API docs under `api` folder and openapi_generated.go file.  `make codegen` really just runs 1) https://github.com/argoproj/argo-events/blob/master/hack/update-codegen.sh 2) https://github.com/argoproj/argo-events/blob/master/hack/update-openapigen.sh and 3) https://github.com/argoproj/argo-events/blob/master/hack/update-api-docs.sh under the hood. So, you can run these files individually as well.

VaibhavPage:
Also, can you resolve the conflicts? Thanks.

Issue #487, opened 2020-02-10
Allow for comparators in data filters


nfickas:
**Is your feature request related to a problem? Please describe.**
From the documentation here: https://argoproj.github.io/argo-events/tutorials/07-filters/. We have to use exact values when filtering on a bool or float data type. I would like the ability to use comparators to filter on my data values being passed in.

**Describe the solution you'd like**

```
dependencies:
    - name: test-dep
      gatewayName: webhook-gateway
      eventName: example
      filters:
        name: data-filter
        data:
          - path: body.value
            type: float
            comparator: '>'
            value: 
             - 50.0 
```

**Describe alternatives you've considered**

**Additional context**

VaibhavPage:
Makes sense. I'll make it a part of v0.13

PR #544, opened 2020-03-17
feat: Add Data Filter Comparator

chaseterry:
Added a comparator option to the sensor data filter for issue #487. Options include omited, "", ">=", ">", "=", "<", and "<=". If omited or empty-string, is treated the same as equality.
ex: 
```
filters:
        name: data-filter
        data:
          - path: body.value
            type: number
            comparator: ">"
            value:
              - "50.0"
```


pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -242,8 +242,23 @@ type DataFilter struct {
 	// Strings are taken as is
 	// Nils this value is ignored
 	Value []string `json:"value" protobuf:"bytes,3,rep,name=value"`
+	// Comparator compares the event data with a user given value.
+	// Can be ">=", ">", "=", "<", or "<=".
+	// Is optional, and if left blank treated as equality "=".
+	Comparator Comparator `json:"comparator,omitempty" protobuf:"bytes,4,opt,name=comparator"`
 }
 
+type Comparator string
```
VaibhavPage: Can you move the type declaration to the top with other types? just to keep it consistent. Also add a comment on the `type Comparator string`.

chaseterry: updated

PR #566, opened 2020-03-29
chore: upgrade to 1.13 and switch to go modules

k:
This PR was built with go 1.13 and uses go modules instead of dep.

`azure-events-hub` is currently the only target that does not build properly


VaibhavPage:
Thanks for PR!! I had this on back-burner for a while. Let me test and make sure everything works.

whynowy:
@k - Would you like to fix the conflict and get this moving on?

whynowy:
See https://github.com/argoproj/argo-events/pull/636.

PR #589, opened 2020-04-12
fix: update sprig template example

VaibhavPage:


biplabMazumdar:
how do i fetch values from arrays elements using dataTemplate . ?
This does not work 
dataTemplate: "{{ .Input.body.commits[0].id  }}"
image

Issue #503, opened 2020-02-24
Support GKE Workload Identity for GCP PubSub


giorgioazzinnaro:
**Is your feature request related to a problem? Please describe.**
When using GCP PubSub as an EventSource, a file is required with a service account credentials.
However, the recommended method to authenticate Google Service Accounts in a GKE cluster is to use [Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity).

**Describe the solution you'd like**
Make the `credentialsFile` parameter optional in the `spec.pubsub` block, similar to #300 and use 

**Describe alternatives you've considered**
I'm not sure if one could simply set up Workload Identity first and then use the path to the default application credentials file in the configuration. Wouldn't be proper though, as the Go client provides ways to autodiscover this.

**Additional context**
I still need to understand whether the `argo-events-sa` service account has anything special to it?
For Workload Identity to work properly, the service account being used in Kubernetes must have a specific annotation set. This is the only requirement from the Kubernetes side of things.

VaibhavPage:
I am not familiar with the workload identity, I need to investigate a bit.

tmshn:
#593 seems to fix this, but the name is wrong. "WorkflowIdentity" should be "WorkloadIdentity".
(but I don't know the policy how field deprecation proceeds on argo-events)

PR #593, opened 2020-04-15
feat: enabling workflow identity for GKE

chaseterry:
Added enableWorkflowIdentity flag to PubSubEventSource.
Added Workflow Identity toggle in gcp-pubsub gateway based on enableWorkflowIdentity flag.

Closing #503


examples/event-sources/gcp-pubsub-workflow-identity.yaml
```diff
@@ -0,0 +1,21 @@
...
+      # id of your project
+      projectID: argo-events-XXXXX
+      # (optional) id of project for topic, same as projectID by default
+      # topicProjectID: "project-id"
+      # topic name
+      topic: test
+      # Empty credentials file when using Workflow Identity
+      credentialsFile: ""
+      # If enableWorkflowIdentity is true, the projects uses Workflow Identity for authentication
+      enableWorkflowIdentity: true
```
whynowy, line 21: Can we do something like:

if len(credentialsFile) == 0 {
  // consider it is using workload identity
} else {
  //
}

chaseterry: That would make it simpler, but I think it's helpful to include the enableWorkflowIdentity flag to inform users that Workflow Identity is an option

whynowy: I think informing the users that Workflow Identity is supported in the doc might be better than in the code.

gateways/server/gcp-pubsub/start.go
```diff
@@ -85,21 +85,32 @@ func (listener *EventListener) listenEvents(eventSource *gateways.EventSource, c
...
+	} else {
+		client, clientErr = pubsub.NewClient(ctx, pubsubEventSource.ProjectID, option.WithCredentialsFile(pubsubEventSource.CredentialsFile))
+	}
+	if clientErr != nil {
+		return errors.Wrapf(clientErr, "failed to set up client for %s", eventSource.Name)
 	}
 
 	// use same client for topic and subscription by default
 	topicClient := client
 	if pubsubEventSource.TopicProjectID != "" && pubsubEventSource.TopicProjectID != pubsubEventSource.ProjectID {
```
VaibhavPage, line 104: This needs to change, We have defined logic to instantiate the client twice literally next to each other :man_facepalming: :man_facepalming: . I don't know how it got passed me in the first place, but it needs to change.

Can you do something like,
```

	var opt []option.ClientOption
	projectId := pubsubEventSource.ProjectID

	if pubsubEventSource.EnableWorkflowIdentity {
		opt = append(opt, option.WithCredentialsFile(pubsubEventSource.CredentialsFile))
	}
	if pubsubEventSource.TopicProjectID != "" && pubsubEventSource.TopicProjectID != pubsubEventSource.ProjectID {
		projectId = pubsubEventSource.TopicProjectID
	}

	client, err := pubsub.NewClient(ctx, projectId, opt...)
	if err != nil {
		return errors.Wrapf(err, "failed to set up client for %s", eventSource.Name)
	}


```

chaseterry: Updated

gateways/server/gcp-pubsub/validate.go
```diff
@@ -65,8 +66,5 @@ func validate(eventSource *v1alpha1.PubSubEventSource) error {
 	if eventSource.Topic == "" {
 		return fmt.Errorf("must specify topic")
 	}
```
VaibhavPage, line 68: You want to check if the EnableWorkloadIdentity is set to true, if so, then  its fine to not have credentialsFile else it is required to have the credentialsFile.

chaseterry: fixed

examples/event-sources/gcp-pubsub-workflow-identity.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: argoproj.io/v1alpha1
```
VaibhavPage, line 1: instead of creating a separate file for workflow identity example, can you put it under existing example https://github.com/argoproj/argo-events/blob/25db621a6c150465442fff27a7a418c4f4402d08/examples/event-sources/gcp-pubsub.yaml

Something like,
```

kind: EventSource
metadata:
  name: gcp-pubsub-event-source
spec:
  type: pubsub
  pubSub:
    example:
      # jsonBody specifies that all event body payload coming from this
      # source will be JSON
      jsonBody: true
      # id of your project
      projectID: argo-events-XXXXX
      # (optional) id of project for topic, same as projectID by default
      # topicProjectID: "project-id"
      # topic name
      topic: test
      # Refers to path of the credential file that is mounted in the gateway pod.
      credentialsFile: /creds/key.json

#    example-workload-identity:
#      # jsonBody specifies that all event body payload coming from this
#      # source will be JSON
#      jsonBody: true
#      # id of your project
#      projectID: argo-events-XXXXX
#      # (optional) id of project for topic, same as projectID by default
#      # topicProjectID: "project-id"
#      # topic name
#      topic: test
#      # Empty credentials file when using Workflow Identity
#      credentialsFile: ""
#      # If enableWorkflowIdentity is true, the projects uses Workflow Identity for authentication
#      enableWorkflowIdentity: true
#


```

chaseterry: done

PR #602, opened 2020-04-21
feat: make deployment template in gateway obj optional. Closes #598

### LINKED ISSUE #598: Make deployment/Service template optional in Gateway
It is not user-friendly that user need to define Deployment (and Service) every time they create a Gateway.

Make those default configurations, and allow the user to override.

**Additional context**

whynowy:
This PR is used to simplify the user experience of declaring `Gateway` spec.

With this change, a simplified Gateway spec looks like below:

```
kind: Gateway
metadata:
  name: webhook-gateway-1
spec:
  type: webhook
  eventSourceRef:
    name: webhook-event-source
  template:
  service:
    ports:
      - port: 13000
        targetPort: 12000
        name: abc
  subscribers:
    http:
      - "http://webhook-sensor.argo-events.svc:9300/"
```
If you want to customize gateway container, you can achieve that by adding `container` under `spec.template` to do it. For mounting volumes, add `volumes` under `spec.template`. 

There will be several new ENVs in gateway-controller deployment spec:
```
          env:
            - name: NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
            - name: CONTROLLER_CONFIG_MAP
              value: gateway-controller-configmap
            - name: GATEWAY_CLIENT_IMAGE
              value: gcr.io/whynowy/gateway-client:{version}
            - name: GATEWAY_IMAGE_REGISTRY
              value: docker.io
            - name: GATEWAY_IMAGE_VERSION
              value: {version}
```

Follow-up PR will make change to release manifests and update the examples.

Closes #598


controllers/gateway/resource.go
```diff
@@ -29,23 +31,49 @@ import (
...
+	if svc == nil {
 		return nil, nil
 	}
-	service := ctx.gateway.Spec.Service.DeepCopy()
-	if err := controllerscommon.SetObjectMeta(ctx.gateway, service, v1alpha1.SchemaGroupVersionKind); err != nil {
+	svc.Name = fmt.Sprintf("%s-gateway-svc", ctx.gateway.Name)
+	if svc.Spec.Selector == nil {
+		svc.Spec.Selector = make(map[string]string)
+	}
+	svc.Spec.Selector[common.LabelGatewayName] = ctx.gateway.Name
```
whynowy, line 50: Always use a fixed `selector`

controllers/gateway/resource.go
```diff
@@ -143,7 +172,7 @@ func (ctx *gatewayContext) createGatewayDeployment() (*appv1.Deployment, error)
 	if err != nil {
 		return nil, err
 	}
-	return ctx.controller.k8sClient.AppsV1().Deployments(deployment.Namespace).Create(deployment)
+	return ctx.controller.k8sClient.AppsV1().Deployments(ctx.gateway.Namespace).Create(deployment)
```
whynowy, line 175: Auto generated Deployment and Service should be in the same namespace as Gateway, it doesn't make sense to create objects in a namespace other than its own.

controllers/gateway/resource.go
```diff
@@ -29,23 +31,49 @@ import (
 
 // buildServiceResource builds a new service that exposes gateway.
 func (ctx *gatewayContext) buildServiceResource() (*corev1.Service, error) {
```
whynowy, line 33: Shall we build Service resource? Or shall we only build a ClusterIP service? Follow-up is needed.

controllers/gateway/controller_test.go
```diff
@@ -37,6 +40,38 @@ func newController() *Controller {
 			Namespace:  common.DefaultControllerNamespace,
 			InstanceID: "argo-events",
 		},
+		TemplatesConfig: map[apicommon.EventSourceType]Template{
+			apicommon.WebhookEvent: Template{
+				Deployment: &corev1.PodTemplateSpec{
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
```
VaibhavPage, line 47: Looking at the pod template of a gateway, I am realizing that we only need to configure the second container which is the actual events container. The first container which is `gateway-client` remains constant for all gateways and its behavior is fixed and there is no need to configure it separately. The only configuration for the `gateway-client` container can be to add `resources` to limit CPU and memory and we can make that default as part of the logic to construct the gateway pod template. 

So, the only thing we should let users configure is events container + pod related fields

If we store implements for different gateway types in controller configmap, then the user will lose the ability to configure different gateways of the same type in different ways.

Here is gateway CR spec I propose,
```

kind: Gateway
metadata:
  name: webhook-gateway
  labels:
    # the labels defined for the gateway CR will be applied on gateway deployment/pod and gateway service.
spec:
  type: webhook
  eventSourceRef:
    name: webhook-event-source
  # Optional
  template:
    # events container. we don't need configuration for gateway-client container.
    container:
      # container spec from core v1 K8s group.
      # user can specify container specific field as per their usecase.
      # we will have default values in controller logic.
    volumes:
      # list of volumes
    dnsPolicy:
      # Pod's DNS policy
    nodeSelector:
      # node selector to schedule the pod
    hostNetwork:
      # host network config
    securityContext:
      # security context
    affinity:
      # node affinity
    tolerations:
      # pod toleration
    priorityClassName:
    priority:
    dnsConfig:
  # Optional. should be used when user wants to expose the gateway pod.
  # the controller will create a ClusterIP service with `port` and `targetPort` as the values
  # in the list.
  # Whenever user adds or removes a port from this list, controller should update the ClusterIP service accordingly.
  service:
    ports:
      - 12000
      - 13000
      - 14000
  subscribers:
    http:
      - "http://webhook-sensor.argo-events.svc:9300/"


```

It is similar to how Argo Workflow spec is laid out.

common/util.go
```diff
@@ -130,3 +130,77 @@ func GetTLSConfig(caCertPath, clientCertPath, clientKeyPath string) (*tls.Config
...
+
+// CopyDeploymentSpecTemplate is used to copy exsired fields from a src podtemplatespc to a dest.
+func CopyDeploymentSpecTemplate(dest, src *v1.PodTemplateSpec) *v1.PodTemplateSpec {
+	if src == nil && dest == nil {
+		return nil
+	}
+	if dest == nil {
+		return src.DeepCopy()
+	}
+	result := dest.DeepCopy()
```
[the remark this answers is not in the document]
whynowy: 👍 I was looking for something like that, thanks for pointing out.

controllers/gateway/controller_test.go
whynowy: I agree not to let the user to configure `gateway-client` container, we can put its image in gateway-controller deployment spec as an ENV or argument. 

Where do you propose to put default event container config? Also, are you suggesting not to let the user to specify the image for event container? 

For service, I tend to not create it at all, because user always need to make change to it, or add new objects like `ingress` to it for public access. 

For gateway deployments like webhook, I suggest to let it listen to a fixed port (8080), I guess no user care about what port the container is listening to. For SSL support in gateway deployment, what we can do is put a self-signed CERT in the image on demand - this is what `argo-server` in workflow does right now, this is good enough to secure the communication from a load balancer to the POD. 

@VaibhavPage

whynowy: Also, do you see any requirements to expose properties like `dnsPolicy`, `nodeSelecor`... to the user at this time? If not, I would suggest not to expose them. I would rather not to expose them until we see lots of demands. Thoughts?

VaibhavPage: Agreed. We can add those fields later as needed.

controllers/gateway/resource.go
```diff
@@ -191,21 +245,20 @@ func (ctx *gatewayContext) updateGatewayDeployment() (*appv1.Deployment, error)
...
 
 	if currentDeployment.Annotations != nil && currentDeployment.Annotations[common.AnnotationResourceSpecHash] != newDeployment.Annotations[common.AnnotationResourceSpecHash] {
 		ctx.updated = true
-		if err := ctx.controller.k8sClient.AppsV1().Deployments(currentDeployment.Namespace).Delete(currentDeployment.Name, &metav1.DeleteOptions{}); err != nil {
-			return nil, err
-		}
-		return ctx.controller.k8sClient.AppsV1().Deployments(newDeployment.Namespace).Create(newDeployment)
+		currentDeployment.Spec = newDeployment.Spec
+		currentDeployment.Annotations[common.AnnotationResourceSpecHash] = newDeployment.Annotations[common.AnnotationResourceSpecHash]
+		return ctx.controller.k8sClient.AppsV1().Deployments(ctx.gateway.Namespace).Update(currentDeployment)
```
whynowy, line 261: Use `update` instead of `delete` and `create` to avoid down time.

VaibhavPage: 👍

controllers/gateway/resource.go
```diff
@@ -29,48 +33,94 @@ import (
...
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: labels,
+			},
+			Spec: corev1.PodSpec{
+				ServiceAccountName: ctx.gateway.Spec.Template.ServiceAccountName,
+				Containers: []corev1.Container{
+					{
+						Name:            "gateway-client",
+						Image:           ctx.controller.clientImage,
+						ImagePullPolicy: corev1.PullAlways,
```
VaibhavPage, line 99: Can you add `Resources` to the `gateway-client` container?

controllers/gateway/resource.go
VaibhavPage: 👍

whynowy:
> Added one comment for `Resources`(memory and cpu limit and requests) for `gateway-client` container. Lets add,
> 
> resources:
> requests:
> memory: "64Mi"
> cpu: "250m"
> limits:
> memory: "128Mi"
> cpu: "500m"

Done. 

Except setting `requests.cpu` = "50m".

PR #607, opened 2020-04-22
feat: Unify manifests structure with Argo

tumido:
I've been lacking some structure in Argo Events manifests, so I've inspired myself in the structure both [Argo](https://github.com/argoproj/argo/tree/master/manifests) and [ArgoCD](https://github.com/argoproj/argo-cd/tree/master/manifests) are using and made the kustomize resources structured the same.

Changes:
- `argo-events` namespace is not created by default, it is expected to exist (should be changed?)
- Separates the cluster wide installation from namespaced install
- `manifests/install.yaml` and `manifests/namespace-install.yaml` are generated as `kubectl kustomize manifests/cluster-install > manifests/install.yaml`
- Gateway and sensor controller configmaps in cluster wide installation are set to watch all namespaces (key `namespace` is unset), this is patched in the namespace scoped installation to limit the controllers to the `argo-events` namespace only.
- move manifests from `hack/k8s/manifests/` to `manifests` folder to respect the folder structure in Argo.
- Use the same naming convention as in Argo

Both installation methods were tested on my cluster, so it should behave :slightly_smiling_face:


manifests/README.md
```diff
@@ -0,0 +1,25 @@
+# Argo Events Install Manifests
```
VaibhavPage, line 1: Can we add this info under https://argoproj.github.io/argo-events/installation/ as well?

manifests/base/cdrs/event-source-crd.yaml
```diff
@@ -1,16 +1,15 @@
-# Define a "event source" custom resource definition
 apiVersion: apiextensions.k8s.io/v1beta1
```
VaibhavPage, line 1: while we are at it, lets update the `apiVersion` to `apiextensions.k8s.io/v1` for all CRDs. We can later add openAPI validation in a separate PR.

manifests/base/gateway-controller/gateway-controller-deployment.yaml
```diff
@@ -16,8 +15,7 @@ spec:
       serviceAccountName: argo-events-sa
       containers:
         - name: gateway-controller
-          image: argoproj/gateway-controller:v0.14.0
-          imagePullPolicy: Always
+          image: argoproj/gateway-controller:latest
```
VaibhavPage, line 18: I am not very comfortable with the latest tag images. Can we pin to a version?

manifests/install.yaml
```diff
@@ -0,0 +1,290 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
VaibhavPage, line 1: We need to update the link in the installation guide.

manifests/README.md
tumido: Added :+1: I've also reworded the setup for both cluster-wide and namespace install and added a section with detail Kustomize setup. Lmk if there's anything to be changed or reworded.

manifests/base/cdrs/event-source-crd.yaml
tumido: Done. FYI: I've had to bump the Kubernetes version requirement in `docs/installation` to v1.16, since that's when `v1` became available.

manifests/base/gateway-controller/gateway-controller-deployment.yaml
tumido: I've made the image tags [customizable](https://kubectl.docs.kubernetes.io/pages/app_customization/customizing_pod_templates.html#customizing-images), so it will be easier to maintain - both controller's image tags are enforced in `base/kustomization.yaml` - together, at the same place. Default is left to latest in the `base/sensor-controller/sensor-controller-deployment.yaml` and `base/gateway-controller/gateway-controller-deployment.yaml`. Would that be a suitable solution?

VaibhavPage: that works

PR #613, opened 2020-04-24
feat: make sensor deployment template spec optional. Closes #599

### LINKED ISSUE #599: Make deployment/Service template optional in Sensor
It is not user-friendly that user need to define Deployment every time they create a Sensor.

Make those default configurations, and allow the user to override.

**Additional context**

whynowy:
This PR is used to simplify the user experience of declaring Sensor spec.

With this change, a simplified Sensor spec looks like below:
```
kind: Sensor
metadata:
  name: webhook-sensor
  labels:
spec:
  template:
  dependencies:
    - name: test-dep
      gatewayName: webhook-gateway
      eventName: example
  subscription:
    http:
      port: 9300
  triggers:
    - template:
... ...
```
If you want to customize sensor container, you can achieve that by adding container under `spec.template` to do it. For mounting volumes, add `volumes` under spec.template.

```
... ...
spec:
  template:
    container:
      volumeMounts:
        - mountPath: /test-data
          name: test-data
    volumes:
      - name: test-data
        emptyDir: {}
... ...
```
There will be a new ENV in sensor-controller deployment spec:
```
            - name: SENSOR_IMAGE
              value: argoproj/sensor:v0.14.0
```

Follow-up PR will make change to release manifests and update the examples.

Closes #599


api/event-source.html
```diff
@@ -2649,5 +2649,5 @@ <h3 id="argoproj.io/v1alpha1.TLSConfig">TLSConfig
 <hr/>
 <p><em>
 Generated with <code>gen-crd-api-reference-docs</code>
-on git commit <code>9f73feb</code>.
+on git commit <code>a825153</code>.
```
whynowy, line 2652: `make codegen` should not update this. Need follow-up.

VaibhavPage:
Similar to gateway template configuration, lets change the sensor template as well.

Also, we need to provide env var to sensor controller for image.

controllers/sensor/resource.go
```diff
@@ -138,21 +170,20 @@ func (ctx *sensorContext) updateDeployment() (*appv1.Deployment, error) {
...
 	}
 
 	if currentDeployment.Annotations != nil && currentDeployment.Annotations[common.AnnotationResourceSpecHash] != newDeployment.Annotations[common.AnnotationResourceSpecHash] {
 		ctx.updated = true
-		if err := ctx.controller.k8sClient.AppsV1().Deployments(currentDeployment.Namespace).Delete(currentDeployment.Name, &metav1.DeleteOptions{}); err != nil {
-			return nil, err
-		}
-		return ctx.controller.k8sClient.AppsV1().Deployments(newDeployment.Namespace).Create(newDeployment)
+		currentDeployment.Spec = newDeployment.Spec
+		currentDeployment.Annotations[common.AnnotationResourceSpecHash] = newDeployment.Annotations[common.AnnotationResourceSpecHash]
```
whynowy, line 185: Use `update` instead of `delete` and `create` to avoid down time.

PR #632, opened 2020-04-30
chore: update release manifests and examples with new spec

whynowy:


manifests/cluster-install/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-bases:
+resources:
```
whynowy, line 4: `bases` is deprecated.

manifests/namespace-install/kustomization.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-bases:
+resources:
```
whynowy, line 4: `bases` is deprecated.

manifests/base/kustomization.yaml
```diff
@@ -1,16 +1,48 @@
...
+  - name: argoproj/sensor-controller
+    newTag: v0.14.0
+
+patchesStrategicMerge:
+  - |-
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      name: gateway-controller
+    spec:
```
whynowy, line 26: Put all the `version` patch in one file to make it easy to update.

manifests/namespace-install.yaml
```diff
@@ -171,6 +175,12 @@ spec:
               fieldPath: metadata.namespace
         - name: CONTROLLER_CONFIG_MAP
           value: gateway-controller-configmap
+        - name: GATEWAY_CLIENT_IMAGE
```
VaibhavPage, line 178: lets get rid of CLIENT_IMAGE env var.

whynowy: Can you give more detail? @VaibhavPage

PR #633, opened 2020-05-01
fix(controller): Make gateway and sensor controllers backward compatible

whynowy:
This PR makes both gateway and sensor controllers backward compatible with deprecated spec.

The old spec will be unsupported in few releases.

### DIFF
--- a/controllers/gateway/resource.go
+++ b/controllers/gateway/resource.go
@@ -34,7 +34,15 @@ import (
 
 // buildServiceResource builds a new service that exposes gateway.
 func (ctx *gatewayContext) buildServiceResource() (*corev1.Service, error) {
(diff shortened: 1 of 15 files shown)


controllers/gateway/resource.go
```diff
@@ -100,7 +133,7 @@ func (ctx *gatewayContext) makeDeploymentSpec() (*appv1.DeploymentSpec, error) {
 						ImagePullPolicy: corev1.PullAlways,
 						Resources: corev1.ResourceRequirements{
 							Requests: corev1.ResourceList{
-								corev1.ResourceCPU:    apiresource.MustParse("50m"),
+								corev1.ResourceCPU:    apiresource.MustParse("10m"),
```
whynowy, line 136: By checking `kubectl top pod`, usually when the gateway POD starts, it only takes `1m`. Lets set `request.cpu=10m` here.

PR #636, opened 2020-05-04
chore: change to use go mod

whynowy:
Switch to `go mod`.

k8s version: v1.17.5

Temporary solution for `codegen` output directory not able to point to codebase dir: create a temp dir under `/tmp/`, copy sub dirs to code base after codegen. See `./hack/update-openapigen.sh` and `./hack/update-codegen.sh`. Let me know if there's a better way.

Expected performance drop when running `make codegen`, see https://github.com/kubernetes/code-generator/issues/69. Need to follow up on this.


VaibhavPage:
@wktmeow fyi

PR #640, opened 2020-05-05
feat(trigger-patch-op): added patch operation

VaibhavPage:
Signed-off-by: VaibhavPage <vaibhav.page@gmail.com>

### DIFF
--- a/sensors/triggers/standard-k8s/starndard-k8s.go
+++ b/sensors/triggers/standard-k8s/starndard-k8s.go
@@ -132,7 +132,7 @@ func (k8sTrigger *StandardK8sTrigger) Execute(resource interface{}) (interface{}
 
 	switch op {
 	case v1alpha1.Update:
-		if err := mergo.Merge(oldObj, obj); err != nil {
+		if err := mergo.Merge(oldObj, obj, mergo.WithOverride); err != nil {
 			return nil, errors.Errorf("failed to update the object. err: %+v\n", err)
 		}
 		return k8sTrigger.namespableDynamicClient.Namespace(namespace).Update(oldObj, metav1.UpdateOptions{})
--- a/sensors/triggers/standard-k8s/starndard-k8s.go
+++ b/sensors/triggers/standard-k8s/starndard-k8s.go
@@ -116,11 +116,6 @@ func (k8sTrigger *StandardK8sTrigger) Execute(resource interface{}) (interface{}
 	}
 	obj.SetNamespace(namespace)
 
-	op := v1alpha1.Update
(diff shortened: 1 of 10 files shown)


sensors/triggers/standard-k8s/starndard-k8s.go
```diff
@@ -114,22 +116,36 @@ func (k8sTrigger *StandardK8sTrigger) Execute(resource interface{}) (interface{}
...
+	}
+
+	switch op {
 	case v1alpha1.Update:
-		if _, err := k8sTrigger.namespableDynamicClient.Namespace(namespace).Get(obj.GetName(), metav1.GetOptions{}); err != nil {
-			if !apierr.IsNotFound(err) {
-				return nil, err
-			}
-			return k8sTrigger.namespableDynamicClient.Namespace(namespace).Create(obj, metav1.CreateOptions{})
+		if err := mergo.Merge(oldObj, obj); err != nil {
```
whynowy, line 135: `mergo.Merge(oldObj, obj, mergo.WithOverride)`

Without `WithOverride`, it won't merge the non-empty fields in the destination.

sensors/triggers/standard-k8s/starndard-k8s.go
```diff
@@ -114,22 +116,36 @@ func (k8sTrigger *StandardK8sTrigger) Execute(resource interface{}) (interface{}
...
 
-	op := v1alpha1.Create
-	if trigger.Template.K8s.Operation != "" {
+	// If object not found, create the object. No need to have a separate `Create` switch case.
+	oldObj, err := k8sTrigger.namespableDynamicClient.Namespace(namespace).Get(obj.GetName(), metav1.GetOptions{})
+	if err != nil {
+		if !apierr.IsNotFound(err) {
+			return nil, err
+		}
+		return k8sTrigger.namespableDynamicClient.Namespace(namespace).Create(obj, metav1.CreateOptions{})
```
whynowy, line 125: When `op==Patch && oldObj not existing`, it also does a `create`, this looks weird.

VaibhavPage: Similar to `kubectl apply`. Most of the users expect the sensor to create the object if not available or update/patch the object otherwise. The user who raised the issue also have similar requirements.

whynowy: I agree with that. I thought the `patch` option had different `resource` than `update` and `create`, just like `kubectl patch deployment xxxxx --patch patch_file`, which `patch_file` is not a valid `deployment` spec.

PR #642, opened 2020-05-06
feat: add headers http trigger

VaibhavPage:


VaibhavPage:
this is branched off of the patch-trigger branch, so you may see some nonrelated changes.

vaporpin:
hi I tried the tutorial on https://argoproj.github.io/argo-events/triggers/http-trigger/ but cannot get it working. On "kubectl -n argo-events logs minio-gateway-pod main", I got the following error "error="A header you provided implies functionality that is not implemented" event-source=example". Is the tutorial still valid after this change? thanks.

PR #643, opened 2020-05-07
fix: K8s trigger patch operation

VaibhavPage:


whynowy:
What is the reason for this chase? @VaibhavPage

PR #644, opened 2020-05-07
fix: gateway and sensor resource names

VaibhavPage:
The latest changes to gateway and sensor controller construct the deployment name as `<gateway-name>-<gateway-type>-gateway-<random-char>`, e.b. webhook-gateway-webhook-gateway-12345 which is not correct naming. We should not add `<gateway-type>-gateway` to the deployment or service name. Let the user decide what they want to call their gateway and sensor.

### DIFF
--- a/controllers/gateway/resource.go
+++ b/controllers/gateway/resource.go
@@ -51,7 +51,7 @@ func (ctx *gatewayContext) buildServiceResource() (*corev1.Service, error) {
 	}
 	svc := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fmt.Sprintf("%s-gateway-svc", ctx.gateway.Name),
+			Name: fmt.Sprintf("%s-svc", ctx.gateway.Name),
 		},
 		Spec: corev1.ServiceSpec{
 			Ports:    ctx.gateway.Spec.Service.Ports,
@@ -73,7 +73,7 @@ func (ctx *gatewayContext) buildLegacyServiceResource() (*corev1.Service, error)
 	}
 	svc := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fmt.Sprintf("%s-gateway-svc", ctx.gateway.Name),
+			Name: fmt.Sprintf("%s-svc", ctx.gateway.Name),
 		},
 		Spec: *ctx.gateway.Spec.Service.Spec,
 	}
@@ -112,7 +112,9 @@ func (ctx *gatewayContext) makeDeploymentSpec() (*appv1.DeploymentSpec, error) {
 	}
 
 	if ctx.gateway.Spec.Template.Container != nil {
-		mergo.Merge(&eventContainer, ctx.gateway.Spec.Template.Container, mergo.WithOverride)
+		if err := mergo.Merge(&eventContainer, ctx.gateway.Spec.Template.Container, mergo.WithOverride); err != nil {
+			return nil, err
+		}
 	}
 
 	return &appv1.DeploymentSpec{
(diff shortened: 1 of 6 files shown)


controllers/gateway/resource.go
```diff
@@ -112,7 +112,9 @@ func (ctx *gatewayContext) makeDeploymentSpec() (*appv1.DeploymentSpec, error) {
 	}
 
 	if ctx.gateway.Spec.Template.Container != nil {
-		mergo.Merge(&eventContainer, ctx.gateway.Spec.Template.Container, mergo.WithOverride)
+		if err := mergo.Merge(&eventContainer, ctx.gateway.Spec.Template.Container, mergo.WithOverride); err != nil {
```
whynowy, line 115: Good catch!

controllers/sensor/resource.go
```diff
@@ -95,10 +95,12 @@ func (ctx *sensorContext) makeDeploymentSpec() appv1.DeploymentSpec {
 		ImagePullPolicy: corev1.PullAlways,
 	}
 	if ctx.sensor.Spec.Template.Container != nil {
-		mergo.Merge(&sensorContainer, ctx.sensor.Spec.Template.Container, mergo.WithOverride)
+		if err := mergo.Merge(&sensorContainer, ctx.sensor.Spec.Template.Container, mergo.WithOverride); err != nil {
```
whynowy, line 98: Good Catch.

controllers/sensor/resource.go
```diff
@@ -41,7 +41,7 @@ func (ctx *sensorContext) generateServiceSpec() *corev1.Service {
 
 	serviceSpec := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        fmt.Sprintf("%s-sensor-svc", ctx.sensor.Name),
+			Name:        fmt.Sprintf("%s-svc", ctx.sensor.Name),
```
whynowy, line 44: This is the same naming convention as Gateway Service, with this, you need to ask the user never give a same name to Gateway and Sensor, otherwise there is an issue.

controllers/sensor/resource.go
```diff
@@ -141,14 +143,17 @@ func (ctx *sensorContext) makeLegacyDeploymentSpec() appv1.DeploymentSpec {
...
-	deploymentSpec := ctx.makeDeploymentSpec()
+	deploymentSpec, err := ctx.makeDeploymentSpec()
+	if err != nil {
+		return nil, err
+	}
 	deployment := &appv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace:    ctx.sensor.Namespace,
-			GenerateName: fmt.Sprintf("%s-sensor-", ctx.sensor.Name),
+			GenerateName: fmt.Sprintf("%s-", ctx.sensor.Name),
```
whynowy, line 153: I don't have a strong opinion about either Gateway or Sensor deployments' `GenerateName`, and unlike Service, giving both Gateway and Sensor a same name won't cause any issue (some of the users might do that), but it will be harder for the users to identify which one is for Gateway, and which one is for Sensor.

controllers/sensor/resource.go
VaibhavPage: I was thinking along the lines that the user will name the gateway CR as `webhook-gateway` or `calendar-gateway` as opposed to just `webhook` or `calendar`. The former is currently demonstrated through the examples. If we change all examples to the latter then it makes sense to add `-gateway` or `-sensor` to the name of the deployments and service. I am leaning towards former because if you consider any core K8s controllers like deployment or replicaset, they also don't `-deployment` or `-replicaset` in the actual pod name.

whynowy: It's not comparable to whether POD has `-deployment` or `-replicaset` in the name, POD always has a `GeneratedName`, it makes sure its name is unique. 

This is the `Name` for a Service, no matter wheter we put `-gateway` in the examples or not, users always can name both Gateway and Sensor `webhook` or a simple `test`, it leads to both Service names are `webhook-svc` or `test-svc`. We put those examples there but we didn't say "you need to strictly follow the examples to name your Gateway and Sensors".

VaibhavPage: I'm fine with adding `-gateway` and `-sensor` to svc as long as we update all examples so that names don't become `webhook-gateway-webhook-gateway-1234` and the service like `webhook-gateway-gateway-svc`.

whynowy: Let me update the examples.

VaibhavPage: 👍

PR #670, opened 2020-05-20
chore: Update Sensor Dockerfile and re-run codegen on master

whynowy:
To make it neat.

### DIFF
--- a/sensors/cmd/Dockerfile
+++ b/sensors/cmd/Dockerfile
@@ -1,20 +1,7 @@
-FROM centos:7
-RUN yum -y install ca-certificates openssh openssh-server openssh-clients openssl-libs curl git
+FROM alpine:latest as certs
+RUN apk --update add ca-certificates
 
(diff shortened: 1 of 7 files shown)


sensors/cmd/Dockerfile
```diff
@@ -1,20 +1,7 @@
-FROM centos:7
-RUN yum -y install ca-certificates openssh openssh-server openssh-clients openssl-libs curl git
-
-# OpenFass CLI
-COPY assets/faas-cli /usr/local/bin/faas
-
-# Argo Workflow CLI
-COPY assets/argo-linux-amd64 /usr/local/bin/argo
```
VaibhavPage, line 8: we need argo cli. you can get rid of OpenFaas cli

whynowy: got it, thanks!

whynowy: done.

PR #671, opened 2020-05-21
feat: Support for using a live K8s resource in standard K8s triggers

jannfis:
This change enables the standard K8s trigger to use a live resource from the K8s cluster as artefact, instead of using the definition of `source`. 

A new option `liveObject` (of type `bool`) is introduced on the trigger template, and if set to true, the specified artefact is used to uniquely identify the resource to fetch from the cluster

```yaml
  triggers:
    - template:
        name: webhook-pod-trigger
        k8s:
          group: apps
          version: v1
          resource: deployments
          operation: update
          liveObject: true
          source:
            resource:
              apiVersion: apps/v1
              kind: Deployment
              metadata:
                name: web-deployment
                namespace: web
          parameters:
            - src:
                dependencyName: test-dep
                contextKey: time
              dest: spec.template.metadata.annotations.events\.argoproj\.io/restartedAt
```
This will fetch the resource of type `deployment` in  `apps/v1` with the name `web-deployment` in namespace `web` from the cluster and then update (or set) the annotation specified in `dest`.

I was initially thinking about implementing it as dedicated artefact store, but since it's really only valuable for the standard K8s trigger, chose to implement it within the trigger.

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -2034,6 +2034,10 @@
         "source"
       ],
       "properties": {
+        "liveObject": {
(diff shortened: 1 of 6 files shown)


ptemmer:
This is cool.
Is this documented somewhere?

PR #674, opened 2020-05-28
feat: Introduce EventBus CRD.

### LINKED ISSUE #619: Introduce EventBus for events delivery
Currently from Gateway to Sensor, there are 2 ways to deliver the messages: 

1.) HTTP POST; 
2.) Gateway -> NATS -> Sensor.

1.) is good for testing purpose, but it's not suitable for PROD deployment;
2.) introduces a pub-sub system in between but it exposes the detail to the user - NATS setting needs to be in both Gateway and Sensor.

1. Introduce a new concept `EventBus` for each namespace, which represents a pub-sub like system configuration, or even an implementation;
2. All the events from Gateway are sent to `EventBus`;
3. Sensors listen to `EventBus` to read events;
4. `EventBus` implementation should be pluggable, options could be `NATS`, `Kafka`, `memory`, `k8s events`, ... recommend `NATS` for PROD deployment;
5. To simplify the implementation and make it secure, `EventBus` should be namespace isolated, which means each namespace uses its own `EventBus`;
6. A namesapce can have multiple `EventBus`, but there is a `default` one;

**Additional context**
Benefits:

- Naturally support pub-sub, 1 event can be subscribed by multiple sensors;
- Make it stable and reliable;
- More secure, eliminate direct HTTP calls in the lifecycle, Sensors don't need to expose a ClusterIP service.
- Make it easy to use, no `subscriber` in Gateway and `Subscription` in Sensor spec.

whynowy:
Partially closes #619.

This change introduces a new CRD named `EventBus`. An EventBus brings a NATS pub-sub service to the system, it will be used for events transmission from event source to sensor in the future.

An `EventBus` spec looks like below:
```
kind: EventBus
metadata:
  name: default
spec:
  nats:
    native:
      size: 3
      auth: token
      persistence: {}
```
Or
```
kind: EventBus
metadata:
  name: default
spec:
  nats:
    native: {}
```
(By default `size=1`, `auth=token`, no `persistence`)

Or
```
kind: EventBus
metadata:
  name: default
spec:
  nats:
    native:
      size: 3
      auth: none
```
Or
```
kind: EventBus
metadata:
  name: default
spec:
  nats:
    exotic:
      url: nats://xxxxx:xxx
      clusterID: xxxxx
      auth: token
      accessSecret:
        key: auth-cred
        name: my-secret
```

A `native` spec will bring up 
1. A NATS StatefulSet
2. A headless `ClusterIP` Service for NATS server
3. A configmap for NATS server config
4. A secret for NATS server auth
5. A secret for NATS client auth (if `auth!=none`)
6. A NATS Streaming StatefulSet (if `persistence` is in place)
7. A headless `ClusterIP` Service for NATS streaming (if `persistence` is in place)
8. A configmap for NATS streaming config (if `persistence` is in place)

An `exotic` spec will make use of the config directly.

Following items will be done in coming up PRs.

1. Namespaced install
2. Persistent options
3. Optional NetworkPolicy creation


controllers/eventbus/controller.go
```diff
@@ -0,0 +1,140 @@
...
+	if !eventBus.DeletionTimestamp.IsZero() {
+		log.Info("deleting eventbus")
+		// Finalizer logic should be added here.
+		r.removeFinalizer(eventBus)
+		return nil
+	}
+	r.addFinalizer(eventBus)
+
+	eventBus.Status.InitConditions()
+	if nats := eventBus.Spec.NATS; nats != nil {
```
VaibhavPage, line 83: can we separate out the event bus status initialization into a separate method?

pkg/apis/eventbus/v1alpha1/types.go
```diff
@@ -0,0 +1,119 @@
...
+	Status common.Status `json:"status,omitempty" protobuf:"bytes,1,opt,name=status"`
+	// Config holds the fininalized configuration of EventBus
+	Config BusConfig `json:"config,omitempty" protobuf:"bytes,2,opt,name=config"`
+}
+
+// NATSBus holds the NATS eventbus information
+type NATSBus struct {
+	// Native means to bring up a native NATS service
+	Native *NativeStrategy `json:"native,omitempty" protobuf:"bytes,1,opt,name=native"`
+	// Exotic holds an exotic NATS config
```
VaibhavPage, line 46: what is exotic configuration?

controllers/eventbus/controller.go
whynowy: 👍 done.

Issue #681, opened 2020-06-05
Flag for resource type eventsource to only watch new events


whynowy:
**Is your feature request related to a problem? Please describe.**
Resource type eventsource watches all the events including those cached history, this makes duplicated messages when the gateway pod restarts.

**Describe the solution you'd like**
A flag in the resource to indicate only wants new messages (which happens after start time)

PR #682, opened 2020-06-05
feat: Add afterStart filter to resource type eventsource

whynowy:
Closes: #681


gateways/server/resource/start.go
```diff
@@ -261,18 +262,21 @@ func FieldSelector(selectors []v1alpha1.Selector) (fields.Selector, error) {
...
 	}
-	if !strings.HasPrefix(uObj.GetName(), filter.Prefix) {
-		return errors.Errorf("resource name does not match prefix. resource-name: %s, prefix: %s", uObj.GetName(), filter.Prefix)
+	uObj := event.Obj.(*unstructured.Unstructured)
+	if len(filter.Prefix) > 0 && !strings.HasPrefix(uObj.GetName(), filter.Prefix) {
+		return false
 	}
 	created := uObj.GetCreationTimestamp()
 	if !filter.CreatedBy.IsZero() && created.UTC().After(filter.CreatedBy.UTC()) {
-		return errors.Errorf("resource is created after filter time. creation-timestamp: %s, filter-creation-timestamp: %s", created.UTC().String(), filter.CreatedBy.UTC().String())
```
whynowy, line 275: It's not good to use `error` as return and log it because it's not an error.

VaibhavPage: Although we need to print the reason why the event did not pass filter, otherwise it is a black box to the user

whynowy: added log.

PR #683, opened 2020-06-05
feat: Add schema validations to CRDs

alexec:
# Changes

* Removed or fixed lint errors.
* Removed redundant Argo CI and e2e code.
* Added a `make start` target to install locally.
* Added CRD validation.


.golangci.yml
```diff
@@ -15,34 +15,25 @@ run:
     - "pkg/apis/eventsources/v1alpha1/zz_generated.deepcopy.go"
 linters:
   enable:
-    - bodyclose
```
alexec, line 18: many linters were failing - so I removed them because there were too many issues to fix - I did fix a couple that looked like code bugs

common/retry.go
```diff
@@ -26,11 +27,27 @@ import (
 // Backoff defines an operational backoff
 type Backoff struct {
 	Duration time.Duration `json:"duration"` // the base duration
-	Factor   float64       `json:"factor"`   // Duration is multiplied by factor each iteration
-	Jitter   float64       `json:"jitter"`   // The amount of jitter applied each iteration
+	Factor   string        `json:"factor"`   // Duration is multiplied by factor each iteration
```
alexec, line 30: this is an API change-  you cannot have `float64` for CRDs

whynowy:
A general question - how to handle the situation that multiple control planes exist in one cluster with different versions (namespaced intall / instancdID), and the CRD validation specs are different?

PR #693, opened 2020-06-10
fix(types)!: Correct pkg and generate .proto files from them

alexec:
See: 

* [go-to-protobuf rules](https://docs.google.com/document/d/18kNPQNAXRdYL7maJp89JwunhGSsokAZ_sBR2L3eD3bo/edit#) 
* https://github.com/kubernetes/code-generator/issues/111


Makefile
```diff
@@ -127,8 +127,26 @@ coverage:
 clean:
 	-rm -rf ${CURRENT_DIR}/dist
 
+$(GOPATH)/bin/go-to-protobuf:
```
whynowy, line 130: This doesn't work when the project is cloned out of $GOPATH, please use the way like

https://github.com/argoproj/argo-events/blob/master/hack/update-codegen.sh

pkg/apis/common/backoff.go
```diff
@@ -0,0 +1,19 @@
+package common
+
+import "time"
+
+// Backoff for an operation
+type Backoff struct {
```
alexec, line 6: all types used by protoc must and should be in pkg

pkg/apis/eventsource/v1alpha1/config_test.go
```diff
@@ -1,4 +1,4 @@
-package fsevent
+package v1alpha1
 
 import (
```
alexec, line 3: all types must and should be in pkg

PR #696, opened 2020-06-11
feat: adding support for multiple events in GitLab webhook

VaibhavPage:


pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -379,26 +379,19 @@ type GitlabEventSource struct {
 	Webhook *webhook.Context `json:"webhook" protobuf:"bytes,1,name=webhook"`
 	// ProjectId is the id of project for which integration needs to setup
 	ProjectId string `json:"projectId" protobuf:"bytes,2,name=projectId"`
-	// Event is a gitlab event to listen to.
+	// Events are gitlab event to listen to.
```
whynowy, line 382: Since you are changing the types, can you update `ProjectId` above to `ProjectID`?

VaibhavPage: Sure

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -377,8 +377,8 @@ type GithubEventSource struct {
 type GitlabEventSource struct {
 	// Webhook holds configuration to run a http server
 	Webhook *webhook.Context `json:"webhook" protobuf:"bytes,1,name=webhook"`
-	// ProjectId is the id of project for which integration needs to setup
-	ProjectId string `json:"projectId" protobuf:"bytes,2,name=projectId"`
+	// ProjectID is the id of project for which integration needs to setup
+	ProjectID string `json:"projectID" protobuf:"bytes,2,name=projectID"`
```
whynowy, line 381: I think only the variable needs to be updated, the one in `json` doesn't need to be.

I just checked an example in `kubernetes` - https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/staging/src/k8s.io/legacy-cloud-providers/azure/auth/azure_auth.go#L50

whynowy: It's quite confusing.... on the same file:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/azure/auth/azure_auth.go#L72

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/azure/auth/azure_auth.go#L48

whynowy: I would suggest to use `ProjectID` in variables, `projectId` in `json:xxx`, it looks like this is more than the other.

VaibhavPage: Lets start using `projectID` in json schema as well. I'll create another PR to update the `Id` mentioned in other event sources to `ID`

PR #701, opened 2020-06-15
fix(types)!: Correct pkg and generate .proto files from them

alexec:
See: 

* [go-to-protobuf rules](https://docs.google.com/document/d/18kNPQNAXRdYL7maJp89JwunhGSsokAZ_sBR2L3eD3bo/edit#) 
* https://github.com/kubernetes/code-generator/issues/111

Replaces #693

# Changes

* Fix lint errors.
* Specific API changes
  * Change `int` to `int32` or `int64`.
  * Change `Size` to `Replicas` or `VolumeSize`.
  * Change `*unstructuted.Unstructured` to `json.RawMessage`.
  * Marked pointer struct (e.g. *corev1.SecretKeySelector` with `omitempty`.
* Fix all `protobuf` tags.


pkg/apis/eventsource/v1alpha1/context.go
```diff
@@ -0,0 +1,18 @@
+package v1alpha1
+
+// Context holds a general purpose REST API context
+type Context struct {
```
whynowy, line 4: `Context` might be too generic, was it specifically for `webhook`? maybe `WebhookContext`?  @VaibhavPage

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -823,50 +817,50 @@ type GitArtifact struct {
...
+	Source string `json:"source" protobuf:"bytes,2,opt,name=source"`
 	// SpecVersion - The version of the CloudEvents specification used by the event.
-	SpecVersion string `json:"specversion" protobuf:"bytes,3,name=specVersion"`
+	SpecVersion string `json:"specversion" protobuf:"bytes,3,opt,name=specversion"`
 	// Type - The type of the occurrence which has happened.
-	Type string `json:"type" protobuf:"bytes,4,name=type"`
+	Type string `json:"type" protobuf:"bytes,4,opt,name=type"`
 	// DataContentType - A MIME (RFC2046) string describing the media type of `data`.
-	DataContentType string `json:"dataContentType" protobuf:"bytes,5,name=dataContentType"`
+	DataContentType string `json:"dataContentType" protobuf:"bytes,5,opt,name=dataContentType"`
```
whynowy, line 856: `datacontenttype `? @VaibhavPage - please confirm.

https://github.com/cloudevents/sdk-go/blob/master/v2/event/eventcontext_v03.go#L35

pkg/apis/eventsource/v1alpha1/context.go
alexec: it was only used here - I could move to common?

whynowy: I think moving it here is ok, but giving a name `WebhookContext` might be better because it is only used for webhook related things.

alexec: done

pkg/apis/sensor/v1alpha1/types.go
alexec: this is unchanged in the PR - fix in another PR?

alexec:
> Let's get this one merged first, as there are so many type changes.
> 
> BTW - we don't have CI to run test cases, `travis` config was deleted in previous PR.

Github Action now runs for tests.

pkg/apis/sensor/v1alpha1/types.go
whynowy: Sounds good.

pkg/apis/eventbus/register.go
```diff
@@ -7,6 +7,6 @@ const (
 	// EventSource constants
 	Kind     string = "EventBus"
 	Singular string = "eventbus"
-	Plural   string = "eventbuses"
+	Plural   string = "eventbus"
```
alexec, line 10: 👍

Issue #662, opened 2020-05-14
Kafka EventSource timeout without TLS


descrepes:
**Describe the bug**
On both 0.14.0 and 0.15.0, the kafka EventSource is not working without TLS.
The gateway timeout while trying to connect to brokers.

**To Reproduce**
1. Create a kafka eventsource without TLS
2. Create a kafka gateway
3. See error

**Expected behavior**
The gateway should connect to the Kafka brokers.

**Environment (please complete the following information):**
 - Version [0.14.0, 0.15.0]

**Additional context**
With the 0.13.0 everything is fine.
I removed this TLS part from the 0.15.0:
https://github.com/argoproj/argo-events/blob/13b38c0d1c0ff07e404a4df50c6d3a391b495311/gateways/server/kafka/start.go#L83-L94

and replaced it with just:
`consumer, err = sarama.NewConsumer([]string{kafkaEventSource.URL}, nil)`

And it's working again without TLS support.

VaibhavPage:
hmm, the config TLS should have been empty if the TLS in the event source is not specified.

PR #704, opened 2020-06-16
fix: Fix kafka gateway

zachaller:
We we always getting a failure to connect to kafka with non tls this fixes that after looking a bit into the sarama docs look like config needs to be nil when passed in or needs to have its NewConfig function called.


zachaller:
Closes #611

zachaller:
Closes #662

PR #712, opened 2020-06-22
chore: Deprecation warning fix

nikopen:
spec.template.container causes log floods on both sensor and gateway controllers,

please document these small changes with version releases

### DIFF
--- a/controllers/gateway/resource.go
+++ b/controllers/gateway/resource.go
@@ -88,7 +88,7 @@ func (ctx *gatewayContext) buildLegacyServiceResource() (*corev1.Service, error)
 func (ctx *gatewayContext) makeDeploymentSpec() (*appv1.DeploymentSpec, error) {
 	// Deprecated spec, will be unsupported soon.
 	if ctx.gateway.Spec.Template.Spec != nil {
-		ctx.logger.WithField("name", ctx.gateway.Name).WithField("namespace", ctx.gateway.Namespace).Warn("spec.template.spec is DEPRECATED, it will be unsupported soon, please use spec.template.container")
+		ctx.logger.WithField("name", ctx.gateway.Name).WithField("namespace", ctx.gateway.Namespace).Warn("spec.template.spec is DEPRECATED, it will be unsupported soon, please use spec.template.containers")
 		return ctx.makeLegacyDeploymentSpec()
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/gateway/resource.go
```diff
@@ -88,7 +88,7 @@ func (ctx *gatewayContext) buildLegacyServiceResource() (*corev1.Service, error)
 func (ctx *gatewayContext) makeDeploymentSpec() (*appv1.DeploymentSpec, error) {
 	// Deprecated spec, will be unsupported soon.
 	if ctx.gateway.Spec.Template.Spec != nil {
-		ctx.logger.WithField("name", ctx.gateway.Name).WithField("namespace", ctx.gateway.Namespace).Warn("spec.template.spec is DEPRECATED, it will be unsupported soon, please use spec.template.container")
+		ctx.logger.WithField("name", ctx.gateway.Name).WithField("namespace", ctx.gateway.Namespace).Warn("spec.template.spec is DEPRECATED, it will be unsupported soon, please use spec.template.containers")
```
whynowy, line 91: Why change to `containers`? it should be `container`.

controllers/sensor/resource.go
```diff
@@ -83,7 +83,7 @@ func (ctx *sensorContext) serviceBuilder() (*corev1.Service, error) {
 func (ctx *sensorContext) makeDeploymentSpec() (*appv1.DeploymentSpec, error) {
 	// Deprecated spec, will be unsupported soon.
 	if ctx.sensor.Spec.Template.Spec != nil {
-		ctx.logger.WithField("name", ctx.sensor.Name).WithField("namespace", ctx.sensor.Namespace).Warn("spec.template.spec is DEPRECATED, it will be unsupported soon, please use spec.template.container")
+		ctx.logger.WithField("name", ctx.sensor.Name).WithField("namespace", ctx.sensor.Namespace).Warn("spec.template.spec is DEPRECATED, it will be unsupported soon, please use spec.template.containers")
```
whynowy, line 86: Same question here.

whynowy:
In regards to the log floods, it will not be existing in the future as we are changing the controller implementation, also you are right, we should have those things documented. Thanks for suggestions!

controllers/gateway/resource.go
nikopen: `container` is what generates the log floods, ` containers` has no problem

whynowy: I didn't get you.

1. Making a log content change will not change anything.
2. The reason why the warning appears is because a to be deprecated spec `spec.template.spec.xxx` is being used, it recommends to use new spec `spec.template.container`.
3. `spec.template.containers` is never supported, I guess you meant `spec.template.spec.containers`, which applies to the one to be deprecated.

PR #715, opened 2020-06-23
build: added clusterworkflowtemplate to argo-events rbac manifests, upgraded argo version

iMoses:
currently argo-events don't have any permissions over `clusterworkflowtemplates`, i assume due to the feature being relatively new

this won't do it by itself, we also need to upgrade the argo cli version argo-events is using to a version which supports `clusterworkflowtemplates`

### DIFF
--- a/sensors/cmd/Dockerfile
+++ b/sensors/cmd/Dockerfile
@@ -1,13 +1,20 @@
 FROM centos:7
+
+ARG ARGO_VERSION=v2.8.2
+
 RUN yum -y install ca-certificates openssh openssh-server openssh-clients openssl-libs curl git
 
 # OpenFass CLI
 COPY assets/faas-cli /usr/local/bin/faas
 
 # Argo Workflow CLI
-COPY assets/argo-linux-amd64 /usr/local/bin/argo
+RUN curl -sSL -o /usr/local/bin/argo \
+    https://github.com/argoproj/argo/releases/download/${ARGO_VERSION}/argo-linux-amd64 \
+ && chmod +x /usr/local/bin/argo
 
-RUN argo version
+# added `|| true` to ignore "invalid configuration: no configuration has been provided"
+# TODO: argo version shouldn't rely on a cluster connection to execute
+RUN argo version || true
 RUN faas version
 
 RUN mkdir /.openfaas
--- a/sensors/cmd/Dockerfile
+++ b/sensors/cmd/Dockerfile
@@ -5,6 +5,7 @@ ARG ARGO_VERSION=v2.8.2
 RUN yum -y install ca-certificates openssh openssh-server openssh-clients openssl-libs curl git
 
 # OpenFass CLI
+# TODO: donwload faas instead of copying it from the svc, same as with argo-cli
 COPY assets/faas-cli /usr/local/bin/faas
 
 # Argo Workflow CLI
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/Shopify/sarama v1.26.1
 	github.com/ahmetb/gen-crd-api-reference-docs v0.2.0
 	github.com/apache/openwhisk-client-go v0.0.0-20190915054138-716c6f973eb2
-	github.com/argoproj/argo v2.5.2+incompatible
+	github.com/argoproj/argo v2.8.2
 	github.com/argoproj/argo-cd v1.5.1
 	github.com/argoproj/argo-rollouts v0.7.2
 	github.com/argoproj/pkg v0.0.0-20200319004004-f46beff7cd54 // indirect
--- a/manifests/cluster-install/rbac/argo-events-cluster-role.yaml
+++ b/manifests/cluster-install/rbac/argo-events-cluster-role.yaml
@@ -42,6 +42,15 @@ rules:
       - eventbus
       - eventbus/finalizers
   - apiGroups:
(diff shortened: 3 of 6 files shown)


whynowy:
Please run `make crds` to regenerate the final manifests. Also, please include the `Dockerfile` change in the same PR, BTW - thanks for raising the issue and fixing it!

iMoses:
I found an issue this `argo version` causing it to raise an error if no cluster configuration is found, even though i don't see a reason for this dependency (so i'm currently assuming it's a bug)
this is why i changed the Dockerfile command to `argo version || true` - to ignore this error

I also ran `make crds` but the files didn't change so there's nothing new to push :)

whynowy:
> I found an issue with `argo version` causing it to raise an error if no cluster configuration is found, even though i don't see a reason for this dependency (so i'm currently assuming it's a bug)
> this is why i changed the Dockerfile command to `argo version || true` - to ignore this error
> 
> I also ran `make crds` but the files didn't change so there's nothing new to push :)

My bad, the command to regenerate manifests is `make manifests`

iMoses:
@whynowy okay, pushed the updated manifests

sensors/cmd/Dockerfile
```diff
@@ -1,13 +1,21 @@
...
 # Argo Workflow CLI
-COPY assets/argo-linux-amd64 /usr/local/bin/argo
+RUN curl -sSL -o /usr/local/bin/argo \
+    https://github.com/argoproj/argo/releases/download/${ARGO_VERSION}/argo-linux-amd64 \
+ && chmod +x /usr/local/bin/argo
 
-RUN argo version
+# added `|| true` to ignore "invalid configuration: no configuration has been provided"
+# TODO: argo version shouldn't rely on a cluster connection to execute
+RUN argo version || true
```
whynowy, line 18: I believe originally `argo version` here was used to check if `argo` binary was existing, adding `|| true` makes it meaningless. Since there's an unexpected error with latest `argo` cli, I suggest to remove this line for now, meanwhile file a bug in argo workflow community.

iMoses: It still outputs as part of the build the version of argo used so we may wanna keep it for that, but if you still prefer it i'll comment it out..

gladdiologist:
hi @iMoses ,
to fix golint: go get github.com/argoproj/argo@release-2.8

iMoses:
> go get github.com/argoproj/argo@release-2.8

this make sense?
```
	github.com/argoproj/argo v0.0.0-20200622215353-c15e817b2fa6
```
that's what i'm getting form it..

gladdiologist:
yes, perfect. I don't know why we can't target with v2.8.2 (go stuff). But it will build now.
May be this shoud be reviewed by contributors, regarding the best practices.

whynowy:
https://github.com/argoproj/argo/issues/3383

tmshn:
Allow me to jump in here: do you have any plan to upgrade to 2.9 rather than 2.8?
It's released yesterday: https://github.com/argoproj/argo/releases/tag/v2.9.0

whynowy:
@iMoses @gladdiologist @tmshn  - I'd like to give some update about recent changes. 

You might have noticed, we are working on some enhancements on argo-events, including using least privilege to run Sensor deployment, and this has been merged into master in https://github.com/argoproj/argo-events/commit/4e8cf29dfe17cae6e9c8b7f119fe81c3abd4aa4a. With this change, you are not required to run sensor with service account `argo-events-sa`.

- If your sensor triggers the action not related to k8s at all, such as HTTP endpoint, or AWS Lambda, Kafka, you don't need to specify a service account which means `default` is good enough.
- If your sensor triggers k8s resources such as `workflow`, or `deployment`, then you need to give a service account with proper privileges (for example, create `workflow`, update `deployment`), but we will not create such a service account for you, you should configure a SA that has the most security based on your own use case.

According to these, adding privilege to `argo-events-sa` does not apply any more, but we still need to upgrade `argo` version for argo-events.

Here is the proposals of recent changes.
https://docs.google.com/document/d/1uPt2DyvzObEzZVbREqjW-o1gKCZ_hB8QS5Syw7_MLT8/edit?usp=sharing

tmshn:
Thank you for the info, the proposed design looks better and more k8s-native!
Do you have any planned timeline for this?

> but we still need to upgrade `argo` version for argo-events.

Actually this is another point I care. We are using argo+argo-events, and this version skew prevents us from using some new features of argo.

whynowy:
> Thank you for the info, the proposed design looks better and more k8s-native!
> Do you have any planned timeline for this?
> 
> > but we still need to upgrade `argo` version for argo-events.
> 
> Actually this is another point I care. We are using argo+argo-events, and this version skew prevents us from using some new features of argo.

1. We will have a 1.0 release in Aug, which includes the features in the doc.
2. We will revisit the `argo` dependency part later, ideally argo-events should not have hard dependency on `argo` source code except `argo` binary version. For now, the easiest approach is to upgrade the version.

whynowy:
https://github.com/argoproj/argo-events/issues/733

whynowy:
Argo CLI upgrade has been address in https://github.com/argoproj/argo-events/pull/758, thanks!

iMoses:
@whynowy you forgot to update the `clusterworkflowtemplates` as well :)

whynowy:
@iMoses - thanks, will add it.

PR #735, opened 2020-07-11
feat: Merge Gateway and EventSource

whynowy:


controllers/eventsource/controller.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type reconciler struct {
+	client client.Client
+	scheme *runtime.Scheme
+
+	// controller namespace
+	namespace        string
+	eventSourceImage string
+	logger           logr.Logger
```
VaibhavPage, line 34: is there a reason why we are using different log packages like logrus, zap and go-logr?

whynowy: `go-logr` is not a real log package but an interface, the options are just `logrus` and `zap`, I think we should pick `logrus`. The reason here using `zap` is just I didn't get time to figure out how to integrate `logrus` with `go-logr` ( log interface used by `controller-runtime`), but I know it's doable. This will be done later.

whynowy: Or we don't use `go-logr` at all but `logrus` directly in controllers. This shouldn't be a big deal, we can easily change it.

whynowy: As we discussed, will switch to `zap` later.

PR #737, opened 2020-07-12
feat(sensor)!: start & stop is required, and allow stop < start in time-filter

### LINKED ISSUE #729: Support time filter which spans across UTC midnight
Sensor dependencies can have a time filter so that workflows are triggered only by events during specific time range, such as 9AM-11AM. However, you cannot specify a time range that spans across midnight, like 11PM-2AM.
We should support such use cases.

Due to this, some of business requirements cannot be expressed by the time filter, simply because its time range position (note that the time range must be written in UTC).
The notion of "day" is a circular thing, rather than has a concrete and global start/end.

Allow `start > stop`, in which case, the day of the stop time is treated as next day of one of the start time.

```yaml
filters:
  time:                # E.g. (on 2020-07-08)
    start: "23:00:00"  # 2020-07-08 23:00:00
    stop:  "02:00:00"  # 2020-07-09 02:00:00
```

- pros:
    - looks natural
- cons:
    - hard for users to notice the bug that swapping start and stop (e.g.: intended 9AM-11AM (2 hours), but actually written 11AM-9AM (22 hours)).

Another choice is to allow hour of `stop` to be in [0, 48) while keeping the restriction of `start < stop`.

```yaml
filters:
  time:                # E.g. (on 2020-07-08)
    start: "23:00:00"  # 2020-07-08 23:00:00
    stop:  "26:00:00"  # 2020-07-09 02:00:00
```

- pros:
    - less bug-prone of swapping start and stop
- cons:
    - looks relatively natural for say 26:00 (2AM), but weird to see like 33:00 (9AM)

One more choise: introduce `duration` field instead of `stop`

```yaml
filters:
  time:                  # E.g. (on 2020-07-08)
    start:    "23:00:00" # 2020-07-08 23:00:00
    duration: "3h"       # 2020-07-09 02:00:00
    # or maybe
    duration: "3:00:00"
```

- pros:
    - looks natural
    - less bug-prone of swapping start and stop
- cons:
    - if format is like `3h` which is supported by `ParseDuration`, hard to represent granular time range. If format is like `3:00:00`, bit confusing among stop vs. duration

**Additional context**
In any case, internal implementation will be quite easy, so the decision can be made solely on ergonomic aspect.

I'm happy to fix this.

tmshn:
## Changed

- support time ranges span midnight by allowing stop to be smaller than start (fix #729)
- making both start and stop required for time filter

### ~Additional suggestion~ (Edit: this suggestion adopted)

If you have a chance to introduce any breaking changes, I suggest to make both `Start` and `Stop` required (while keeping time filter itself is optional).

The rationale behind this is: 

- the cost of specifying both Start and Stop is trivial
    - due to its repetitive (circular) nature of time filter, users are already required to be aware of its start/stop even though they specified only one.
- making things optional lead the situation less intuitive (unless the default behavior is super explicit)
    - specifying only one side of "range" sounds like "the range never ends", which is not the case for the time filter
- it makes the implementation much simpler

I guess they are optional simply because they used to be non-repetitive thing (not time but date+time) and the default values were ±inf. The situation has changed.


controllers/sensor/validate.go
```diff
@@ -469,32 +468,32 @@ func validateEventFilter(filter *v1alpha1.EventDependencyFilter) error {
...
 		}
-		if stopTime.Before(startTime) || startTime.Equal(stopTime) {
-			return errors.Errorf("invalid event time filter: stop '%s' is before or equal to start '%s", tFilter.Stop, tFilter.Start)
+		if stopTime.Equal(startTime) {
+			return errors.Errorf("invalid event time filter: stop '%s' is equal to start '%s", tFilter.Stop, tFilter.Start)
 		}
+		return nil
+	}
+
+	if tFilter.Start != "" {
```
whynowy, line 487: If both `Start` and `Stop` == `""`, they will be parsed twice, should be able to only do once.

Also need to check if both of them `==""`.

sensors/dependencies/filter.go
```diff
@@ -56,50 +56,78 @@ func filterEvent(filter *v1alpha1.EventDependencyFilter, event *v1alpha1.Event)
...
-// applyTimeFilter checks the eventTime against the timeFilter:
-// 1. the eventTime is greater than or equal to the start time
-// 2. the eventTime is less than the end time
-// returns true if 1 and 2 are true and false otherwise
+// filterTime checks the eventTime falls into time range specified by the timeFilter.
+// Start is inclusive, and Stop is exclusive.
+//
+// if Start < Stop: eventTime must be in [Start, Stop)
+//
+//   0:00        Start       Stop        0:00
```
whynowy, line 64: Very nice comments!

sensors/dependencies/filter.go
```diff
@@ -56,50 +56,78 @@ func filterEvent(filter *v1alpha1.EventDependencyFilter, event *v1alpha1.Event)
...
+	if timeFilter == nil {
+		return true, nil
+	}
 
-			stopTime, err := time.Parse(common.StandardTimeFormat, fmt.Sprintf("%s %s", currentTime, timeFilter.Stop))
-			if err != nil {
-				return false, err
-			}
-			stopTime = stopTime.UTC()
+	if timeFilter.Start != "" && timeFilter.Stop != "" {
```
whynowy, line 96: Parse both `Start` and `Stop` before doing the real `if, else` logic to simplify the code.

controllers/sensor/validate.go
tmshn: No, the entire structure of this function is like

```go
if Start != "" && Stop != "" {
	// parse Start and Stop
	return ...
}
if Start != "" { // it means `Stop == ""`
	// parse only Start
	return ...
}
if Stop != "" { // it means `Start ==""`
	// parse only Stop
	return ...
}
return nil
```

If this looks less straightforward, maybe this'd be better?

```go
if Start != "" && Stop != "" {
	// parse Start and Stop
	return ...
} else if Start != "" { // it means `Stop == ""`
	// parse only Start
	return ...
} else if Stop != "" { // it means `Start ==""`
	// parse only Stop
	return ...
} else {
	return nil
}
```

sensors/dependencies/filter.go
tmshn: Is this what you mean? 1d37e962abc831ed3af3c873409fd84d5385df38

tmshn:
Will add doc and example!

tmshn:
Docs added, but examples not yet.

---

## A few ceveats on running `make codegen`

- `mockery` must be in v1, otherwise `-name` option won't work. The latest one from homebrew is [v2](https://github.com/vektra/mockery/releases/tag/v2.0.0), which uses `--name` (two hyphens)
    - How to fix? ...... `mockery` v1 also seem to work with two hyphens, so I believe `s/-name/--name/` will do
- `kustomize` must be [v3.2](https://github.com/kubernetes-sigs/kustomize/releases/tag/v3.2.0)+, otherwise inline `patchesStrategicMerge` [won't work](https://github.com/kubernetes-sigs/kustomize/issues/1525#issuecomment-530448557)
    - How to fix? ...... documentation or version check is needed
- `sed` must be BSD, otherwise `-i ""` won't work. In GNU sed, in-place suffix must be, if any, provided as single argument like `-i.bak` or provided without suffix `-i`.
    - How to fix? ...... check if sed is BSD version or GNU version, and change arguments
- `./hack/generate-proto.sh` fails with the following error,
    - How to fix? ...... I don't know the "root cause" of the error, but the path below worked for me

```
2020/07/14 00:07:25 protoc-gen-gogo: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--gogo_out: protoc-gen-gogo: Plugin failed with status code 1.
2020/07/14 00:33:14 protoc -I . -I /var/folders/z0/p22kh2xn5klf_wvdkscmj8yc0000gn/T/tmp.S0XesL1FD6/src -I ./vendor --gogo_out=/var/folders/z0/p22kh2xn5klf_wvdkscmj8yc0000gn/T/tmp.S0XesL1FD6/src /var/folders/z0/p22kh2xn5klf_wvdkscmj8yc0000gn/T/tmp.S0XesL1FD6/src/github.com/argoproj/argo-events/vendor/k8s.io/apimachinery/pkg/util/intstr/generated.proto
2020/07/14 00:33:14 Unable to generate protoc on k8s.io.apimachinery.pkg.util.intstr: exit status 1
```

```diff
diff --git a/hack/generate-proto.sh b/hack/generate-proto.sh
index 374735d..e5ae655 100755
--- a/hack/generate-proto.sh
+++ b/hack/generate-proto.sh
@@ -16,6 +16,8 @@ export GO111MODULE="off"
 cd "${FAKE_REPOPATH}"
 
 go install ./vendor/k8s.io/code-generator/cmd/go-to-protobuf
+go install ./vendor/github.com/gogo/protobuf/protoc-gen-gogo
+export PATH="${GOPATH}/bin:${PATH}"
 
 ${GOPATH}/bin/go-to-protobuf \
         --go-header-file=./hack/custom-boilerplate.go.txt \
```

api/openapi-spec/swagger.json
```diff
@@ -2434,15 +2434,15 @@
       }
     },
     "io.argoproj.sensor.v1alpha1.TimeFilter": {
-      "description": "TimeFilter describes a window in time. DataFilters out event events that occur outside the time limits. In other words, only events that occur after Start and before Stop will pass this filter.",
+      "description": "TimeFilter describes a window in time. It filters out events that occur outside the time limits. In other words, only events that occur after Start and before Stop will pass this filter.",
```
tmshn, line 2437: Auto-generated file

api/sensor.html
```diff
@@ -2374,7 +2374,7 @@ <h3 id="argoproj.io/v1alpha1.TimeFilter">TimeFilter
 </p>
 <p>
 <p>TimeFilter describes a window in time.
-DataFilters out event events that occur outside the time limits.
+It filters out events that occur outside the time limits.
```
tmshn, line 2377: Auto-generated file

api/sensor.md
```diff
@@ -4646,9 +4646,9 @@ TimeFilter
 
 <p>
 
-TimeFilter describes a window in time. DataFilters out event events that
-occur outside the time limits. In other words, only events that occur
-after Start and before Stop will pass this filter.
+TimeFilter describes a window in time. It filters out events that occur
+outside the time limits. In other words, only events that occur after
+Start and before Stop will pass this filter.
```
tmshn, line 4651: Auto-generated file

manifests/install.yaml
```diff
@@ -265,37 +265,15 @@ subjects:
...
-    \ # namespace is used for namespaced installation, if it is specified, the controller
-    only \n  # watches gateways in this namespace.\n  namespace: <your_namespace>\n"
+  _example: "#\n# This is not real configuration but an example.\n# Copy this block out of \"_example\" and unindent it \n# under \"data\" to make it functional.\n#\nconfig: |\n  # instanceID is a label selector to limit the controller's watch to a specific instance. It\n  # contains an arbitrary value that is carried forward into its pod labels, under the key\n  # gateways.argoproj.io/gateway-controller-instanceid, for the purposes of date segregation.\n  # This enables a controller to only watch the gateway that it is interested about,\n  # in order to support multiple controllers in a single cluster, and ultimately allows the\n  # controller itself to be bundled as part of a higher level application. If omitted, the\n  # controller watches gateways that *are not* labeled with an instance id.\n  instanceID: <instanceID>\n\n  # namespace is used for namespaced installation, if it is specified, the controller only \n  # watches gateways in this namespace.\n  namespace: <your_namespace>\n"
```
tmshn, line 268: This is also auto-generated file
idk why there are such diffs, added as-is...

pkg/apis/sensor/v1alpha1/generated.proto
```diff
@@ -596,18 +596,20 @@ message Template {
 }
 
 // TimeFilter describes a window in time.
-// DataFilters out event events that occur outside the time limits.
+// It filters out events that occur outside the time limits.
```
tmshn, line 599: Auto-generated file

tmshn: Maybe it's good idea to use `.gitattributes` to mark them as \**generated*\*. It makes reviewers life easier.
https://docs.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github

E.g.:

```gitattributes
**/generated.proto linguist-generated=true
```

sensors/dependencies/filter.go
whynowy: How about this:

```
var startTime, stopTime time.Time
var err Error
if timeFilter.Start != "" {
  startTime, err = common.ParseTime(timeFilter.Start, eventTime)
  if err != nil {
    return false, err
  }
}
// same for timeFilter.Stop
xxxx

// real logic
if timeFilter.Start != "" && timeFilter.Stop != "" {
  .....
```

pkg/apis/sensor/v1alpha1/generated.proto
whynowy: 👍 , good to know this, would you please come out another PR for it?

sensors/dependencies/filter.go
tmshn: Ah I see!

tmshn: ✅ dccbd181c189c01c87f51dc6a5d1c076a5c9a3d8

pkg/apis/sensor/v1alpha1/generated.proto
tmshn: ✅ https://github.com/argoproj/argo-events/pull/739

tmshn:
It seems there's already breaking changes introduced on the head, so good chance to think about my suggestion to make start/stop required (described on my top comment), isn't it?

VaibhavPage:
I am fine introducing `start/stop` change as suggested in the Additional Suggestion section.

tmshn:
I've made start/stop required: add3a27dd0196b0a73bcc29c6cc190d9bdefda42

PR #739, opened 2020-07-14
chore: mark some files as auto-generated

tmshn:
Those files are auto-generated by `make codegen`, so their diffs should be minimized to make reviewers' life easier!

refs:

- How to mark files as "auto-generated": [Customizing how changed files appear on GitHub - GitHub Docs](https://docs.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github)
- How to confirm a file is marked as intended: [Git - git-check-attr Documentation](https://git-scm.com/docs/git-check-attr)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+/api/**/swagger.json    linguist-generated=true
+/api/*.html             linguist-generated=true
+/api/*.md               linguist-generated=true
+/manifests/install.yaml linguist-generated=true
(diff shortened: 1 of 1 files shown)


tmshn:
🤔 Why is CLA status check pending? Am I missing something?

VaibhavPage:
was there any commit with an account different than one that signed CLA?

tmshn:
The email was correct...
But anyway, I just amended the commit and force-pushed, now it looks okay.

Issue #742, opened 2020-07-15
GCP Pub/Sub auto mount credentials from Secret


tmshn:
**Is your feature request related to a problem? Please describe.**
To use GCP Pub/Sub with JSON credentials, you have to write numbers of lines like this, which us quite bothersome:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: EventSource
metadata:
  name: gcp-pubsub
spec:
  pubsub:
    projectID: foo
    topicID: bar
    credentialsFile: /gcp-pubsub-creds-dir/serviceAccount.json
  template:
    container:
      volumeMounts:
        - mountPath: /gcp-pubsub-creds-dir
          name: gcp-pubsub-data
    volumes:
      - name: gcp-pubsub-data
        secret:
          secretName: gcp-pubsub
```

**Describe the solution you'd like**

The manifest can be simplified like this:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: EventSource
metadata:
  name: gcp-pubsub
spec:
  pubsub:
    projectID: foo
    topicID: bar
    credentials:
      secretName: gcp-pubsub
      # secretKey: serviceAccount.json # optional
```

Event source controller mount the secret to the event source pod appropriately according to this.

**Describe alternatives you've considered**
None.

**Additional context**
None.

whynowy:
@tmshn - appreciate if you can contribute on this.

tmshn:
Yeah, I put this one at the end of my task queue, and then #743. Will work on some weekends.

PR #748, opened 2020-07-16
feat(eventsource)!: auto-mount secret for GCP PubSub credentials

tmshn:
fix: #742 

todos:

- [ ] install this to real cluster and test
- [ ] add more cases to validate_test.go


controllers/eventsource/resource.go
```diff
@@ -409,3 +414,36 @@ func envFromSecretsOrConfigMaps(source interface{}, t reflect.Type) []corev1.Env
 	}
 	return result
 }
+
+func buildGcpPubsubCredentialsVolumes(pubsubs map[string]v1alpha1.PubSubEventSource) ([]corev1.Volume, []corev1.VolumeMount) {
```
whynowy, line 418: Don't need this. All the `secretKeySelector` and `configMapKeySelector` will be automatically added to `envFrom`, as a result you can get the secret from ENV. 

See `func envFromSources`, and refer to other event sources with credential secrets.

tmshn: So is it better to inject secret via environmental variable, rather than file?

whynowy: It is not better or worse, but easier.

Also, we have more than 20 types of event sources, about half of them have secrets need to inject, if use volume for them, we should also use an unified way instead of doing it one by one.

tmshn: Ah I see! Your point makes sence.
I'll switch to use env var so that it'll be aligned better with other even source.

whynowy:
@tmshn - would you like to finish this?

tmshn:
@whynowy Sorry to leave this still wip, I couldn't take enough time to spare for this. I'll prioritize this on my work.
BTW, do you have a fixed timeline towards 1.0 in you mind? (I'd like to see this on 1.0)

whynowy:
@tmshn - I have got it done as part of #778

tmshn:
@whynowy Great! So I close this one.

PR #749, opened 2020-07-16
fix(eventbus): NATS eventbus auto reconnect and code refactory

whynowy:
1. Fixed a bug that `()` in sensor dependency expression was trimmed when converting group expr to dependency expr.
2. 2 layers of event bus auto reconnect: 
  a. utilize NATS reconnect feature - for short term connection lost such as NATS POD crash
  b. Extra Goroutins in eventsource and sensor service to auto reconnect - for long term connection lost
3. Code refactory to use `ctx.Done` instead of `stopCh`

### DIFF
--- a/eventbus/driver/nats.go
+++ b/eventbus/driver/nats.go
@@ -73,7 +73,6 @@ func (n *natsStreaming) Connect() (Connection, error) {
 	log := n.logger.WithField("clientID", n.clientID)
 	opts := []nats.Option{
 		nats.MaxReconnects(-1),
-		// TODO: somehow reconnection doesn't work here
 		nats.ReconnectWait(3 * time.Second),
 		nats.DisconnectErrHandler(func(nc *nats.Conn, err error) {
 			log.Errorf("NATS connection lost, reason: %v", err)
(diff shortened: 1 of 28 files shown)


eventbus/driver/nats.go
```diff
@@ -73,7 +71,16 @@ func NewNATSStreaming(url, clusterID, subject, clientID string, auth *Auth, logg
 
 func (n *natsStreaming) Connect() (Connection, error) {
 	log := n.logger.WithField("clientID", n.clientID)
-	opts := []nats.Option{}
+	opts := []nats.Option{
+		nats.MaxReconnects(-1),
```
VaibhavPage, line 75: what's the significance of `-1`?

eventsources/eventing.go
```diff
@@ -261,7 +286,7 @@ func (e *EventSourceAdaptor) Start(ctx context.Context, stopCh <-chan struct{})
 				return err
 			}
 			go func(s EventingServer) {
-				err := s.StartListening(cctx, stopCh, func(data []byte) error {
+				err := s.StartListening(cctx, func(data []byte) error {
```
VaibhavPage, line 289: You can now remove // TODO: Need Retry

whynowy: done.

eventbus/driver/nats.go
whynowy: `-1` means retry forever.

PR #755, opened 2020-07-17
feat: Different deployment update strategy for different event sources

### LINKED ISSUE #752: EventSource deployment strategy and replicas control
The default EventSource deployment strategy is `rollingUpdate`, it works for most of event sources, but doesn't for the ones who do not allow multiple clients running at the same time, for example, multiple NATS clients with same client ID are not allowed. 

Same problem applies for replicas.

For the eventsources like that,
1. use `recreate` strategy;
2. replicas always set to 1.

**Additional context**

whynowy:
1.  Use `recreate` strategy for some of the eventsources deployment update.
2. Now EventSource allows multiple types of sources defined in one spec, following the rules:
    a.) **Event source with `rollingUpdate` and `recreate` are not allowed in one EventSource object** - Otherwise when the EventSource is updated, there will be downtime for those `rollingUpdate` type event sources such as `webhook`.
    b.) **Event name (the map key) needs to be unique in one EventSource object, even different type of event source can not use same name** - Because we use EventSourceName and EventName in Sensor as dependency.
    c.) `rollingUpdate` types use replicas defined in the spec, defaults to 1;
    d.) `recreate` ignores replica field in the spec, it always uses `1`
3. **Added missing `Resource` type EventSource.**
4. Fixed a bug for `HDFS` event source - a bug of `configMapSelector` to `envFrom` conversion.
5. Added eventsource-controller test cases.

Closes #752

### DIFF
--- a/controllers/eventsource/cmd/main.go
+++ b/controllers/eventsource/cmd/main.go
@@ -76,7 +76,7 @@ func main() {
 	}
 	// A controller with DefaultControllerRateLimiter
 	c, err := controller.New(eventsource.ControllerName, mgr, controller.Options{
-		Reconciler: eventsource.NewReconciler(mgr.GetClient(), mgr.GetScheme(), namespace, eventSourceImage, log.WithName("reconciler")),
+		Reconciler: eventsource.NewReconciler(mgr.GetClient(), mgr.GetScheme(), eventSourceImage, log.WithName("reconciler")),
 	})
 	if err != nil {
 		mainLog.Error(err, "unable to set up individual controller")
--- a/pkg/apis/common/common.go
+++ b/pkg/apis/common/common.go
@@ -57,6 +57,7 @@ var (
 		MQTTEvent,
 		EmitterEvent,
 		NSQEvent,
+		ResourceEvent,
 	}
 )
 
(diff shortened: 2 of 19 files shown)


pkg/apis/common/common.go
```diff
@@ -46,6 +46,17 @@ var (
 	GenericEvent     EventSourceType = "generic"
 )
 
+var (
+	// RecreateStrategyEventSources refers to the list of event source types
+	// that need to use Recreate strategy for its Deployment
+	RecreateStrategyEventSources = []EventSourceType{
```
whynowy, line 52: @VaibhavPage - Could you point me the right types here?

controllers/eventsource/cmd/main.go
```diff
@@ -76,7 +76,7 @@ func main() {
 	}
 	// A controller with DefaultControllerRateLimiter
 	c, err := controller.New(eventsource.ControllerName, mgr, controller.Options{
-		Reconciler: eventsource.NewReconciler(mgr.GetClient(), mgr.GetScheme(), namespace, eventSourceImage, log.WithName("reconciler")),
```
whynowy, line 79: This `namespace` is never used, remove it.

PR #757, opened 2020-07-18
docs: update documentation for v0.17.0

VaibhavPage:


manifests/base/gateway-controller/gateway-controller-deployment.yaml
```diff
@@ -1,25 +0,0 @@
-apiVersion: apps/v1
```
whynowy, line 1: I think we should keep the gateway-controller running for some time, it is used to clean up old generated gateway deployment and service. Or you have a better way for upgrade?

VaibhavPage: I have the gateway controller removed from normal installation/manifests because for new users, we want them to just install the latest components. I'll add a gateway-controller manifest in the `migration path` documentation.

examples/eventbus/native.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: argoproj.io/v1alpha1
```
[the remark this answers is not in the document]
VaibhavPage: Instead of removing, I'll comment it.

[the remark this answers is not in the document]
whynowy: Nice, then more information should be added, `persistence` and `antiAffinity` are more important.

```
kind: EventBus
metadata:
  name: default
spec:
  nats:
    native:
      replicas: 3
      auth: token
      antiAffinity: false
      persistence:
        storageClassName: standard
        accessMode: ReadWriteOnce
        volumeSize: 10Gi
```

manifests/base/gateway-controller/gateway-controller-deployment.yaml
whynowy: 👍

examples/event-sources/azure-events-hub.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: EventSource
 metadata:
-  name: azure-events-hub-event-source
+  name: azure-events-hub
 spec:
   type: azureEventsHub
```
whynowy, line 6: There's not a `type` field in EventSource, and we don't need it.

VaibhavPage: removed from examples

docs/tutorials/09-service-account.md
```diff
@@ -0,0 +1,21 @@
...
+## K8s Resource Trigger
+
+* To execute Argo workflow trigger, make sure to grant `create` permission for workflows to the service account.
+
+* To trigger a any other K8s resource, make sure to grant `create` permission for that resource. 
+
+## AWS Lambda, HTTP, Slack and OpenWhisk Trigger
+
+These triggers may need access to secrets for access tokens/auth related configuration. Make
+sure to grant `get` and `list` permissions for the secret resource.
```
whynowy, line 17: This is something we can enhance later, just do the same thing as we do for event sources.

docs/tutorials/09-service-account.md
```diff
@@ -0,0 +1,21 @@
+# Service Account for Triggers
```
whynowy, line 1: Is `service account for eventsourcs` missing?

examples/event-sources/resource.yaml
```diff
@@ -1,9 +1,8 @@
 apiVersion: argoproj.io/v1alpha1
 kind: EventSource
 metadata:
-  name: resource-event-source
+  name: resource
 spec:
-  type: resource
```
whynowy, line 6: Need to have service account in `resource` type event source.

Issue #738, opened 2020-07-13
The "metrics" field is dropped by the sensor when spawning a new workflow


benhackett4:
Argo supports a `metrics` field on workflow objects. (For [example](https://github.com/argoproj/argo/blob/7581025ffac0da6a4c9b125dac3173d0c84aef4f/examples/custom-metrics.yaml#L10-L19)). When I create a sensor object that contains a `metrics` field in the workflow template object, it appears fine when I do `kubectl get sensor webhook -o yaml -n argo-events`. However, when an event occurs and a workflow is actually created, it is somehow missing the `metrics` field.

### Installing argo
```
$ tree
.
├── kustomization.yaml
└── overlays
    └── workflow-controller-configmap.yaml

1 directory, 2 files
```

```
$ cat kustomization.yaml 
bases:
  - https://github.com/argoproj/argo/manifests/cluster-install

patchesStrategicMerge:
  - overlays/workflow-controller-configmap.yaml
```

```
$ cat overlays/workflow-controller-configmap.yaml 
apiVersion: v1
data:
  metricsConfig: |
    disableLegacy: true
    enabled: true
    path: /metrics
    port: 9090
kind: ConfigMap
metadata:
  name: workflow-controller-configmap
```

`kustomize build . | kubectl apply -f -`

### Installing argo-events
`kubectl create ns argo-events`

```
$ ls
kustomization.yaml

$ cat kustomization.yaml 
bases:
  - https://github.com/argoproj/argo-events/manifests/cluster-install
``` 

`kustomize build . | kubectl apply -f -`

### Installing A Basic Webhook
I just followed this [link](https://argoproj.github.io/argo-events/quick_start/), except I added a tiny `metrics` section to the sensor:
```
$ curl https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/sensors/webhook.yaml > supplied-sensor.yaml

$ diff supplied-sensor.yaml my-sensor.yaml 
30a31,39
>                 metrics:
>                   prometheus:
>                     - name: duration  # Metric name (will be prepended with "argo_workflows_")
>                       labels:
>                         - key: workflow_type
>                           value: "webhook-hello"
>                       help: "Duration of argo workflow"
>                       gauge:
>                         value: "{{workflow.duration}}"

```

### Then After Triggering a Workflow
`kubectl get workflow $WORKFLOW_NAME -o yaml` # <- No "metrics" field :(

whynowy:
[...]

Issue #733, opened 2020-07-09
Eliminate the code dependency on argo workflow


whynowy:
**Is your feature request related to a problem? Please describe.**
argo-events currently has 2 dependencies on argo-workflow:
1. `argo` binary in sensor Dockerfile
2. `github.com/argoproj/argo` import in source code, this is mainly used in https://github.com/argoproj/argo-events/blob/master/sensors/triggers/argo-workflow/argo-workflow.go.

Every time when argo-workflow releases new features, both of these 2 places need to be upgraded.

Considering CLI won't have changes frequently, frequency of upgrading CLI version could unfollow `argo` release frequency. But the code dependency sometimes makes big differences. For example, if there's a new field in `workflow` spec, if we don't upgrade, the new field will not be recognized by argo-events, thus the submitted workflow will not be the same as user provides.

**Describe the solution you'd like**
Eliminate the code dependency on `argo-workflow`.

Don't convert the workflow spec provided by the user to a `workflow` object.

**Describe alternatives you've considered**

**Additional context**

tmshn:
Let me just clarify: AFAIK, the current implementation offers two options to kick argo workflows from sensors

1. Using `argoWorkflow` in TriggerTemplate: (this is what described in this issue) Render a workflow yaml and exec `argo` cli using `os/exec`
2. Using `k8s` in TriggerTemplate: Create any k8s resource from an "unstructured" object using `k8s.io/client-go/dynamic`

Hence, for those who want to be free from the version skew problem, the latter option should work (while it lacks some features like `generateName`), right?

tumido:
Just throwing an idea in here:

Argo has a [REST API](https://github.com/argoproj/argo/blob/master/api/openapi-spec/swagger.json) which supports all that submit/resubmit/retry... requests. That means it can be used instead of bundling `argo` CLI binary and also argo codebase into here. Both can be eliminated.

whynowy:
> Just throwing an idea in here:
> 
> Argo has a [REST API](https://github.com/argoproj/argo/blob/master/api/openapi-spec/swagger.json) which supports all that submit/resubmit/retry... requests. That means it can be used instead of bundling `argo` CLI binary and also argo codebase into here. Both can be eliminated.

REST API is another option, but it need to give argo-server URL.

PR #758, opened 2020-07-19
feat: remove wf dep

VaibhavPage:


sensors/triggers/argo-workflow/argo-workflow.go
```diff
@@ -126,19 +116,14 @@ func (t *ArgoWorkflowTrigger) Execute(events map[string]*v1alpha1.Event, resourc
 
 	switch op {
 	case v1alpha1.Submit:
-		file, err := ioutil.TempFile("/bin/workflows", workflow.Name)
+		file, err := ioutil.TempFile("/bin/workflows", name)
```
whynowy, line 119: `file, err := ioutil.TempFile("", name)`

This will create the file under the default temp directory, then `/bin/workflows` is not required in Dockerfile.

See https://golang.org/pkg/io/ioutil/#TempFile

terev:
@whynowy this should close https://github.com/argoproj/argo-events/issues/738 and https://github.com/argoproj/argo-events/issues/733 right?

whynowy:
@terev - It should.

terev:
Amazing tysm!

PR #761, opened 2020-07-21
fix(eventbus): Refine auto-reconnection and bugfix

whynowy:
There were some cases auto-reconnection didn't work, this change fixes them.

Stop using `reconnect` feature provided by NATS because sometimes there was inconsistent natsConn and stanConn statuses. Handling reconnection in Sensor and EventSource is good enough.


sensors/cmd/Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN yum -y update && yum -y install ca-certificates openssh openssh-server opens
 # Argo Workflow CLI
 COPY assets/argo-linux-amd64 /usr/local/bin/argo
 
-RUN argo version
+RUN argo version || true
```
whynowy, line 7: Latest `argo` cli has an issue that it throws an error if no `kubeconfig` is provided.

whynowy: It looks like the issue has been fixed in `workflow`, let me create a PR to release a new CLI version and then we update it.

https://github.com/argoproj/argo/issues/3383

whynowy: It was merged 4 days ago, not in any release. 

It only impacts `docker build`, let's use `argo version || true` here for now to workaround the issue, and update it later with new CLI later.

PR #762, opened 2020-07-23
docs: Fix typo in docs and update mockery

jonnylangefeld:
There was a word duplicated in docs. On running codegen I noticed that the mockery command was outdated as well


hack/update-mocks.sh
```diff
@@ -24,6 +24,6 @@ do
   MOCK_DIR=$(echo $i | awk -F: '{print $1}')
   MOCK_NAME=$(echo $i | awk -F: '{print $2}')
   cd ${REPO_ROOT}/${MOCK_DIR}
-  mockery -name ${MOCK_NAME}
+  mockery --name ${MOCK_NAME}
```
whynowy, line 27: Why?

```
mockery
Use -name to specify the name of the interface or -all for all interfaces found
```

jonnylangefeld: The version that gets installed via `brew install vektra/tap/mockery` is `2.1.0`. From that version's help string: 
```
--name string            name or matching regular expression of interface to generate mock for
```

jonnylangefeld: They changed to viper for v2 in [this commit](https://github.com/vektra/mockery/commit/1549900a3e99e3d13c9585005cbe0922b41e5d35)

whynowy: In Golang `flag`, double dash is same as single dash, though golang prefers single.

whynowy: @jonnylangefeld - Thanks for sharing that!

jonnylangefeld: hmm 🤔  I don't think it's the same 😄 
run from within the `eventbus/driver` directory:
```
mockery -name Driver
Error: unknown command "Driver" for "mockery"
Run 'mockery --help' for usage.
```
It works with two dashes.

`make codegen` also fails with the newer version of mockery with only a single dash.

whynowy: That is because you are using new version of mockery (as you said, it changed to use `viper`, where double dash does not equal to single dash any more).

PR #777, opened 2020-07-28
fix: switch slack lib and stop using deprecated APIs. Fixes #726

### LINKED ISSUE #726: The `JoinChannel` api call is fully deprecated for new Slack applications
**Describe the bug**
Slack has deprecated the channels.join api call and suggests to use conversations.join instead. API calls now return with `method_deprecated` for `JoinChannel` when you are using a Slack app created after June 10th (or so).

Slack API links:
https://api.slack.com/methods/channels.join
https://api.slack.com/methods/conversations.join

Code:
https://github.com/argoproj/argo-events/blob/693e2109d33d152175cb38dc63198b9517a2032d/sensors/triggers/slack/slack.go#L115

**To Reproduce**
1) Create a new slack app
2) Try to use it with argo event Slack Triggers. You will get an error when argo events tries to send a message to Slack.

**Expected behavior**
1) argo-events should be able to send a message using both older and newer Slack Apps.

 - OS: Linux (Kubernetes)

**Additional context**

`JoinConversation` is already written in the Slack package: https://github.com/slack-go/slack/blob/c664c83937182df61c6532cff2569ad50ec6b117/conversation.go#L536

JoinCoversation requires a channel id instead of channel name. This is different then the current call which expects a name. For backwards compatibility it would make sense to still expect a name and then lookup the ID and pass it to JoinConversation.

**Note**: `github.com/slack-go/slack` is the new community fork of `github.com/nlopes/slack` and is most likely what should be used going forward.

whynowy:
1. Switch slack lib from `github.com/nlopes/slack` to `github.com/slack-go/slack`;
2. Stop using deprecated slack APIs;
3. Removed `namespace` field from slack trigger spec;
4. Fixed a `labels` related bug introduced by #773;
5. Also updated ClusterRole and Role for `argo-events-sa` in this PR.

Fixes #726


controllers/common/util.go
```diff
@@ -37,13 +37,6 @@ func SetObjectMeta(owner, obj metav1.Object, gvk schema.GroupVersionKind) error
 		obj.SetNamespace(owner.GetNamespace())
 	}
 
-	objLabels := obj.GetLabels()
```
whynowy, line 40: The added `owner-id` is not used by anything, but it changes the passed in object metadata labels (here, the passed in objects are `Sensor`, `EventSource` and `EventBus`), as a result, it causes some unexpected issues, such as the `Service` `labelSelector` generated by `EventSource` doesn't match the POD's labels - This is side effect of PR #773. 

Simply remove this to fix the issue. To avoid issues when upgrading (removing a label from EventBus StatefulSet is not allowed), the `owner-id` label need to be added to the passed in objects as below.

controllers/eventbus/installer/installer.go
```diff
@@ -47,8 +48,9 @@ func getInstaller(eventBus *v1alpha1.EventBus, client client.Client, natsStreami
 
 func getLabels(bus *v1alpha1.EventBus) map[string]string {
 	return map[string]string{
-		"controller":    "eventbus-controller",
-		"eventbus-name": bus.Name,
+		"controller":          "eventbus-controller",
+		"eventbus-name":       bus.Name,
+		common.LabelOwnerName: bus.Name,
```
whynowy, line 53: Add `owner-id` label to make it consistent with before.

controllers/eventsource/controller.go
```diff
@@ -84,6 +84,7 @@ func (r *reconciler) reconcile(ctx context.Context, eventSource *v1alpha1.EventS
 		Labels: map[string]string{
 			"controller":                "eventsource-controller",
 			common.LabelEventSourceName: eventSource.Name,
+			common.LabelOwnerName:       eventSource.Name,
```
whynowy, line 87: Add `owner-id` label to make it consistent with before.

controllers/sensor/controller.go
```diff
@@ -99,6 +99,7 @@ func (r *reconciler) reconcile(ctx context.Context, sensor *v1alpha1.Sensor) err
 		Labels: map[string]string{
 			"controller":           "sensor-controller",
 			common.LabelSensorName: sensor.Name,
+			common.LabelOwnerName:  sensor.Name,
```
whynowy, line 102: Add `owner-id` label to make it consistent with before.

eventsources/common/webhook/webhook.go
```diff
@@ -127,14 +127,14 @@ func activateRoute(router Router, controller *Controller) {
 }
 
 // manageRouteChannels consumes data from route's data channel and stops the processing when the event source is stopped/removed
-func manageRouteChannels(router Router, dispatcher func([]byte) error) {
+func manageRouteChannels(router Router, dispatch func([]byte) error) {
```
whynowy, line 130: Name change.

whynowy:
Thanks @spkane for testing the RC images!

PR #788, opened 2020-07-30
chore: Add labels to issue templates + enable bots

alexec:
@VaibhavPage @whynowy 

This PR makes changes to help with prioritisation:

* Automatically labels new issues.
* Asks users to 👍  to help with prioritisation. 
* Adds configuration (but does not enable) `no-response` bot - a bot that closes issues labelled `more-information-needed` if there is no reply for 14 days.
* Adds configuration (but does not enable) `stable` bot - a bot that labels issues with `wontfix` if there is no reply for 60 days and then closes them after 7 days.

This is copied from Argo Workflows.


whynowy:
@alexec - why are those `actions` deleted?

alexec:
@whynowy. Ops. Deleted the workflows in my first commit, fixed in my second commit, but didn't push it. Pushed now.

PR #816, opened 2020-08-05
fix: retry when connect to eventbus in sensor

exherb:
when lots of sensors got updated, eventbus connect is really unstable.

### DIFF
--- a/sensors/listener.go
+++ b/sensors/listener.go
@@ -33,6 +33,7 @@ import (
 	"github.com/argoproj/argo-events/common/logging"
 	"github.com/argoproj/argo-events/eventbus"
 	eventbusdriver "github.com/argoproj/argo-events/eventbus/driver"
+	"github.com/argoproj/argo-events/eventsources/sources"
 	"github.com/argoproj/argo-events/pkg/apis/sensor/v1alpha1"
 	sensordependencies "github.com/argoproj/argo-events/sensors/dependencies"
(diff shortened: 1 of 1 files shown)


whynowy:
We already have a retry function at https://github.com/argoproj/argo-events/blob/master/eventsources/sources/util.go#L18, can you switch to use that?

exherb:
eventsource will crash and restart, so I think it’s ok to not retry connect when start?

exherb:
ok, I will switch to util retry

exherb:
> We already have a retry function at https://github.com/argoproj/argo-events/blob/master/eventsources/sources/util.go#L18, can you switch to use that?

done

sensors/listener.go
```diff
@@ -102,9 +103,14 @@ func (sensorCtx *SensorContext) ListenEvents(ctx context.Context, stopCh <-chan
...
-			conn, err := ebDriver.Connect()
+			var conn eventbusdriver.Connection
+			err = sources.Connect(&common.DefaultRetry, func() error {
+				var err error
+				conn, err = ebDriver.Connect()
+				return err
+			})
 			if err != nil {
-				logger.Error("failed to connect to event bus", zap.Error(err))
+				logger.Fatal("failed to connect to event bus", zap.Error(err))
```
whynowy, line 113: No need to change to `fatal`, the returned error will be wrapped as a `fatal` error in main.go

PR #817, opened 2020-08-05
feat: Kafka Consumer Group

### LINKED ISSUE #264: Add consumer group to Kafka gateway
It would be nice to have the option to add a consumer group to the Kafka config for clarity and monitoring purposes.

Possibility to put Kafka consumer group in kafka-gateway-configmap yaml:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kafka-gateway-configmap
data:
  foo: |-
    url: kafka.argo-events:9092
    topic: foo
    partition: "0"
    consumer-group: "test-group-1"
```

zachaller:
This adds consumer group support to kafka event source.

Closes #264


eventsources/sources/kafka/start.go
```diff
@@ -166,3 +264,55 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+		if err != nil {
+			consumer.logger.Desugar().Error("failed to marshal the event data, rejecting the event...", zap.Error(err))
+			continue
+		}
+
+		if err = consumer.dispatch(eventBody); err != nil {
+			consumer.logger.Desugar().Error("failed to dispatch kafka event...", zap.Error(err))
+		}
+
+		session.MarkMessage(message, "")
```
whynowy, line 314: Confirm - Is it expected to `MarkMessage` even `consumer.dispatch() fails`?

eventsources/sources/kafka/start.go
```diff
@@ -72,6 +76,100 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+		return err
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			if err := client.Consume(ctx, []string{kafkaEventSource.Topic}, &consumer); err != nil {
+				log.Panicf("Error from consumer: %v", err)
```
whynowy, line 138: We support multiple events configured in one EventSource, `Panicf()` exits everything, it is not expected to do that because the other "good" events will be stopped.

eventsources/sources/kafka/start.go
```diff
@@ -72,6 +76,100 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+	}
+
+	consumer := Consumer{
+		ready:            make(chan bool),
+		dispatch:         dispatch,
+		logger:           log,
+		kafkaEventSource: kafkaEventSource,
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
```
whynowy, line 124: The passed in `ctx` is already `WithCancel`.

eventsources/sources/kafka/start.go
```diff
@@ -72,6 +76,100 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			if err := client.Consume(ctx, []string{kafkaEventSource.Topic}, &consumer); err != nil {
+				log.Panicf("Error from consumer: %v", err)
+			}
+			if ctx.Err() != nil {
+				return
```
whynowy, line 141: Returning here means `<-consumer.ready` wait forever.

eventsources/sources/kafka/start.go
```diff
@@ -72,6 +76,100 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+
+	<-consumer.ready // Await till the consumer has been set up
+	log.Info("Sarama consumer group up and running!...")
+
+	sigterm := make(chan os.Signal, 1)
+	signal.Notify(sigterm, syscall.SIGINT, syscall.SIGTERM)
+	select {
+	case <-ctx.Done():
+		log.Info("terminating: context cancelled")
+	case <-sigterm:
```
whynowy, line 155: This is not needed as the passed in `ctx.Done()` is used to take care of it.

eventsources/sources/kafka/start.go
```diff
@@ -72,6 +76,100 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+		cancel()
+		return err
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			if err := client.Consume(ctx, []string{kafkaEventSource.Topic}, &consumer); err != nil {
```
whynowy, line 137: Is this a blocking call? for loop here means it keeps calling it, or do I miss anything?

eventsources/sources/kafka/start.go
zachaller: Good to know

eventsources/sources/kafka/start.go
zachaller: Good catch fixed

eventsources/sources/kafka/start.go
zachaller: I still need the cancel function in the code here

```WithCancel returns a copy of parent with a new Done channel. The returned context's Done channel is closed when the returned cancel function is called or when the parent context's Done channel is closed, whichever happens first.```

File: eventsources/sources/kafka/start.go (hunk 6 above)
Comment (zachaller) [reply eventsources/sources/kafka/start.go:136]: This function dose block.

```
Consume joins a cluster of consumers for a given list of topics and starts a blocking ConsumerGroupSession through the ConsumerGroupHandler.
The life-cycle of a session is represented by the following steps:
1. The consumers join the group (as explained in https://kafka.apache.org/documentation/#intro_consumers )
   and is assigned their "fair share" of partitions, aka 'claims'.
2. Before processing starts, the handler's Setup() hook is called to notify the user
   of the claims and allow any necessary preparation or alteration of state.
3. For each of the assigned claims the handler's ConsumeClaim() function is then called
   in a separate goroutine which requires it to be thread-safe. Any state must be carefully protected
   from concurrent reads/writes.
4. The session will persist until one of the ConsumeClaim() functions exits. This can be either when the
   parent context is cancelled or when a server-side rebalance cycle is initiated.
5. Once all the ConsumeClaim() loops have exited, the handler's Cleanup() hook is called
   to allow the user to perform any final tasks before a rebalance.
6. Finally, marked offsets are committed one last time before claims are released.
Please note, that once a rebalance is triggered, sessions must be completed within Config.Consumer.Group.Rebalance.Timeout. This means that ConsumeClaim() functions must exit as quickly as possible to allow time for Cleanup() and the final offset commit. If the timeout is exceeded, the consumer will be removed from the group by Kafka, which will cause offset commit failures.
```

File: eventsources/sources/kafka/start.go (hunk 4 above)
Comment (zachaller) [reply eventsources/sources/kafka/start.go:143]: not sure exactly what the correct thing to do here is, I am taking a most of the code and ideas from https://github.com/Shopify/sarama/blob/master/examples/consumergroup/main.go#L110

Comment (zachaller) [reply eventsources/sources/kafka/start.go]: instead of return, I now have it calling client.Close()

File: eventsources/sources/kafka/start.go [hunk 7]
```diff
@@ -72,6 +73,99 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+			consumer.ready = make(chan bool)
+		}
+	}()
+
+	<-consumer.ready // Await till the consumer has been set up
+	log.Info("Sarama consumer group up and running!...")
+
+	<-ctx.Done()
+	log.Info("terminating: context cancelled")
+	cancel()
```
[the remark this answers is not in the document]
whynowy: I understand you are referencing the example code from https://github.com/Shopify/sarama/blob/master/examples/consumergroup/main.go#L121-L127, but that doesn't seem very correct, the following part is not need for example at all, and the case here is a little bit different.

```
         case <-ctx.Done():
		log.Println("terminating: context cancelled")
```

whynowy:
I think the following code is good enough - bear me I don't have a kafka testing env, didn't test it.

```
	client, err := sarama.NewConsumerGroup([]string{kafkaEventSource.URL}, kafkaEventSource.ConsumerGroup.GroupName, config)
	if err != nil {
		log.Errorf("Error creating consumer group client: %v", err)
		return err
	}

	wg := &sync.WaitGroup{}
	wg.Add(1)
	go func() {
		defer wg.Done()
		for {
			// `Consume` should be called inside an infinite loop, when a
			// server-side rebalance happens, the consumer session will need to be
			// recreated to get the new claims
			if err := client.Consume(ctx, []string{kafkaEventSource.Topic}, &consumer); err != nil {
				log.Errorf("Error from consumer: %v", err)
			}
			// check if context was cancelled, signaling that the consumer should stop
			if ctx.Err() != nil {
				log.Errorf("Error from context: %v", ctx.Err())
				return
			}
			consumer.ready = make(chan bool)
		}
	}()

	<-consumer.ready // Await till the consumer has been set up
	log.Info("Sarama consumer group up and running!...")

	<-ctx.Done()
	log.Info("terminating: context cancelled")
	wg.Wait()
```

zachaller:
Ok made requested changes, and it is still working in my testing environment as well both partition client and consumer groups.

eventsources/sources/kafka/start.go
whynowy: Take back my previous comment "Returning here means <-consumer.ready wait forever."

eventsources/sources/kafka/start.go
```diff
@@ -72,6 +73,97 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+		for {
+			// `Consume` should be called inside an infinite loop, when a
+			// server-side rebalance happens, the consumer session will need to be
+			// recreated to get the new claims
+			if err := client.Consume(ctx, []string{kafkaEventSource.Topic}, &consumer); err != nil {
+				log.Errorf("Error from consumer: %v", err)
+			}
+			// check if context was cancelled, signaling that the consumer should stop
+			if ctx.Err() != nil {
+				log.Errorf("Error from context: %v", ctx.Err())
```
whynowy, line 140: It might be better to be a `Info("xxxx existing xxxxx")`, it should happen when a stop signal is triggered (killing the POD).

whynowy:
Also, could you please update the kafka example?

zachaller:
Yes I will update the example, also the logs are not exactly what i expected when deleting the pod I can try to play around with it some if your saying it should be able to cleanup before exiting here is what I currently am getting both with and without the return statment

```
{"level":"info","ts":1596681498.1279385,"logger":"argo-events.eventsource","caller":"eventsources/eventing.go:327","msg":"Eventing server started.","eventSourceName":"kafka"}
{"level":"info","ts":1596681498.1281083,"logger":"argo-events.eventsource","caller":"eventsources/eventing.go:268","msg":"starting eventbus connection daemon...","eventSourceName":"kafka"}
{"level":"info","ts":1596681498.1284497,"logger":"argo-events.eventsource","caller":"kafka/start.go:73","msg":"start kafka event source...","eventSourceName":"kafka","eventSourceType":"kafka","eventName":"example"}
{"level":"info","ts":1596681503.176573,"logger":"argo-events.eventsource","caller":"kafka/start.go:148","msg":"Sarama consumer group up and running!...","eventSourceName":"kafka","eventSourceType":"kafka","eventName":"example"}
{"level":"info","ts":1596681809.267002,"logger":"argo-events.eventsource","caller":"eventsources/eventing.go:329","msg":"Shutting down...","eventSourceName":"kafka"}
```

whynowy:
> {"level":"info","ts":1596681809.267002,"logger":"argo-events.eventsource","caller":"eventsources/eventing.go:329","msg":"Shutting down...","eventSourceName":"kafka"}

This is a known issue in eventing.go, it doesn't give enough time after calling `cancel()`. For now, I think your code is ok.

zachaller:
Last question just to confirm then do we want this return or should i just remove it.

```
// check if context was cancelled, signaling that the consumer should stop
			if ctx.Err() != nil {
				log.Infof("Error from context: %v", ctx.Err())
				return
			}
```

zachaller:
> Last question just to confirm then do we want this return or should i just remove it.
> 
> ```
> // check if context was cancelled, signaling that the consumer should stop
> 			if ctx.Err() != nil {
> 				log.Infof("Error from context: %v", ctx.Err())
> 				return
> 			}
> ```

The logic there I think that <-consumer.ready would generally already be passed because it is closed in the Setup pretty much right away. Then the ```consumer.ready = make(chan bool)``` just re inits the ready channel for the next for loop in the cases of a server rebalance.

whynowy:
> Last question just to confirm then do we want this return or should i just remove it.
> 
> ```
> // check if context was cancelled, signaling that the consumer should stop
> 			if ctx.Err() != nil {
> 				log.Infof("Error from context: %v", ctx.Err())
> 				return
> 			}
> ```

That `return` is needed, that's the only exit of the goroutine.

whynowy:
> > Last question just to confirm then do we want this return or should i just remove it.
> > ```
> > // check if context was cancelled, signaling that the consumer should stop
> > 			if ctx.Err() != nil {
> > 				log.Infof("Error from context: %v", ctx.Err())
> > 				return
> > 			}
> > ```
> 
> The logic there I think that <-consumer.ready would generally already be passed because it is closed in the Setup pretty much right away. Then the `consumer.ready = make(chan bool)` just re inits the ready channel for the next for loop in the cases of a server rebalance.

Correct. I misunderstood that at the beginning, that's why I took back the previous comment. The `return` is needed.

zachaller:
Awesome ok, I think all changes are made let me know if there is anything else needed, and thanks for the help and thorough review

zachaller:
Also one last note, I think with supporting consumer groups you could in theory now scale the EventSource but only if consumer groups are used I noticed the docs here have it as a recreate deployment https://github.com/argoproj/argo-events/blob/master/docs/eventsource-deployment-strategies.md It's probably fine to leave it that way as well not sure scaling kafka consumer in that way buys a whole lot when most of the work happens in the sensor but anyways thought I would mention it at least.

whynowy:
> Also one last note, I think with supporting consumer groups you could in theory now scale the kafka EventSource but only if consumer groups are used I noticed the docs here have it as a recreate deployment https://github.com/argoproj/argo-events/blob/master/docs/eventsource-deployment-strategies.md It's probably fine to leave it that way as well not sure scaling kafka consumer in that way buys a whole lot when most of the work happens in the sensor but anyways thought I would mention it at least.

That will add lots of if..else, let's leave it as is.

whynowy:
> > {"level":"info","ts":1596681809.267002,"logger":"argo-events.eventsource","caller":"eventsources/eventing.go:329","msg":"Shutting down...","eventSourceName":"kafka"}
> 
> This is a known issue in `Start()` in eventing.go, it doesn't give enough time after calling `cancel()` or use a callback to indicate the clean up is done.. For now, I think your code is ok.

Following PR should be able to fix the issue.

https://github.com/argoproj/argo-events/pull/819

PR #818, opened 2020-08-06
docs: refine docs for service accounts and managed namespace

whynowy:


VaibhavPage:
Let's keep the issue open so we can incrementally enhance docs till 1.0

PR #829, opened 2020-08-11
feat: Rate limit for Kafka and initial offset support and refactor config

zachaller:
This adds a simple rate limit to kafka event creation. This also refactor and cleans up a bit of the code to setup kafka config for sarama. I also moved the kafka version variable out of consumer groups it actually needs to be set in order to pull  ```msg.Timestamp.String()``` Can see issue [here](https://github.com/Shopify/sarama/issues/750#issuecomment-250920441)


pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -316,13 +316,21 @@ type KafkaEventSource struct {
...
 	// +optional
 	ConsumerGroup *KafkaConsumerGroup `json:"consumerGroup,omitempty" protobuf:"bytes,8,opt,name=consumerGroup"`
+
+	// Sets a limit on how many events get read from kafka per second which in turn limits the number of pods
+	// created per second if using an argo workflow or k8s sensor.
+	// +optional
+	LimitEventsPerSecond int64 `json:"limitEventsPerSecond,omitempty" protobuf:"varint,9,opt,name=limitEventsPerSecond"`
+
+	// Specify what kafka version is being connected to enables certain features in sarama, defaults to 1.0.0
+	KafkaVersion string `json:"kafkaVersion" protobuf:"bytes,10,opt,name=kafkaVersion"`
```
whynowy, line 326: `Version` might be better since it's already under `kafka`.

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -316,13 +316,21 @@ type KafkaEventSource struct {
 	// Consumer group for kafka client
 	// +optional
 	ConsumerGroup *KafkaConsumerGroup `json:"consumerGroup,omitempty" protobuf:"bytes,8,opt,name=consumerGroup"`
+
+	// Sets a limit on how many events get read from kafka per second which in turn limits the number of pods
```
whynowy, line 320: Less comments here - feel like removing `which in turn limits .....` is better.

eventsources/sources/kafka/start.go
```diff
@@ -81,14 +82,10 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 }
 
 func (listener *EventListener) consumerGroupConsumer(ctx context.Context, log *zap.SugaredLogger, kafkaEventSource *v1alpha1.KafkaEventSource, dispatch func([]byte) error) error {
-	config := sarama.NewConfig()
-
-	version, err := sarama.ParseKafkaVersion(kafkaEventSource.ConsumerGroup.KafkaVersion)
+	config, err := getSaramaConfig(log, kafkaEventSource)
```
whynowy, line 85: Move `log` to the second.

eventsources/sources/kafka/start.go
```diff
@@ -310,7 +321,20 @@ func (consumer *Consumer) ConsumeClaim(session sarama.ConsumerGroupSession, clai
 		} else {
 			session.MarkMessage(message, "")
 		}
+		if consumer.kafkaEventSource.LimitEventsPerSecond > 0 {
+			//1000000000 is 1 second in nanoseconds
+			d := (1000000000 / time.Duration(consumer.kafkaEventSource.LimitEventsPerSecond) * time.Nanosecond) * time.Nanosecond
+			consumer.logger.Infof("Sleeping for: %v.", d)
+			sleepContext(consumer.ctx, d)
```
whynowy, line 328: Shouldn't this simply be `time.Sleep(d)`?

zachaller: So the reason behind this is to just be kind to cleanup code this a comment from sarama ```Please note, that once a rebalance is triggered, sessions must be completed within Config.Consumer.Group.Rebalance.Timeout. This means that ConsumeClaim() functions must exit as quickly as possible to allow time for Cleanup() and the final offset commit. If the timeout is exceeded, the consumer will be removed from the group by Kafka, which will cause offset commit failures.``` By doing a sleep this way if the ctx is canceled the sleep will also get canceled.

whynowy: I don't see any relationship between them, or I missed anything? `ctx` only gets canceled when the event source service gets a SIGTERM, however the rebalance is something happens on the kafka server side, it will not lead to the a ctx cancel.

zachaller: So it is pretty minor I suppose on an exit from SIGTERM I thought it would still be nice to give the cleanup as much time as possible granted it currently does nothing so probably moot. We also only go down to a min of 1 rps so that would be a max sleep of 100ms which really is not that long. I can move it to just a sleep.

eventsources/sources/kafka/start.go
```diff
@@ -259,12 +238,44 @@ func (el *EventListener) partitionConsumer(ctx context.Context, log *zap.Sugared
...
+	return config, nil
+}
+
 // Consumer represents a Sarama consumer group consumer
 type Consumer struct {
 	ready            chan bool
 	dispatch         func([]byte) error
 	logger           *zap.SugaredLogger
 	kafkaEventSource *v1alpha1.KafkaEventSource
+	ctx              context.Context
```
whynowy, line 278: Then we don't need `ctx` here.

PR #838, opened 2020-08-17
chore: Sensor base image change

whynowy:
Updated sensor base image to `alpine` to gain more security.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,9 @@ ENTRYPOINT [ "/bin/eventsource" ]
 # sensor
 ####################################################################################################
 FROM alpine as sensor
+RUN apk update && apk upgrade && \
+    apk add --no-cache git
+
 COPY --from=base /usr/share/zoneinfo /usr/share/zoneinfo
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -44,8 +44,9 @@ ENTRYPOINT [ "/bin/eventsource" ]
 ####################################################################################################
 # sensor
 ####################################################################################################
-FROM centos:8 as sensor
-RUN yum -y update && yum -y install ca-certificates openssh openssh-server openssh-clients openssl-libs curl git
+FROM alpine as sensor
+COPY --from=base /usr/share/zoneinfo /usr/share/zoneinfo
+COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
VaibhavPage, line 49: If I remember correctly, we need the git and other utilities for sensor to pull workflow definitions from git.

whynowy: ahhh, I totally missed that..... is `curl` required?

whynowy: Update. Verified pulling from git works with both `git://` and `https://`.

PR #858, opened 2020-09-02
chore: code refactory and cleaning up unused code

whynowy:
Code rafactory and a bugfix for Makefile - order matters for `eventsource` and `eventsource-controller`, `sensor` and `sensor-controller`.

### DIFF
--- a/common/retry.go
+++ b/common/retry.go
@@ -65,3 +65,17 @@ func GetConnectionBackoff(backoff *apicommon.Backoff) *wait.Backoff {
 	}
 	return &result
 }
+
+// Connect is a general connection helper
+func Connect(backoff *wait.Backoff, conn func() error) error {
+	if backoff == nil {
+		backoff = &DefaultRetry
+	}
+	err := wait.ExponentialBackoff(*backoff, func() (bool, error) {
+		if err := conn(); err != nil {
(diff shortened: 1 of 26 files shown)


common/retry.go
```diff
@@ -65,3 +65,17 @@ func GetConnectionBackoff(backoff *apicommon.Backoff) *wait.Backoff {
 	}
 	return &result
 }
+
+// Connect is a general connection helper
+func Connect(backoff *wait.Backoff, conn func() error) error {
```
whynowy, line 70: Moved from `eventsources/util` to `common`, as sensor service also uses it.

eventbus/driver/nats.go
```diff
@@ -147,7 +149,7 @@ func (n *natsStreaming) SubscribeEventSources(ctx context.Context, conn Connecti
 	}, stan.DurableName(durableName),
 		stan.SetManualAckMode(),
 		stan.StartAt(pb.StartPosition_LastReceived),
-		stan.AckWait(1*time.Second),
+		stan.AckWait(3*time.Second),
```
whynowy, line 152: Slightly increase a little bit, though it does not matter much.

VaibhavPage (approved):
lgtm!

PR #868, opened 2020-09-06
feat: pass eventbus cr labels to resources

VaibhavPage:


whynowy (commented):
Overall, I think allowing customizing labels will bring lots of uncertainty. 

The reason we allow `EventSource` and `Sensor` to change POD metadata (**only POD metadata**) is because there are lots of requirements that POD `annotations` should be customized to allow things like [kiam](https://github.com/uswitch/kiam) to work, that's a very popular service on AWS platforms. And the way to customize it is through a dedicated field `spec.template.metadata`, not inheriting from CR objects. However, EventBus does not have the requirement at all.

In fact, even for EventSource and Sensor, we should **ONLY** allow the customization on POD `annotations`, not `labels`, because those labels are also populated to `spec.selector.matchLabels` of Deployments, and `matchLabels` became immutable starting from `apps/v1beta2`, which is now in `apps/v1`, see https://github.com/kubernetes/kubernetes/issues/50808. Changing labels on an existing object will fail the reconciliation - we should fix this.

VaibhavPage:
> Overall, I think allowing customizing labels will bring lots of uncertainty.
> 
> The reason we allow `EventSource` and `Sensor` to change POD metadata (**only POD metadata**) is because there are lots of requirements that POD `annotations` should be customized to allow things like [kiam](https://github.com/uswitch/kiam) to work, that's a very popular service on AWS platforms. And the way to customize it is through a dedicated field `spec.template.metadata`, not inheriting from CR objects. However, EventBus does not have the requirement at all.
> 
> In fact, even for EventSource and Sensor, we should **ONLY** allow the customization on POD `annotations`, not `labels`, because those labels are also populated to `spec.selector.matchLabels` of Deployments, and `matchLabels` became immutable starting from `apps/v1beta2`, which is now in `apps/v1`, see [kubernetes/kubernetes#50808](https://github.com/kubernetes/kubernetes/issues/50808). Changing labels on an existing object will fail the reconciliation - we should fix this.

That's fine. We will only set the labels for the first time and any subsequent updates to eventbus CR labels will be a no-op. Not providing the ability to configure labels to the actual resources severely limits the ability to use Argo Events in enterprises who run Admission controllers and expect every object within K8s has certain labels.

whynowy:
> > Overall, I think allowing customizing labels will bring lots of uncertainty.
> > The reason we allow `EventSource` and `Sensor` to change POD metadata (**only POD metadata**) is because there are lots of requirements that POD `annotations` should be customized to allow things like [kiam](https://github.com/uswitch/kiam) to work, that's a very popular service on AWS platforms. And the way to customize it is through a dedicated field `spec.template.metadata`, not inheriting from CR objects. However, EventBus does not have the requirement at all.
> > In fact, even for EventSource and Sensor, we should **ONLY** allow the customization on POD `annotations`, not `labels`, because those labels are also populated to `spec.selector.matchLabels` of Deployments, and `matchLabels` became immutable starting from `apps/v1beta2`, which is now in `apps/v1`, see [kubernetes/kubernetes#50808](https://github.com/kubernetes/kubernetes/issues/50808). Changing labels on an existing object will fail the reconciliation - we should fix this.
> 
> That's fine. We will only set the labels for the first time and any subsequent updates to eventbus CR labels will be a no-op. Not providing the ability to configure labels to the actual resources severely limits the ability to use Argo Events in enterprises who run Admission controllers and expect every object within K8s has certain labels. I am not sure how this brings uncertainty at all.

Subsequent no-op means reconciliation on same object might lead to different behavior - I think this should be avoided.

If we do want to apply labels from CR to underlying objects, I think what we can do is only applying to StatusfulSets, Deployments, Service, etc, but not PODs, which means not impacting `matchedLabels` and `templateLabels`,

whynowy:
As discussed, close this one.

Issue #836, opened 2020-08-14
Official ARM docker images 


dgiebert:
# Summary

As in the main project it would be great if there are mulit-arch images for this piece of the argo project.

# Motivation

Currently building a Raspberry Pi Cluster with a working argo installation, and want to extend it with event driven workflows.

# Proposed solution

Rework the Makefile in sync with argo: https://github.com/argoproj/argo/pull/2491
Rework the Github Workflow in sync with argo: https://github.com/argoproj/argo/pull/3744

If this is desired I could take care and create a PR for it ?

whynowy:
@dgiebert - sure, welcome for the contribution!

flavio:
I was looking for ARM64 support and I found this issue and its PR. It's great to see it has been merged recently :partying_face: 

Is the fix going to be included inside of the next release of argo-events? Currently `v1.0.0-rc3` is the latest release, its tagging happened 7 days before the PR got merged, hence there are no ARM64 container images available yet.
Are they going to be available with `v1.0.0-rc4` or with the final `v1.0.0`? :pray:

whynowy:
@flavio - We just released v1.0.0, unfortunately `arm64` is not there - I tried to do it, but failed, there are still some problems with the new experimental `docker buildx` feature. I hope we can get it addressed in next release.

flavio:
@whynowy there's no reason to feel sorry :)

Is there anything that I could do to help, are these issues tracked somewhere?

dgiebert:
Found the issue and will create a PR soon.

For clean repos `rm dist/*` does not work as there are no files in the dist folder, will therefore force it to ignore this error

PR #869, opened 2020-09-07
ci: Fix empty dist folder and wrong copy

dgiebert:
Fixes the broken build in the CI and was introduced here: https://github.com/argoproj/argo-events/pull/843

Also relates to: https://github.com/argoproj/argo-events/issues/836

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,8 +38,8 @@ COPY . .
 # check we can use Git
 RUN git rev-parse HEAD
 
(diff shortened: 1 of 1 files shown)


whynowy:
See https://github.com/argoproj/argo-events/issues/870.

alexec:
Is this PR still needed?

whynowy:
We only release amd64 image for now, will revisit this in the future. Thanks for your contribution, we learned a lot!

PR #876, opened 2020-09-11
feat: Pass CRD metadata to generated  resources

whynowy:
1. Pass CRD labels in metadata to generated resources like StatefulSet, Deployment and Service.
2. Added `metadata` under `native` for EventBus, to allow POD Labels and Annotations customization for native EventBus.
3. Updated EventSource and Sensor POD selector to use fixed internal labels, instead of customized user input.


akonale (commented):
I tried to build this version and test. There seems to be some bug, when eventsource deployment gets created, it goes to some kind of loop where it tries to continuously update the deployment. This causes corresponding pods and services to get recreated continuously. I have tested with master to make sure it doesn't happen.

whynowy:
> I tried to build this version and test. There seems to be some bug, when eventsource deployment gets created, it goes to some kind of loop where it tries to continuously update the deployment. This causes corresponding pods and services to get recreated continuously. I have tested with master to make sure it doesn't happen.

Thanks! Fixed.

akonale:
Hi,
Is it possible to extend labels to statefulsets created by nats as well ?

whynowy:
> I tried to build this version and test. There seems to be some bug, when eventsource deployment gets created, it goes to some kind of loop where it tries to continuously update the deployment. This causes corresponding pods and services to get recreated continuously. I have tested with master to make sure it doesn't happen.

We will not extend it to POD spec, and for Sensor and EventSource, we will refrain a little bit on the POD metadata customization. The reason we allow POD metadata customization is because there are requirements to annotate the POD when using `kiam` or `k2iam`. However customized labels on POD will pollute the immutable `matchLabels` POD selector.

whynowy:
> > I tried to build this version and test. There seems to be some bug, when eventsource deployment gets created, it goes to some kind of loop where it tries to continuously update the deployment. This causes corresponding pods and services to get recreated continuously. I have tested with master to make sure it doesn't happen.
> 
> We will not extend it to POD spec, and for Sensor and EventSource, we will refrain a little bit on the POD metadata customization. The reason we allow POD metadata customization is because there are requirements to annotate the POD when using `kiam` or `k2iam`. However customized labels on POD will pollute the immutable `matchLabels` POD selector.

Another reason we don't want to extend it to StatefulSet POD is, everything in StatefulSet spec except replica is immutable, we don't want to see unexpected behaviors when a metadata changes.

akonale:
Yup. This definitely makes sense. But instead of customising labels of each and every resources, isn't it possible to add set of labels at top level ? eg. here. [link](https://github.com/argoproj/argo-events/blob/master/controllers/eventsource/controller.go#L84) .  I saw these set of static labels will be there in all the resources being created, and one of the use of it to see all the resources created by this CRD. If user supplied eventbus labels are merged there then there is no need to pick and customise each resources being created right ? Do you think merging user defined labels here would cause the problems you described ?

whynowy:
> Yup. This definitely makes sense. But instead of customising labels of each and every resources, isn't it possible to give user ability to define labels for all resources created at top level ? eg. here. https://github.com/argoproj/argo-events/blob/master/controllers/eventsource/controller.go#L84 . I saw these set of static labels will be there in all the resources being created(including statefulsets and subsequent pods), and maybe one of the use of it to see all the resources created by this CRD. If user supplied eventbus labels are merged there then there is no need to pick and customise each resources being created right ? Do you think adding user defined labels to these set of static labels would cause the problems you described ?

Yes, customized labels will cause problems. Think about what will happen when updating the customized labels.

akonale:
I don't know much about kiam, but updating templates in statefulsets are allowed. This shouldn't cause any problems. Updating fields other than `fields other than 'replicas', 'template', and 'updateStrategy'` are forbidden.

whynowy:
> I don't know much about kiam, but updating templates in statefulsets are allowed. This shouldn't cause any problems. Updating fields other than `fields other than 'replicas', 'template', and 'updateStrategy'` are forbidden. I can definitely help testing some of the scenarios you mentioned which could cause problems by adding few extra labels to the underlying resources being created.

I think you'd better do a real testing to see if it is possible to update `spec.selector.matchLabels` and `spec.template.metadata.labels` for StatefulSet.

whynowy:
@akonale - `metadata` added to `native` for EventBus.

VaibhavPage (approved):

PR #879, opened 2020-09-12
chore: Update swagger file to include unused definitions.

whynowy:
Definitions like `EventSourceList` are necessary for clients generation such as https://github.com/argoproj-labs/argo-events-client-java.


VaibhavPage (approved):
lgtm!

PR #895, opened 2020-09-26
feat: generic event source

VaibhavPage:
Checklist:

* [ ] Either (a) I've created an [feature request](https://github.com/argoproj/argo-events/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [ ] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [ ] I've signed the CLA.
* [ ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ ] My builds are green. Try syncing with master if they are not. 


eventsources/eventing.go
```diff
@@ -223,6 +224,13 @@ func GetEventingServers(eventSource *v1alpha1.EventSource) map[apicommon.EventSo
 		}
 		result[apicommon.PulsarEvent] = servers
 	}
+	if len(eventSource.Spec.Generic) != 0 {
```
whynowy, line 227: Is `Generic` `Recreat` or `RollingUpdate` strategy? If it's `recreate`, need to add it in the slice in pkg/apis/common/common.go

eventsources/sources/generic/start.go
```diff
@@ -0,0 +1,114 @@
...
+			eventBytes, err := json.Marshal(eventData)
+			if err != nil {
+				logger.Errorw("failed to marshal the event data", zap.Error(err))
+				continue
+			}
+
+			logger.Info("dispatching event...")
+			if err := dispatch(eventBytes); err != nil {
+				logger.Errorw("failed to dispatch the event", zap.Error(err))
+				continue
```
whynowy, line 110: It looks like `continue` is not needed.

eventsources/sources/generic/start.go
```diff
@@ -0,0 +1,114 @@
...
+			logger.Info("event source is stopped")
+			clientConn.Close()
+			return nil
+
+		default:
+			event, err := eventStream.Recv()
+			if err != nil {
+				if err == io.EOF {
+					logger.Info("event source is stopped by the server")
+					return nil
```
whynowy, line 84: `return nil` means this event server GoRoutine (https://github.com/argoproj/argo-events/blob/master/eventsources/eventing.go#L305-L340) will be ended, so the POD will still be Running status but do nothing but wait for SigTerm (https://github.com/argoproj/argo-events/blob/master/eventsources/eventing.go#L342).

I feel like we need to have something here (i.e. a daemon GoRoutine) to auto-reconnect, otherwise the server restart will lead to that situation.

eventsources/sources/generic/start.go
```diff
@@ -0,0 +1,114 @@
...
+				}
+
+				clientConn.Close()
```
whynowy, line 87: Usually what will be the error here? For example, could it be a temporary network issue caused error? I ask this because returning the error here means this event server GoRoutine will be ended, it will not restart unless the POD restarts.

examples/event-sources/generic.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: generic
+spec:
+  generic:
+    example:
+      insecure: true
+      # URL of the gRPC server that implements event source
+      url: "generic-event-source-server.argo-events.svc:8080"
+      # Config represents the configuration required to run the event source
+      config: |-
```
whynowy, line 12: Not quite sure how to use `config`, expecting a document to explain how to use `GenericEventSource`.

eventsources/sources/generic/start.go
whynowy: Actually similar issue exists in most of the event sources (i.e. somebody updated SQS permission config on AWS console, and it makes SQS event source stop running). This issue exists because we allow multiple events configured in one event source, and if one of the events stopped running, we need to make sure the others still run well.

A better idea is handling the auto re-connection (auto re-starting) in https://github.com/argoproj/argo-events/blob/master/eventsources/eventing.go, but I'm not sure if that makes lots of senses, because different event sources might have different cases. Before we have (or we can have) that, implementing auto-reconnection here is necessary, as it's very common for the user to restart the GRPC server (or redeploy).

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -120,7 +120,7 @@ type Template struct {
 	// Container is the main container image to run in the event source pod
 	// +optional
 	Container *corev1.Container `json:"container,omitempty" protobuf:"bytes,3,opt,name=container"`
-	// Volumes is a list of volumes that can be mounted by containers in a workflow.
+	// Volumes is a list of volumes that can be mounted by containers in a eventsource.
```
magaldima, line 123: in "an" eventsource

eventsources/sources/generic/start.go
```diff
@@ -0,0 +1,114 @@
...
+		return errors.Wrapf(err, "failed to start the event source stream for %s", el.GetEventName())
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("event source is stopped")
+			clientConn.Close()
+			return nil
+
```
magaldima, line 78: why the extra spaces?

eventsources/eventing.go
VaibhavPage: Done.

examples/event-sources/generic.yaml
VaibhavPage: Added docs on generic es

eventsources/sources/generic/start.go
VaibhavPage: done

eventsources/eventing.go
VaibhavPage: done

eventsources/sources/generic/start.go
VaibhavPage: done

examples/event-sources/generic.yaml
VaibhavPage: added docs

eventsources/sources/generic/start.go
VaibhavPage: done

pkg/apis/eventsource/v1alpha1/types.go
VaibhavPage: done. good to see a comment from you :)

eventsources/sources/generic/start.go
```diff
@@ -0,0 +1,113 @@
...
+	logger.Info("started processing the generic event source...")
+	defer sources.Recover(el.GetEventName())
+
+	logger.Info("connecting to eventsource server in 5 seconds...")
+	ticker := time.NewTicker(5 * time.Second)
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("closing client connection and exiting eventsource...")
+			el.conn.Close()
```
whynowy, line 56: Minor - need to check if `el.conn` is not nil and connected.

whynowy (commented):
LGTM expect 2 minor stuffs. We might need to figure out a way for auth later.

eventsources/sources/minio/start.go
```diff
@@ -82,24 +82,24 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 
 	doneCh := make(chan struct{})
 
-	logger := log.With("bucketName", minioEventSource.Bucket.Name).Desugar()
+	logger := log.With("bucketName", minioEventSource.Bucket.Name)
```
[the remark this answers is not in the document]
VaibhavPage: done. shouldve used the same `log`, dont know why i didnt see it :)

eventsources/sources/generic/start.go
VaibhavPage: good catch.

VaibhavPage:
> LGTM expect 2 minor stuffs. We might need to figure out a way for auth later.

Yes, lets address that in a separate PR.

whynowy (approved):
LGTM

PR #915, opened 2020-10-15
feat: extend resource eventsource field filter. Closes #913

### LINKED ISSUE #913: Field selector regex support in Resource EventSource
Currently the field selector in resource event source is delegated to k8s field selector, which does not support something like `field=abc*`.

Implement our own selector to support regex.

**Additional context**

---

whynowy:
This PR changed the implementation of field filter of `resource` EventSource.

Field field used to utilize [K8s field selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/), but it:

1. Only limited fields (such as `metadata.name`, `status.phase`) can be used for filtering, you can not do something like 

```
kubectl get ingress --field-selector foo.bar=baz
```
2. The value needs to be a fixed string, no regex support.

The PR makes above requirements supported. 

In term of operators, same as K8s field selector, only `=`, `==` and `!=` are supported.

Checklist:

* [x] Either (a) I've created an [feature request](https://github.com/argoproj/argo-events/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed the CLA.
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] My builds are green. Try syncing with master if they are not. 

 Closes #913


eventsources/sources/resource/start.go
```diff
@@ -272,17 +247,56 @@ func passFilters(event *InformerEvent, filter *v1alpha1.ResourceFilter, startTim
 	}
 	uObj := event.Obj.(*unstructured.Unstructured)
 	if len(filter.Prefix) > 0 && !strings.HasPrefix(uObj.GetName(), filter.Prefix) {
-		log.Infof("resource name does not match prefix. resource-name: %s, prefix: %s\n", uObj.GetName(), filter.Prefix)
+		log.Debugf("resource name does not match prefix. resource-name: %s, prefix: %s\n", uObj.GetName(), filter.Prefix)
```
VaibhavPage, line 250: Don't we want to log messages as to why event failed?

whynowy: Changed back to Infof.

VaibhavPage (approved):
LGTM

PR #927, opened 2020-10-24
feat: Retry starting event server and quick fail. Closes #926

### LINKED ISSUE #926: Event server retry and quick fail
EventSource allow multiple events defined in one objects, which mean there are multiple eventing servers running in one POD, each of them runs in a GoRoutine.

In this case, if any of the eventing servers fails due to an error, the GoRoutine will quit and there's no way to get it back unless restart the POD.

See a related issue #898

There's not a way to fix this issue thoroughly because we allow multiple events running in same POD, and we don't want the whole POD panic because of a failure from one of the events. The way we can mitigate it is:

1. Add retry in eventing.go, which means if there's failure to start the eventing server (or there's something wrong during the eventing server runs), we do several retries.

2. Panic the POD if there's no active eventing server runs, and then k8s will take over the restart/retry, this makes it better for the case that only one event is configured in the EventSource object.

---

whynowy:
1. Retry initial EventBus connection;
2. Retry starting eventing servers;
3. Panic EventSource POD if there's no active eventing servers(failed after retry)

Checklist:

* [x] Either (a) I've created an [feature request](https://github.com/argoproj/argo-events/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed the CLA.
* [ ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ ] My builds are green. Try syncing with master if they are not. 

Closes #926

### DIFF
--- a/common/retry.go
+++ b/common/retry.go
@@ -30,7 +30,7 @@ var DefaultRetry = wait.Backoff{
 	Steps:    5,
 	Duration: 1 * time.Second,
 	Factor:   1.0,
-	Jitter:   5,
+	Jitter:   1,
 }
 
 // IsRetryableKubeAPIError returns if the error is a retryable kubernetes error
--- a/common/retry.go
+++ b/common/retry.go
@@ -28,9 +28,9 @@ import (
 // DefaultRetry is a default retry backoff settings when retrying API calls
 var DefaultRetry = wait.Backoff{
 	Steps:    5,
-	Duration: 10 * time.Millisecond,
+	Duration: 1 * time.Second,
 	Factor:   1.0,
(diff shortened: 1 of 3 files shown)


whynowy:
@VaibhavPage - Can you review it?

VaibhavPage (approved):

PR #954, opened 2020-11-16
feat: Add string comparators EqualTo and NotEqualTo

jsalinaspolo:
Support comparator EqualTo and NotEqualTo in Data filters for Strings

### DIFF
--- a/sensors/dependencies/filter.go
+++ b/sensors/dependencies/filter.go
@@ -205,13 +205,24 @@ filter:
 
 		case v1alpha1.JSONTypeString:
 			for _, value := range f.Value {
-				exp, err := regexp.Compile(value)
-				if err != nil {
-					return false, err
-				}
+				switch f.Comparator {
+				case v1alpha1.EqualTo:
+					if res.Str == value {
+						continue filter
+					}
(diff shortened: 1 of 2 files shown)


sensors/dependencies/filter.go
```diff
@@ -205,13 +205,24 @@ filter:
 
 		case v1alpha1.JSONTypeString:
 			for _, value := range f.Value {
-				exp, err := regexp.Compile(value)
-				if err != nil {
-					return false, err
-				}
+				switch f.Comparator {
```
whynowy, line 208: A better logic might be:

case EqualTo: regex match
    
case NotEqualTo: !(regex match)

default: same as EqualTo

Reference to field filter logic in `resource` type event source (except `default`):
https://github.com/argoproj/argo-events/blob/master/eventsources/sources/resource/start.go#L288-L295

jsalinaspolo: Thanks @whynowy, I applied the changes.
I was not sure if would be better using regex or equallity, but at least it will be consistent with the rest.

whynowy (approved):
LGTM, thanks for the PR!

PR #981, opened 2020-12-04
Fix/eventsource resource return entire event object

scottweitzner:
Checklist:

### DIFF
--- a/eventsources/sources/resource/start.go
+++ b/eventsources/sources/resource/start.go
@@ -129,7 +129,7 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 		for {
 			select {
 			case event := <-informerEventCh:
-				objBody, err := json.Marshal(event)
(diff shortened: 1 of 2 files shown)


scottweitzner:
related to #982

eventsources/sources/resource/start.go
```diff
@@ -135,14 +135,25 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 					continue
 				}
```
VaibhavPage, line 137: `json.Marshal` on line 132 should be `json.Marshal(event.Obj)

VaibhavPage (changes requested):

eventsources/sources/resource/start.go
scottweitzner: good catch. fixed

VaibhavPage (approved):

PR #991, opened 2020-12-12
feat: ability to specify version in kafka trigger. Closes #990.

### LINKED ISSUE #990: Kafka trigger produces wrong timestamp
**Describe the bug**
When an event is pushed to kafka the associated timestamp is always "-1". 

**To Reproduce**
Just connect any eventSource with a kafka trigger and consume the message via kafka-console-consumer. Your output is going to look like this

`CreateTime:-1	{"message":"test"}`

**Expected behavior**
A valid timestamp should be send to kafka.

 - Kubernetes: v1.17.12
 - Argo Events: v1.1.0

**Additional context**
I guess this is the same problem: https://stackoverflow.com/questions/52704889/invalid-timestamp-when-writing-a-kafka-producer-with-sarama

---

whynowy:
Also defaults Kafka version to v1.0.0 which is the same as Kafka event source.

Closes #990.

Checklist:


pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -439,6 +439,10 @@ type KafkaTrigger struct {
 	// Defaults to broker url.
 	// +optional.
 	PartitioningKey string `json:"partitioningKey,omitempty" protobuf:"bytes,10,opt,name=partitioningKey"`
+
+	// Specify what kafka version is being connected to enables certain features in sarama, defaults to 1.0.0
+	// +optional
+	Version string `json:"version" protobuf:"bytes,11,opt,name=version"`
```
alexec, line 445: should be `version,omitempty`

sensors/triggers/kafka/kafka.go
```diff
@@ -49,6 +49,16 @@ func NewKafkaTrigger(sensor *v1alpha1.Sensor, trigger *v1alpha1.Trigger, kafkaPr
 		var err error
 		config := sarama.NewConfig()
 
+		if kafkatrigger.Version == "" {
```
alexec, line 52: I'd add a `kafkatrigger.GetVersion() (string, error)`

pkg/apis/sensor/v1alpha1/types.go
whynowy: Good catch!

sensors/triggers/kafka/kafka.go
whynowy: Dot not want to import `sarama` package into models.

alexec: good point

alexec (approved):

PR #992, opened 2020-12-14
fix: handle slack slash command payloads. Fixes #653

### LINKED ISSUE #653: Slack incoming events error failed to parse interaction event: unexpected end of JSON input
**Describe the bug**
Used the setup from examples for a Slack incoming gateway and sensor. When the Slack app sends a message the gateway receives the POST but fails on parsing the interaction. I think this is related to how Slack sends all the data as URL parameters instead of a JSON body. 

**To Reproduce**
1. Setup up a Slack app and add secrets to argo-events namespace (Same process as https://argoproj.github.io/argo-events/triggers/slack-trigger/)
2. Add Slack gateway, sensor, and event source from examples
3. Setup an external ingress or load balancer that Slack can hit
4. Update Slack app and add a slash command with URL for the ingress/load balancer
3. In slack test the slash command you added

**Expected behavior**

**Screenshots**

- OS: amazon-linux-2
- Kubernetes: v1.16.8-eks-e16311
- Version: argoproj/slack-gateway:v0.14.0

**Additional context**
POST data sent from Slack to test webhook catcher:
```
POST /test HTTP/1.1

Host: slack.requestcatcher.com
Accept: application/json,*/*
Accept-Encoding: gzip,deflate
Content-Length: 357
Content-Type: application/x-www-form-urlencoded
User-Agent: Slackbot 1.0 (+https://api.slack.com/robots)
X-Slack-Request-Timestamp: 1588953976
X-Slack-Signature: v0=...

token=token&team_id=team_id&team_domain=team_domain&channel_id=channel_id&channel_name=argo&user_id=user_id&user_name=user_name&command=%2Ftest&text=message&response_url=response_url&trigger_id=trigger_id
```

lheinlen:
Checklist:

I believe that this should correct the issue described by #653. We are using it internally to generate events for both interactions and slash commands.


eventsources/sources/slack/start.go
```diff
@@ -118,14 +118,36 @@ func (rc *Router) HandleRoute(writer http.ResponseWriter, request *http.Request)
...
-	// If request was generated by an interactive element, it will be a POST form
+	// If request was generated by an interactive element or a slash command, it will be a POST form
 	if len(request.Header["Content-Type"]) > 0 && request.Header["Content-Type"][0] == "application/x-www-form-urlencoded" {
-		logger.Info("handling slack interaction...")
-		data, err = rc.handleInteraction(request)
-		if err != nil {
-			logger.Desugar().Error("failed to process the interaction", zap.Error(err))
+		if err := request.ParseForm(); err != nil {
+			logger.Desugar().Error("failed to parse form data", zap.Error(err))
+			common.SendInternalErrorResponse(writer, err.Error())
```
whynowy, line 128: Missing a `return` here.

whynowy (commented):
Just one small thing need to be fixed.

Thanks for your contribution!

eventsources/sources/slack/start.go
lheinlen: Yikes! Fixed.

whynowy (approved):
LGTM!

Thanks!

PR #997, opened 2020-12-18
fix: Build and publish latest images automatically

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>

Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,7 @@ on:
       - 'v*'
     branches:
       - master
+      - fix-latest
 
 jobs:
   release:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,6 @@ on:
       - 'v*'
     branches:
       - master
-      - fix-latest
 
 jobs:
   release:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - 'v*'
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -42,4 +45,7 @@ jobs:
           DOCKERIO_ORG: ${{ secrets.DOCKERIO_ORG }}
         run: |
           tag=$(basename $GITHUB_REF)
+          if [ $tag = "master" ]; then
```
whynowy, line 48: Need clarification on `latest` tag, we define it as latest master build, or latest stable build? I had discussion with @VaibhavPage before, he thinks it should be the latter.

.github/workflows/release.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     tags:
       - 'v*'
+    branches:
+      - master
+      - fix-latest
```
whynowy, line 9: rm

.github/workflows/release.yml
alexec: 1. It's currently neither of those. It is 3 months old.
2. This is not consistent with either workflows or CD. Both of which are the tip of master.
3. That isn't it's intent: https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375

.github/workflows/release.yml
alexec: done

whynowy (approved):
LGTM

PR #1001, opened 2020-12-24
chore: Upgrade to k8s 1.9 and controller-runtime v0.7. Closes #999

### LINKED ISSUE #999: Upgrade go-client to v1.19
See https://github.com/argoproj/argo/issues/4785
We need this for Argo Workflows v3.0
We also need v1.2 to be released for https://github.com/argoproj/argo/issues/888

whynowy:

Checklist:

Closes #999


eventsources/common/aws/aws_test.go
```diff
@@ -68,7 +68,7 @@ func TestAWS(t *testing.T) {
 	})
 
 	convey.Convey("Given kubernetes secret that hold credentials, create AWS credential", t, func() {
-		secret, err := client.CoreV1().Secrets(namespace).Create(context.Background(), &corev1.Secret{
+		secret, err := client.CoreV1().Secrets(namespace).Create(context.TODO(), &corev1.Secret{
```
alexec, line 71: I think that `Background` might be fine here. Practically `Background` and `TODO` are the same - empty contexts, but semantically they are different. I believe you should use `TODO` when you think this will need to changed in the future. If this is not the case, I think it is fine to leave it.

alexec (approved):
I approve this. I'd like @VaibhavPage to give is blessing and we should decide when v1.2 should land

PR #1018, opened 2021-01-13
fix(sensor-controller): Return err when a sensor spec is invalid. Fixes #1017

### LINKED ISSUE #1017: Invalid sensor spec still generates deployments
**Describe the bug**
When an invalid sensor spec is provided (e.g. same event source is referenced multiple times), the Sensor status reports failure however a deployment is still created.

**Expected behavior**
The deployment should not be created.

**Screenshots**

 - Kubernetes: [e.g. v1.18.6]
 - Argo Events: [e.g. v1.2.0]

**Additional context**

---

whynowy:
Fixes #1017

Checklist:


alexec (approved):

nielstenboom:
Would it be possible to undo this change? Because after this change new deployments to Kubernetes with Argo events objects fail silently if you made a mistake in the spec. Before this change the deployment would end up in a Crashloopbackoff causing alerts to go off and ArgoCD to show visually something is wrong, this is not the case any longer. I ran into this problem today. 

I'd like to argue that generating a deployment when you made an error in the spec is safer.

PR #1021, opened 2021-01-15
feat: Introducing a Validating Admission Controller

### LINKED ISSUE #1019: Introducing a ValidatingWebhook admisson controller to validate specs
Spec validation is something needed for all the CRD objects, for example, following operations are not expected, such as creating a Sensor object without defining any dependencies, updating auth strategy of an EventBus object, and so on.

Currently spec validation happens in the controller reconciliation, which means users will not know if the spec he/she provided, or the operation he/she did is allowed until he/she checks the `status` field of the CRD object.

To let the user know if the spec or operation is valid earlier, we can introduce a [ValidatingAdmissionWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook), so that people can know the errors before the spec is stored in etcd.

This will introduce:

1. A deployment and a Service running in `argo-events` namespace, it implements the validation logic.
2. A `ValidatingWebhookConfigurations` object for admission controller configuration. It will be automatically created by the deployment in 1.)

With this validating webhook, right after an invalid spec is applied, you will get an error message like below (assume a native EventBus requires a replica number >=3):
```
kubectl apply -f test-eventbus.yaml
Error from server (BadRequest): error when creating "test-eventbus.yaml": admission webhook "webhook.argo-events.argoproj.io" denied the request: Replicas must be no less than 3
```

**Additional context**
This validating webhook will be included in the cluster scope installation spec, might not be in the namespace scope installation.

---

whynowy:
This PR introduces a Validating Admission Controller to Argo Events, it validates the input spec and notifies the errors when doing `kubectl apply`.

Main features:

1. Validating CREATE actions for EventBus, EventSource and Sensor objects;
2. Validating UPDATE actions to see:  
- If the new spec is valid;
- Is there any immutable fields being updated (e.g. `spec.nats.native.auth` in EventBus can not be updated);

This change brings in a new Deployment and Service for the validating webhook, currently it would be made optional for installation.

Closes #1019.

Checklist:


manifests/install-no-extension.yaml
```diff
@@ -0,0 +1,370 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
VaibhavPage, line 1: Let's call this file `install-with-extension`. By default Argo-Events should be installed without Validating webhook as it is an addon.

whynowy: I made it a separate installation file `install-validating-webhook.yaml`.

common/tls/tls.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+func certTemplate(org string, hosts []string, notAfter time.Time) (*x509.Certificate, error) {
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to generate serial number")
```
alexec, line 19: minor: I much prefer `fmt.Errorf("%w")`

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/fatih/color v1.9.0 // indirect
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
+	github.com/go-openapi/inflect v0.19.0
```
alexec, line 28: what licence?

alexec (approved):
Conditional approval:

* I understand this is low-risk change because it is optional and orthogonal.

go.mod
whynowy: MIT.

whynowy: MIT

PR #1030, opened 2021-01-22
feat: use status resources for controller reconciliation. Closes #1029

### LINKED ISSUE #1029: Update status subresource in controller reconciliation
Currently during controller reconciliation, we use CRD object update to update status field, this leads to another UPDATE notification to the controller. And there's a bug that the controller checks if the old object status is deep equal the reconciled object status to determine if needs another update, however because the condition lastTransitionTime always changes, so it keeps updating the object status.

1. Enable `status` subresource, use status update api to the object status;
2. Always update status field when there's reconciliation loop;
3. Add filters to controller watcher to only watch `generation` changes (which is caused by a spec change).

**Additional context**

---

whynowy:

Checklist:

Closes #1029

This PR:
1. Enabled `status` subresource for CRDs, and controllers use `status` update API to update object status;
2. Controllers do not watch `status` change any more;
3. Fixed a bug that it kept emitting UPDATE notifications and updating status field infinitely.
4. Some code refactory and enhancement.

### DIFF
--- a/controllers/eventbus/cmd/main.go
+++ b/controllers/eventbus/cmd/main.go
@@ -3,15 +3,18 @@ package main
 import (
 	"flag"
 	"os"
+	"reflect"
 
 	"go.uber.org/zap"
 	appv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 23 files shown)


alexec:
How do we test this please?

whynowy:
Here is the data about reconciliation times and workqueue adding times after applying an EventSource object.

Before:
u01
u02

After:
u03
u04

whynowy:
> How do we test this please?



> How do we test this please?

We should get some e2e test into the project, I can work on this later. For now, I had to test manually.

alexec (approved):

Issue #1040, opened 2021-01-27
Delete functionality for k8s trigger


tinkerborg:
Currently the standard k8s trigger does not support delete operations.

I would like to be able to use this for things like creating an argo-cd Application object when a pull request is opened in Github, and deleting the object when the associated PR is closed. 

As an alternative, I could trigger a job but this would be a much cleaner approach.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

Issue #638, opened 2020-05-05
Deleting a deployment


lueenavarro:
So i was a able to create a deployment using argo-events, but I could not find any example on how to delete a deployment and if it is even possible. Do you have examples? I believe I just have to add a few lines on the yaml config below but I do not know what to add. 

Lets say what if I want to delete `example-deployment` in the `default` namespace

```
apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: webhook-sensor-http
spec:
  template:
    spec:
      containers:
        - name: sensor
          image: argoproj/sensor:v0.13.0
          imagePullPolicy: Always
      serviceAccountName: argo-events-sa
  dependencies:
    - name: test-dep
      gatewayName: webhook-gateway
      eventName: example
  subscription:
    http:
      port: 9300
  triggers:
    - template:
        name: webhook-deployment-trigger
        k8s:
          group: apps
          version: v1
          resource: deployments
          operation: delete
```

VaibhavPage:
`delete` operation is currently not supported.

PR #1041, opened 2021-01-27
feat: implement delete function for k8s objects

tinkerborg:
This implements `delete` functionality for the standard-k8s trigger.

resolves argoproj/argo-events#1040
fixes argoproj/argo-events#638


whynowy:
@VaibhavPage - Is there any reason we didn't do `DELETE` before?

sensors/triggers/standard-k8s/standar-k8s.go
```diff
@@ -212,6 +212,22 @@ func (k8sTrigger *StandardK8sTrigger) Execute(ctx context.Context, events map[st
 
 		return k8sTrigger.namespableDynamicClient.Namespace(namespace).Patch(ctx, obj.GetName(), k8sTrigger.Trigger.Template.K8s.PatchStrategy, body, metav1.PatchOptions{})
 
+	case v1alpha1.Delete:
+		k8sTrigger.Logger.Info("deleting the object...")
+		_, err := k8sTrigger.namespableDynamicClient.Namespace(namespace).Get(ctx, obj.GetName(), metav1.GetOptions{})
+
+		if err != nil && apierrors.IsNotFound(err) {
+			k8sTrigger.Logger.Info("object not found, nothing to delete...")
```
whynowy, line 220: Need to return something here.

whynowy (changes requested):

sensors/triggers/standard-k8s/standar-k8s.go
tinkerborg: I made it return nil, nil for no-op, which I'm pretty sure makes sense here

whynowy:
Fix the DCO please.

whynowy (approved):
LGTM, thanks!

PR #1047, opened 2021-02-02
feat: Expose Prometheus metrics. Closes #204

### LINKED ISSUE #204: Expose prometheus metrics in sensors and eventsources
It's useful to be able to monitor argo by collecting and analyzing prometheus metrics.  I can alert on failed  workflows easily. I can track activity over time, etc.

I would like to be able to do the same within argo-events.  monitor the system for successes, failures, etc.

for example, if I am using a trigger with a remote source (git|http|s3|etc) in a sensor, it would be useful to emit a metric when there's a failure to fetch this.  right now these types of errors are difficult to catch

whynowy:

Closes #204

Checklist:


sensors/metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/argoproj/argo-events/common/logging"
+)
+
+const (
+	prefix = "argo_events"
+)
+
+// Metrics represents EventSource metrics information
+type Metrics struct {
```
alexec, line 19: I'm sure workflows was correct to have a metrics service- when global variables are easier to use

whynowy: Merged them to one struct.

eventsources/eventing.go
```diff
@@ -316,7 +320,9 @@ func (e *EventSourceAdaptor) Start(ctx context.Context) error {
 				continue
 			}
 			wg.Add(1)
+			e.metrics.IncRunningServices(server.GetEventSourceName())
```
alexec, line 323: move one line down so code change cannot create regression?

docs/metrics.md
```diff
@@ -0,0 +1,94 @@
...
+rules:
+  - apiGroups: [""]
+    resources:
+      - pods
+    verbs: ["get", "list", "watch"]
+```
+
+### EventSource
+
+- `event_service_running_total` - How many configured events in the EventSource
```
alexec, line 55: can I ask for a section state which metrics is which Golden Metric?

docs/metrics.md
```diff
@@ -52,15 +52,16 @@ rules:
 
 ### EventSource
 
-- `event_service_running_total` - How many configured events in the EventSource
```
alexec: in Argo Workflow the Prometheus HELP contains deep-links to the docs, you could do this same here, allowing you to change the help with a release

docs/metrics.md
```diff
@@ -0,0 +1,130 @@
...
+  - `argo_events_action_triggered_total`
+
+- Errors
+
+  - `argo_events_events_sent_failed_total`
+  - `argo_events_action_failed_total`
+
+- Saturation
+
+  - TBD.
```
alexec, line 130: this is probably covered by out of the box Kubernetes metrics such as CPU or memory (RSS)

docs/metrics.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Golden Signals
+
+Following metrics are considered as
+[Golden Signals](https://sre.google/sre-book/monitoring-distributed-systems/#xref_monitoring_golden-signals)
+of monitoring your applictions running with Argo Events.
+
+- Latency
+
+  - TBD.
```
alexec, line 116: "The time it takes to service a request. "

What is a request in events? 

* Event Source - the time from getting event to send message to EventBus?
* EventBus - the time from receive a NATS message to dispatching it?
* Sensor - the time from receiving event to getting trigger finishing?

whynowy: - Added the latency for Sensor triggering actions.
- Event Source latency would be added in future PRs as it requires to make code change to every event source implementation.
- EventBus latency needs further investigation - we could not calculate it from the clients side (when eventsource sends it, and when sensor receives it), but should look into if NATS exposes this kind of metrics, I haven't seen it so far, need more research.

alexec (commented):
please set-up 20m

docs/metrics.md
```diff
@@ -0,0 +1,141 @@
...
+How many events have been sent successfully.
+
+#### argo_events_events_sent_failed_total
+
+How many events failed to send.
+
+#### argo_events_event_processing_duration_milliseconds
+
+Event processing duration (from getting the event to send it to EventBus) in
+milliseconds - TBD.
```
alexec, line 70: TBD?

alexec (approved):

PR #1058, opened 2021-02-08
fix(slack trigger): properly iterate api response

### LINKED ISSUE #1057: Slack Trigger - Incorrect assumption about API response leads to erros
**Describe the bug**
Slack Trigger, iterates all slack channels returned from the API (via pagination) to find appropriate channel ID - [that](https://github.com/argoproj/argo-events/blob/master/sensors/triggers/slack/slack.go#L127)  implementation makes incorrect assumption - it expects the count of returned items to always be equal to pagination limit, unless last page is returned.

Such assumption clearly goes again the Slack API design - see [link](https://api.slack.com/methods/conversations.list#arg_limit), which states that returned chunk of data may be smaller than pagination limit, even if that is not the last page or results. 

With large enough and dynamic enough Slack Workspace, aforementioned implementation leads to 'failed to get channelID of %s' exceptions, even if channel with appropriate name exists - the pagination simply stops, before it can reach proper item.

**To Reproduce**
1. Have Slack API return list of channels which does not neatly have 200 items per page
2. Choose channel to sent message to, which is at the end of the pagination
3. Attempt to run Slack Trigger
4. See error 'failed to get channelID of %s'

**Expected behavior**
I would expect the Slack Trigger to work as long as the channel with given name exists and is returned by the API pagination, irrespectively of pages size. 

 - Kubernetes: v1.19.6
 - Argo: v2.12.7
 - Argo Events: v1.2.2

**Additional context**
Proposed fix: Simply paginate until returned cursor is empty. 

---

AleksanderGondek:
Fixes #1057  The iteration should not assume every page has size of limit set (200)

Checklist:

### DIFF
--- a/sensors/triggers/slack/slack.go
+++ b/sensors/triggers/slack/slack.go
@@ -124,7 +124,7 @@ func (t *SlackTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
 				break
 			}
 		}
-		if len(channels) < params.Limit || nextCursor == "" {
+		if nextCursor == "" {
 			break
 		}
 		params.Cursor = nextCursor


whynowy (approved):
LGTM, thanks for fixing it!

PR #1076, opened 2021-02-19
fix: tls config validation, follow up on #1070

whynowy:

Checklist:

### DIFF
--- a/pkg/apis/common/validate.go
+++ b/pkg/apis/common/validate.go
@@ -1,0 +1,32 @@
+package common
+
+import "errors"
+
+// ValidateTLSConfig validates a TLS configuration.
+func ValidateTLSConfig(tlsConfig *TLSConfig) error {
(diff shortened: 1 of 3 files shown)


pkg/apis/common/validate.go
```diff
@@ -0,0 +1,32 @@
+package common
+
+import "errors"
+
+// ValidateTLSConfig validates a TLS configuration.
+func ValidateTLSConfig(tlsConfig *TLSConfig) error {
```
alexec, line 6: do you actually need to validate it at all? won't it just fail if it is wrong?

alexec (approved):

pkg/apis/common/validate.go
whynowy: This validation is called by controller reconciliation not the real eventsource application, so it's needed.

PR #1079, opened 2021-02-19
Add missing apiVersion to EventBus example

scompt:
Checklist:

* [x] ~~My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).~~


whynowy (approved):
LGTM, thanks for fixing it!

Issue #1097, opened 2021-03-04
Allow for JSON string to be processed in the Sensor DataFilter


jayfray12:
**Is your feature request related to a problem? Please describe.**
When using gjson to evaluate the path of the DataFilter it returns a [`Result`](https://github.com/tidwall/gjson#result-type).  You are using `Result.Str` [here](https://github.com/argoproj/argo-events/blob/master/sensors/dependencies/filter.go#L213).  With gjson you can return a JSON object by utilizing a [Multipath](https://github.com/tidwall/gjson/blob/master/SYNTAX.md#multipaths).  However, when utilizing this Multipath the `Result.Type` is JSON and thus the `Result.Str` returns an empty string and nothing is ever matched by the filter.  You can however call [`Result.String()`](https://github.com/tidwall/gjson/blob/master/gjson.go#L70) which correctly returns the string based on the `Result.Type`.

**Describe the solution you'd like**
What I'm trying to achieve is I have a webhook defined for an event and I want to filter the calls based on a value and depending on that value I may or may not want to filter if another value is something.  For example:
```json
{
  "body": {
    "action":"opened",
    "labels": [
      {"id":"1234", "name":"Webhook"}, 
      {"id":"5678", "name":"Approved"}
    ]
  }
}
```
I want the sensor to fire if my action is "opened" and it has a label of "Webhook" or if the action is "closed" and it has a label of "Webhook" and "Approved".  Given the value of the DataFilter can be a regex I can use the gjson syntax to do something like:
```
[body.action,body.labels.#(name=="Webhook").name,body.labels.#(name=="Approved").name]
```
This produces a String like this:
```
["opened","Webhook","Approved"]
```
Then I could do a regex with something like:
```
(\bopened\b.*\bWebhook\b)|(\blabeled\b.*(\bWebhook\b.*\bApproved\b))
```
To filter the way I want.  However, like I mentioned above the `Result.Str` is empty when the `Result.Type` is JSON and thus would need to code changed to call `Result.String()`

**Describe alternatives you've considered**
I tried I ton of different ways to filter based on what I want but can't come up with anything else.

**Additional context**

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

jayfray12:
Issue resolved with PR

PR #1098, opened 2021-03-04
Changed Result.Str to Result.String()

jayfray12:
Updated Result.Str to be more correct and use Result.String().  This fixes issue #1097 

Checklist:

* [ x ] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).


whynowy:
Thanks for the detailed description in the reported issue! Could you add some test cases in filter_test.go?

jayfray12:
> Thanks for the detailed description in the reported issue! Could you add some test cases in filter_test.go?

Sure thing!

whynowy (approved):
LGTM, thanks for you contribution!

PR #1116, opened 2021-03-15
fix(docs): Update link to Argo Workflows

AnaisUrlichs:
Checklist:


whynowy (approved):
Thanks for fixing it!

PR #1118, opened 2021-03-15
chore: upgrade kube libraries to 0.20.4

sbose78:
Checklist:


whynowy:
Workflows and Events have the agreement that we would like to stay 1 major version behind of the latest, which is v1.19.x.

sbose78:
Gotcha, thanks. 

Curious, what's the reason for this? :)

whynowy:
> Gotcha, thanks.
> 
> Curious, what's the reason for this? :)

Just don't want to be too aggressive, and it's a safer option.

sbose78:
Got it, thanks.

PR #1123, opened 2021-03-20
feat(sensor): add encoding decode in filter matching

tczhao:
When using GCP pubsub with Google cloud storage, one of the field ["data"] from the json message is encoded in base64.

This PR added the functionality to sensor-filter,
enables decoding the field before doing the comparator matching

Signed-off-by: Tianchu Zhao <evantczhao@gmail.com>

Checklist:


whynowy:
Instead of providing a `encoding` option, I more prefer to make it support `Sprig` template like the feature in `parameters`, and then you could use [encode/decode](http://masterminds.github.io/sprig/encoding.html) features provided by Sprig to achieve that.

Check the `Sprig` parameter implementation at https://github.com/argoproj/argo-events/blob/master/sensors/triggers/params.go.

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -249,6 +249,11 @@ type DataFilter struct {
 	// Can be ">=", ">", "=", "!=", "<", or "<=".
 	// Is optional, and if left blank treated as equality "=".
 	Comparator Comparator `json:"comparator,omitempty" protobuf:"bytes,4,opt,name=comparator,casttype=Comparator"`
+	// DataTemplate is a go-template for extracting a string from the event's data.
```
whynowy, line 252: Let's name it `template` directly, which means `template function`.

This `template` will work together with `path`, it uses the result of `path` as its input, for example, spec below means lower case of `body.value`.

```
            filters:
            data:
              - path: body.value
                 template: {{ .Input | lower }}
```

The template should work not only string, but types like number, it is used to additional processing for the value of `path` before filtering.

sensors/dependencies/filter.go
```diff
@@ -149,6 +152,43 @@ func filterData(data []v1alpha1.DataFilter, event *v1alpha1.Event) (bool, error)
 	}
 filter:
 	for _, f := range data {
+		if f.DataTemplate != "" {
```
whynowy, line 155: Could you change the logic based on the comments I gave above?

whynowy (commented):
Appreciate if you could also update the doc below in the PR, together with some examples in `./examples/sensors`.

https://github.com/argoproj/argo-events/blob/master/docs/tutorials/07-filters.md

Thanks!

tczhao:
Thanks for the review
updated, all comments addressed

whynowy (approved):
LGTM, thanks!

Issue #1112, opened 2021-03-15
Deprecate Github Webhook creation feature


whynowy:
**Is your feature request related to a problem? Please describe.**
`GithubEventSource` provides a feature to set up a Github webhook when the event source starts, this actually brings some problems which are difficult to resolve.

- When creating the webhook in Github with `GithubEventSource`, as a user, the ingress (a LB Service or other kinds of ingress) for the event source is most probably not created yet (because no matter it's a LB Service or whatever ingress, you need to know the Pod Selector Labels to create it, the user doesn't know the labels before the EventSource Pod is in place unless he or she is very familiar with Argo EventS source code), so the URL to be configured in the GithubEventSource is still unknown.
-  Webhook creation is just a one time effort, it's very easy to configure it in Github Settings GUI, comparing with setting it up in the GUI, I don't believe most of the users prefer to configure a Kubernetes Secret for the API token, and let the GIthubEventSource do it.
- **[Fixed]** There's a bug currently in Argo Events source code. When the EventSource Pod starts, if the creating actions responds that the webhook is already existing, Argo Events tries to compare the URL and events from the existing webhook list with the configuration in the spec to find the existing one. Github stores the order of configured events, but the code doesn't sort and compare, so if you create the EventSource with `pull_request, push`, and then update the order to `push, pull_request`, it will cause to an error, it complains the webhook is already existing, but failed to find the webhook. 
- **[Fixed]** If the events list is something like `*, push, pull_request`, it actually create a webhook with `all_events`, and this brings another bug, the events list comparing again above always return false.
- Github allows same webhook URL to be configured multiple times, for example, `webhook1 (pull_request, push)` and `webhook2 (pull_request, create)` with same URL are valid. So if you create a GithubEventSource with `pull_request, push`, it creates a webhook, later on, you update the EventSource to only have `pull_request`, it creates another webhook. This is very tricky for Argo Events to handle this situation.

**Describe the solution you'd like**
1. Deprecate the webhook creating feature, the GithubEventSource would only be responsible for handling the post events from Github.
2. It doesn't care about what the repository is (or it does if a valid list is given).
3. Events config in the spec are still available, that only means what are the interested events types (event though Github already has the events configured), and the events config could be extended, such as pull_request of what branch, etc.

Typical operations of using GithubEventSource would be:

1. Create a GithubEventSource;
2. Create a LB Service (or other kinds of ingress) for the event source, you get the URL;
3. Configure a webhook in Github Settings GUI.

---

[...]

PR #1124, opened 2021-03-22
fix(GithubEventSource): Compare events ignoring order and duplicate

whynowy:

Fix the bugs mentioned in https://github.com/argoproj/argo-events/issues/1112.

Function `sliceEqual` is used to compare the Github events listed in the spec and the ones from Github API response.

It should:

1. If "*" (`all_events`) is in both slices, that should be considered as equal.
2. It should compare ignoring the duplicate and order, which means `{"a", "c", "a", "b"}` = `{"b", "c", "c", "a"}` = `{"a", "b", "c"}`

Checklist:

### DIFF
--- a/eventsources/sources/github/hook_util.go
+++ b/eventsources/sources/github/hook_util.go
@@ -1,30 +1,62 @@
 package github
 
 import (
+	"sort"
+
 	gh "github.com/google/go-github/v31/github"
 )
 
 // sliceEqual returns true if the two provided string slices are equal.
 func sliceEqual(first []string, second []string) bool {
-	if len(first) != len(second) {
+	if len(first) == 0 && len(second) == 0 {
+		return true
+	}
+	if len(first) == 0 || len(second) == 0 {
 		return false
 	}
 
-	if first == nil && second == nil {
+	uniqFirst := unique(first)
+	uniqSecond := unique(second)
+
+	sort.Strings(uniqFirst)
+	sort.Strings(uniqSecond)
+
+	// all_request
+	if contains(uniqFirst, "*") && contains(uniqSecond, "*") {
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


eventsources/sources/github/hook_util.go
```diff
@@ -1,32 +1,64 @@
...
 	return true
 }
 
+func contains(sortedList []string, str string) bool {
+	i := sort.SearchStrings(sortedList, str)
+	return i < len(sortedList) && sortedList[i] == str
+}
+
+func unique(stringSlice []string) []string {
+	if len(stringSlice) == 0 {
```
alexec, line 48: maybe some util for this some already?

eventsources/sources/github/hook_util_test.go
```diff
@@ -19,6 +19,9 @@ func TestSliceEqual(t *testing.T) {
 	assert.False(t, sliceEqual([]string{"hello"}, []string{"hello", "world"}))
 	assert.False(t, sliceEqual([]string{"hello", "world"}, []string{"hello"}))
 	assert.False(t, sliceEqual([]string{"hello", "world"}, []string{"hello", "moon"}))
+	assert.True(t, sliceEqual([]string{"hello", "world"}, []string{"world", "hello"}))
```
alexec, line 22: too few tests for code changes

eventsources/sources/github/hook_util.go
alexec: https://stackoverflow.com/questions/36000487/check-for-equality-on-slices-without-order

eventsources/sources/github/hook_util_test.go
whynowy: Added more.

alexec (approved):

Issue #1120, opened 2021-03-16
Github template source failing with `some refs were not updated`


rory-lamendola:
**Describe the bug**
The github template source starts failing with `failed to fetch. err: some refs were not updated` after about 30 minutes to an hour after the sensor is created.  This happens even when not pulling from a particular ref or branch, just pulling from master.  From what I understand, setting the ref should result in a shallow clone and fetch, so not sure why it wouldn't be able to update that ref.  And it makes even less sense for that to fail when removing that ref, and just using the default master branch. 

**To Reproduce**

Sensor.yaml:
```
---
apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: github
  namespace: argo
spec:
  template:
    serviceAccountName: argo-events-sa
    containers:
      - volumeMounts:
          - mountPath: /git/shared
            name: shared
    volumes:
      - name: shared
        emptyDir: {}
  dependencies:
    - name: webhook-dep
      eventSourceName: github
      eventName: webhook
  triggers:
    - template:
        name: workflow-trigger
        k8s:
          group: argoproj.io
          version: v1alpha1
          resource: workflows
          operation: create
          source:
            git:
              url: "git@github.com:petal-engineering/shared.git"
              cloneDirectory: "/git/shared"
              sshKeySecret:
                name: github
                key: ssh-private-key
              filePath: "ci/argo/ci.yml"
      parameters:
        - src:
            dependencyName: webhook-dep
            dataKey: body.pull_request.head.ref
          dest: k8s.source.git.branch
```

I'm not able to reproduce it using the public argo workflow repo instead of our private repo.

**Expected behavior**
I expect the sensor to trigger the yaml that exists on the given branch on github.  

**Screenshots**

**Environment (please complete the following information):**
 - Kubernetes: 1.18.9 (eks)
 - Argo: v0.16.7
 - Argo Events: v1.2.3

**Additional context**

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
@rory-lamendola could you remove `cloneDirectory` from the spec (and the `volume` and `volumeMount` are not needed), and see if the issue still exits?

rory-lamendola:
[...]

PR #1128, opened 2021-03-23
fix(git-sensor): force fetch from git repository

rory-lamendola:
We've run into an issue with using the existing git sensor with our repository.  The sensor works for a while until it starts erroring on `failed to fetch. err: some refs were not updated`.  This seems to happen when we merge code to our repository, and our deploy process updates tags.  Updating tags on existing branches causes git to throw an error unless you force fetch.  The artifact already force fetches when given a particular ref, and I can't think of a downside of doing it everywhere given that you wouldn't be making local changes to branches in a sensor.

Fixes: #1120

Checklist:

### DIFF
--- a/sensors/artifacts/git.go
+++ b/sensors/artifacts/git.go
@@ -125,6 +125,7 @@ func (g *GitArtifactReader) readFromRepository(r *git.Repository, dir string) ([
 		fetchOptions := &git.FetchOptions{
 			RemoteName: g.artifact.Remote.Name,
 			RefSpecs:   fetchRefSpec,
+			Force:      true,
 		}
 		if auth != nil {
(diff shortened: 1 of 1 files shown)


whynowy:
I believe removing `cloneDirectory` will also fix your issue, but that will make it re-clone the repo every time when there's an event. Adding `force` seems to be reasonable to me. Is this PR ready for review?

rory-lamendola:
@whynowy yeah, do you want me to fix the semantic pull request checker though?

whynowy:
> @whynowy yeah, do you want me to fix the semantic pull request checker though?

No really, you are good.

whynowy (approved):
LGMT, thanks!

PR #1140, opened 2021-03-28
feat: azure event hubs trigger

### LINKED ISSUE #1132: Azure Event Hub Trigger
There doesn't seem to be a trigger dedicated for Azure Event Hubs. Though you can use a Kafka Sensor to use Azure Event Hubs (since you can deploy a managed Kafka in Azure Event Hubs), it would still be nice to have this integration. The sample trigger in https://github.com/argoproj/argo-events/blob/master/examples/sensors/azure-events-hub.yaml triggers an Argo Workflow. I also can't seem to find the code for Event Hubs in the `sensors/trigger` folder

A sensor for Azure Event Hubs

Use [Event Hubs for Kafka](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-for-kafka-ecosystem-overview) as the `Sensor` trigger, however, I want to leverage the other features in Azure Event Hubs.

**Additional context**
I can contribute the feature if this is something of interest to the community.

---

joshuajorel:
Checklist:

* [x] Fixes #1132 

Test Cases:

- [x] Send to Azure Event Hubs
- [x] Test sending invalid payload (key does not exist from event payload)
- [x] Test sending event to invalid FQDN
- [x] Test sending event to invalid  HubName
- [x] Test sending event with invalid sharedAccessKeyName
- [x] Test sending event with invalid sharedAccessKey
- [x] Test parametrization on payload values

Notes:
- parametrization will not work for `fqdn` and `hubName` parameters since the Azure Event Hubs Client is initialized in the `NewAzureEventHubsTrigger` function. Should this be the case, I'll need to add code that does this (but might be a bit tricky since these values are required to initialize the client. Suggestions are welcome on how to solve this
- advanced features for Azure Events Hub such as sending with a message ID or sending to a specific partition are yet to be done. I can probably commit this as another feature (just to reduce the scope of this PR)


whynowy:
It looks good to me over all, could you add a doc under https://github.com/argoproj/argo-events/tree/master/docs/triggers, and an example to https://github.com/argoproj/argo-events/tree/master/examples/sensors?

For the first note, I think mentioning it in the doc is good enough for now. This is a common issue for most of the triggers,  but I don't recommend to solve it unless there are many use cases like that.

Thanks for your contribution!

joshuajorel:
@whynowy for your review. Thanks!

joshuajorel:
@whynowy I will be forcing a new commit. Just realized I committed over an existing document. Apologies for the confusion.

docs/triggers/azure-event-hubs.md
```diff
@@ -0,0 +1,88 @@
...
+          sharedAccessKey: <base64-shared-access-key>
+          sharedAccessKeyName: <base64-shared-access-key-name>
+
+1. Let's set up a webhook event-source to process incoming requests.
+
+        kubectl apply -n argo-events -f https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/event-sources/webhook.yaml
+
+1. Create the sensor with the following template. Replace the necessary values for `fqdn` and `hubName`:
+
+        apiVersion: argoproj.io/v1alpha1
```
whynowy, line 38: Incorrect indents.

docs/triggers/azure-event-hubs.md
```diff
@@ -0,0 +1,88 @@
...
+        kind: Sensor
+        metadata:
+        name: azure-events-hub
+        spec:
+        template:
```
whynowy, line 43: template/serviceAccountName here is not needed here, a `default` service account would work.

examples/sensors/azure-event-hubs-sensor.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Sensor
+metadata:
+  name: azure-events-hub
+spec:
+  template:
```
whynowy, line 6: `serviceAccountName` not needed.

docs/triggers/azure-event-hubs.md
joshuajorel: Adjusted the indentation.

docs/triggers/azure-event-hubs.md
joshuajorel: Removed service account reference

examples/sensors/azure-event-hubs-sensor.yaml
joshuajorel: Removed `serviceAccountName` reference.

whynowy (approved):
LGTM, thanks!

PR #1145, opened 2021-03-30
docs: fix gitlab -> github

kandros:
fix docs


whynowy (approved):
Thanks for fixing it!

Issue #338, opened 2019-08-14
Trigger source git doesn't work when private repo via HTTPS URL.


hidekuro:
**Describe the bug**
Trigger source git doesn't work when using bitbucket private repos via HTTPS URL.

```
ERRO[2019-08-14 01:50:43] trigger failed to execute                     error="failed to clone repository. err: authentication required" trigger-name=my-trigger-1
```

**To Reproduce**

1. import repository https://github.com/argoproj/argo to bitbucket as private repo.
2. create sandbox environments (see below)

create minikube VM and install helm.

```bash
minikube start --memory=4096
helm init

echo "waiting for helm server tiller..."
until (helm version -s); do sleep 3; done
echo "tiller ready."
```

setup argo-workflows, argo-events

```bash
# argo-wf
kubectl create namespace argo
kubectl apply -n argo -f https://raw.githubusercontent.com/argoproj/argo/stable/manifests/install.yaml
kubectl create rolebinding default-admin --clusterrole=admin --serviceaccount=default:default

# argo-events
kubectl create namespace argo-events
helm repo add argo https://argoproj.github.io/argo-helm
helm install --namespace argo-events --name argo-events argo/argo-events
```

create secret for git.

```bash
# If you using 2FA on bitbucket, you need set "App Password" to set to YOUR_BITBUCKET_PASSWORD.
kubectl -n argo-events create secret generic git-secret --from-literal=username=YOUR_BITBUCKET_USERNAME --from-literal=password=YOUR_BITBUCKET_PASSWORD
```

create argo-events resources.

```bash
kubectl -n argo-events apply -f webhook-event-source.yaml
kubectl -n argo-events apply -f webhook-gateway.yaml
kubectl -n argo-events apply -f webhook-sensor.yaml
```

yaml resources here. (using [examples/sensors/trigger-source-git.yaml](https://github.com/argoproj/argo-events/blob/master/examples/sensors/trigger-source-git.yaml) resouce name modified.)

```yaml
# webhook-event-source.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: webhook-event-source1
  labels:
    argo-events-event-source-version: v0.10
data:
  example: |-
    port: "12000"
    endpoint: "/example"
    method: "POST"
```

```yaml
# webhook-gateway.yaml
apiVersion: argoproj.io/v1alpha1
kind: Gateway
metadata:
  name: webhook-gateway1
  labels:
[...]

PR #1146, opened 2021-03-30
fix: Trim newline of the content read from Secrets or Configmaps

whynowy:

Secrets edited by tools like `vim` always have an invisible `\n` in the end, and it is often neglected by the user, however this makes differences for some of the applications. We trim that `\n` to avoid issues.

Also upgraded `go-git` package to the latest.

Fixes: #338

Might fix: #1068 as well.
Checklist:


alexec (approved):
are we saying go 16 is broken?

whynowy:
> are we saying go 16 is broken?



> are we saying go 16 is broken?

I take back what I said, it's NOT an issue of `ioutil.ReadFile()` or `os.ReadFile()`, this is actually an issue of `vim`. 

We always use `vim` to create a credential file on the local, and then use `kubectl create secret xxx --from-file=xxx ` to create k8s secrets. However, `vi` or `vim` always put newlines (`\n`) at the end of every line - especially including the last line. This `newline` is not a visible new line, but it does have an `\n` there. 

For example:

`vi test1.txt` with a string `hello` and `:wq`, you will see the file size is 6 instead of 5, the extra 1 is invisible `\n`, though you don't see a new line with `cat test1.txt`.

If you do the same thing with another file `text2.txt`, but do `:set binary`, `:set noeol` before `:wq`, the file size will be 5 indeed.

https://unix.stackexchange.com/questions/31807/what-does-the-noeol-indicator-at-the-bottom-of-a-vim-edit-session-mean

Here we still trim the ending `\n` in our code to avoid issues.

PR #1153, opened 2021-03-31
docs: document release cycle

whynowy:

Checklist:


docs/releases.md
```diff
@@ -0,0 +1,35 @@
...
+
+Image versions of `eventsource`, `sensor`, `eventbus-controller`,
+`eventsource-controller`, `sensor-controller` and `events-webhook` should be the
+same.
+
+## Release Cycle
+
+For **unstable**, we build and tag `latest` images for every commit to master.
+
+New major versions are released roughly every 6 months. Release candidates for
```
whynowy, line 28: @alexec - Do you think this is good?

alexec: * v1.2.0 = Jan 2021
* v1.1.0 = Nov 2020
* v1.0.0 = Sep 2020

That's about every 3 months?

alexec (approved):

docs/releases.md
whynowy: hmmm. Major means `v1.x.x`, `v2.x.x`, I can not tell how frequently it would be for major releases. Let's not make that statement but only do minor release (every 2 months), is that ok?

alexec (approved):

alexec:
@whynowy I think you should use your admin privilege (if you have it) to merge doc changes. Do create PRs. Only request a review if you think you need it.

Issue #985, opened 2020-12-07
Auto catch-up missing events when Resource type of EventSource is offline 


whynowy:
**Is your feature request related to a problem? Please describe.**
For `Resource` type of EventSource, a field named `afterStart` is needed to watch k8s resources especially for k8s `Event`, otherwise it will catch duplicate events. However, with this field set, the missing events during the POD offline will also not be able to be detected.

**Describe the solution you'd like**
TBD

**Describe alternatives you've considered**

**Additional context**

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

alexec:
Can we prioritize a fix to this because it is popular and we are exposed?

whynowy:
Addressed by making the event source HA.

PR #1158, opened 2021-04-05
feat: HA support for event sources and sensors

whynowy:

1. `Active-Passive` HA for some of event sources (the rest ones have been able to do `Active-Active`);
2. `Active-Passive` HA for sensors.
3. `spec.replicas` for Sensor spec.
4. `spec.replicas` for EventSource spec (deprecate `spec.replica`). 

Mitigate #985

Checklist:


eventsources/cmd/main.go
```diff
@@ -58,8 +59,10 @@ func main() {
 	ctx := logging.WithLogger(signals.SetupSignalHandler(), logger)
 	m := metrics.NewMetrics(eventSource.Namespace)
 	go m.Run(ctx, fmt.Sprintf(":%d", common.EventSourceMetricsPort))
+
+	logger.Infow("starting eventsource server", "version", argoevents.GetVersion())
 	adaptor := eventsources.NewEventSourceAdaptor(eventSource, busConfig, ebSubject, hostname, m)
 	if err := adaptor.Start(ctx); err != nil {
-		logger.Desugar().Fatal("failed to start eventsource server", zap.Error(err))
+		logger.Fatalw("failed to start eventsource server", zap.Error(err))
 	}
```
alexec, line 67: Maybe it is time to have a PR to de-desugar the whole codebase at once?

eventsources/eventing.go
```diff
@@ -263,20 +268,90 @@ func NewEventSourceAdaptor(eventSource *v1alpha1.EventSource, eventBusConfig *ev
...
 	servers := GetEventingServers(e.eventSource, e.metrics)
+	for k := range servers {
+		if _, ok := recreateTypes[k]; ok {
+			isRecreatType = true
+		}
+		// This is based on the presumption that all the events in one
+		// EventSource object use the same type of deployment strategy
+		break
+	}
+	if !isRecreatType || e.eventSource.Spec.GetReplicas() == 1 {
```
alexec, line 287: Minor - is workflows we always use leader election.  What happens when you scale from 1 to 2 and the first replica is not already using it? Don’t you get two leaders?

eventsources/eventing.go
```diff
@@ -263,20 +268,90 @@ func NewEventSourceAdaptor(eventSource *v1alpha1.EventSource, eventBusConfig *ev
...
+
+	leaseName := "eventsource-" + e.eventSource.Name
+	lock := &resourcelock.LeaseLock{
+		LeaseMeta: metav1.ObjectMeta{
+			Name:      leaseName,
+			Namespace: e.eventSource.Namespace,
+		},
+		Client: kubeClient.CoordinationV1(),
+		LockConfig: resourcelock.ResourceLockConfig{
+			Identity: e.hostname,
```
alexec, line 306: I think we also emit events on leadership change in workflows

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -111,6 +111,19 @@ type SensorSpec struct {
 	// Circuit is a boolean expression of dependency groups
 	// Deprecated: will be removed in v1.5, use Switch in triggers instead.
 	DeprecatedCircuit string `json:"circuit,omitempty" protobuf:"bytes,7,opt,name=circuit"`
+	// Replicas is the sensor deployment replicas
+	Replicas *int32 `json:"replicas,omitempty" protobuf:"varint,8,opt,name=replicas"`
+}
+
+func (s SensorSpec) GetReplicas() int32 {
+	if s.Replicas == nil {
```
alexec, line 119: Test?

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -109,6 +110,24 @@ type EventSourceSpec struct {
 	Pulsar map[string]PulsarEventSource `json:"pulsar,omitempty" protobuf:"bytes,27,opt,name=pulsar"`
 	// Generic event source
 	Generic map[string]GenericEventSource `json:"generic,omitempty" protobuf:"bytes,28,rep,name=generic"`
+	// Replicas is the event source deployment replicas
+	Replicas *int32 `json:"replicas,omitempty" protobuf:"varint,29,opt,name=replicas"`
+}
+
+func (e EventSourceSpec) GetReplicas() int32 {
+	if e.Replicas == nil && e.DeprecatedReplica == nil {
```
alexec, line 118: This is just complicated enough to warrant a unit test.

test/util/util.go
```diff
@@ -265,7 +284,24 @@ func SensorPodLogContains(ctx context.Context, kubeClient kubernetes.Interface,
 		return false, fmt.Errorf("error getting sensor pod name: %w", err)
 	}
 	podName := podList.Items[0].GetName()
-	fmt.Printf("Sensor POD name: %s", podName)
+	if len(podList.Items) > 1 {
```
alexec, line 287: I think you can simplify the code by removing this guard and the outcome should be the same?

test/util/util.go
```diff
@@ -252,6 +254,23 @@ func EventSourcePodLogContains(ctx context.Context, kubeClient kubernetes.Interf
 		return false, fmt.Errorf("error getting event source pod name: %w", err)
 	}
 	podName := podList.Items[0].GetName()
+	if len(podList.Items) > 1 {
```
alexec, line 257: I think you can simplify the code by removing this guard condition.

alexec (commented):
@sarabala1979 should review too? How did you test this?

eventsources/eventing.go
sarabala1979: I would also recommend using leader election by default which will avoid the code complexity to deal separately. Workflow enables election leader by default

sarabala1979:
> @sarabala1979 should review too? How did you test this?
I tested locally switching the pods (killing current leader) and verify the logs for leader switching.

pkg/apis/sensor/v1alpha1/types.go
whynowy: Test cases added.

pkg/apis/eventsource/v1alpha1/types.go
whynowy: Test cases added.

eventsources/eventing.go
alexec: good point

eventsources/eventing.go
alexec: Ah - so users must all install the extra RBAC - a process that does not scale. Good point.

alexec (approved):

PR #1161, opened 2021-04-05
chore: fix e2e ci build on release branch

whynowy:

On release branch, the image tags in the manifests are not `latest` but the real version like `v1.3.0`, this makes `make start` fail like https://github.com/argoproj/argo-events/runs/2272839939?check_suite_focus=true. This PR is used to fix this issue.

Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ GIT_TREE_STATE=$(shell if [ -z "`git status --porcelain`" ]; then echo "clean" ;
 DOCKER_PUSH?=false
 IMAGE_NAMESPACE?=argoproj
 VERSION?=latest
-BASE_VERSION?=latest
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,6 +14,7 @@ GIT_TREE_STATE=$(shell if [ -z "`git status --porcelain`" ]; then echo "clean" ;
 DOCKER_PUSH?=false
 IMAGE_NAMESPACE?=argoproj
 VERSION?=latest
+BASE_VERSION:=latest
```
whynowy, line 17: To clarify:

`BASE_VERSION` is the original image tag in the manifests.
`VERSION` is the new image tag that could be passed in during image build.

alexec (approved):

PR #1163, opened 2021-04-07
feat: EventSource and Sensor HA without extra RBAC

whynowy:

A new approach to do `Active-Passive` HA for Sensors and some of the EventSources, which does not require configuring extra RBAC. It utilizes `raft` leader election algorithm implemented over NATS.

See https://github.com/nats-io/graft.

Checklist:


controllers/eventsource/resource.go
```diff
@@ -332,26 +335,6 @@ func buildDeploymentSpec(args *AdaptorArgs) (*appv1.DeploymentSpec, error) {
 		spec.Template.Spec.PriorityClassName = args.EventSource.Spec.Template.PriorityClassName
 		spec.Template.Spec.Priority = args.EventSource.Spec.Template.Priority
 	}
-	allEventTypes := eventsources.GetEventingServers(args.EventSource, nil)
```
whynowy, line 335: We don't need this any more, with leader election, all the event source deployments can run with `rolling update` strategy.

controllers/sensor/resource.go
```diff
@@ -270,10 +275,6 @@ func buildDeploymentSpec(args *AdaptorArgs) (*appv1.DeploymentSpec, error) {
 			MatchLabels: args.Labels,
 		},
 		Replicas: &replicas,
-		Strategy: appv1.DeploymentStrategy{
```
whynowy, line 273: With leader election, sensor deployments can also run with `rolling update` strategy.

eventsources/eventing.go
```diff
@@ -284,54 +280,24 @@ func (e *EventSourceAdaptor) Start(ctx context.Context) error {
 		// EventSource object use the same type of deployment strategy
 		break
 	}
-	if !isRecreatType || e.eventSource.Spec.GetReplicas() == 1 {
+	if !isRecreatType {
```
whynowy, line 283: Event sources like `webhook` should not run with leader election.

test/util/util.go
```diff
@@ -283,28 +261,41 @@ func SensorPodLogContains(ctx context.Context, kubeClient kubernetes.Interface,
...
-	if len(podList.Items) > 1 {
-		// HA
-		var l *coordinationv1.Lease
-		for {
-			l, err = kubeClient.CoordinationV1().Leases(namespace).Get(ctx, "sensor-"+sensorName, metav1.GetOptions{})
+
+	return PodsLogContains(ctx, kubeClient, namespace, regex, podList, timeout), nil
+}
+
+func PodsLogContains(ctx context.Context, kubeClient kubernetes.Interface, namespace, regex string, podList *corev1.PodList, timeout time.Duration) bool {
```
whynowy, line 268: Watch logs from multiple Pods.

alexec:
I'm in two minds about this.

Leadership-election is a core cloud-native feature that is well understood by the Argo Team and community in general. We're replacing it with something that makes us more reliant on NATS. Given that a user needs to make changes to use the feature either way, would it be better to make using leadership elections straight forward?

whynowy:
> I'm in two minds about this.
> 
> Leadership-election is a core cloud-native feature that is well understood by the Argo Team and community in general. We're replacing it with something that makes us more reliant on NATS. Given that a user needs to make changes to use the feature either way, would it be better to make using leadership elections straight forward?



> I'm in two minds about this.
> 
> Leadership-election is a core cloud-native feature that is well understood by the Argo Team and community in general. We're replacing it with something that makes us more reliant on NATS. Given that a user needs to make changes to use the feature either way, would it be better to make using leadership elections straight forward?

If we don't have NATS existing in the architecture, or the HA is for controllers, there's no doubt k8s leader election is the first choice. However the HA is for a dynamic service (EventSource or Sensor), and it so happened NATS is already there, I think we should utilize it, because that benefits much - it saves lots of extra configuration.

This approach requires the minimal spec change - specifying `spec.replicas`, however using k8s leader elections, besides the `replicas`, user need to make series of RBAC change:

1. Service Account
2. Role and RoleBinding
3. Specify `spec.template.serviceAccountName`

And these extra changes are required at least per namespace, if the user has a strong sense of security, he probably will specify `resourceName` in the `Role` definition (which limits the service account used by the EventSource/Sensor only can operate on the `Lease` object dedicated for it), this makes the RBAC settings even per EventSource (or Sensor).

I understand the concern of reliability of using leader election implemented over NATS, even I have done all kinds of testing upon it without seeing any issue, I still can not say it's reliable enough. How about this, we can defer using leader election when `replicas = 1`, and let's see the feedback from the community, and then make a decision if we use it by default (even when `replicas=1`)?

@alexec

alexec:
I think we should go with NATS now. It's a big ask on our users to add a lot of extra RBAC, which people struggle with.

common/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,158 @@
...
+		default:
+			log.Fatalf("Unknown state: %s", sc.To)
+		}
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info("exiting...")
+			cancel()
```
alexec, line 150: normally defer a cancel?

controllers/sensor/resource.go
```diff
@@ -185,9 +185,14 @@ func buildDeployment(args *AdaptorArgs, eventBus *eventbusv1alpha1.EventBus) (*a
 					},
 				},
 			})
+			emptyDirVolName := "tmp-volume"
```
alexec, line 188: minor - maybe just call this `tmp`

alexec (approved):

common/leaderelection/leaderelection.go
whynowy: When current node is changed from `leader` to non-leader (line 128-133), `cancel()` need to be called to terminate the running service, and re-initiate a cctx and cancel. Not quite sure if `defer cancel()` still works in that case, let me do more testing.

controllers/sensor/resource.go
whynowy: done.

common/leaderelection/leaderelection.go
whynowy: Using defer works, updated.

PR #1170, opened 2021-04-08
Allow Argo Rollout manipulation from RBAC

davidkulak:
Hey, 

It would be great to allow Argo Rollout manipulation from Argo Events.

Regards.

### DIFF
--- a/manifests/cluster-install/rbac/argo-events-cluster-role.yaml
+++ b/manifests/cluster-install/rbac/argo-events-cluster-role.yaml
@@ -46,6 +46,7 @@ rules:
       - eventbus
       - eventbus/finalizers
       - eventbus/status
+      - rollouts
   - apiGroups:
       - ""
     resources:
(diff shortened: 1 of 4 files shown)


whynowy (commented):
Thanks for raising this PR!

This PR actually reminds me this is the wrong direction - `argo-events-sa` is only supposed to be used by the controllers, and it does not need to have the privileges to access `workflows` and `rollouts`, we should clean up extra privileges from the RBAC. We have an instruction on how to properly manage the RBAC for the service accounts used by EventSource and Sensor, see [here](https://github.com/argoproj/argo-events/blob/master/docs/service-accounts.md).

Currently in the examples, we used `argo-events-sa` for demonstration, this is not the right approach we encourage the users to do, I will make changes to the examples.

whynowy:
https://github.com/argoproj/argo-events/pull/1175

whynowy:
Updated the examples to use another dedicated service account, and that should be the right approach to run EventSources and Sensors in user namespaces.

PR #1174, opened 2021-04-08
fix(docs): Update link in 02-parameterization.md

niekoost:
The link to the first page of the tutorial was wrong. An underscore instead of a dash after 01. 
This change repairs that error.

Checklist:


whynowy (approved):
Thanks!

PR #1186, opened 2021-04-22
feat: kafka sasl auth

### LINKED ISSUE #703: SASL authentication to Kafka
I may be doing something wrong, but I can't seem to find out how to make a SASL connection to a Kafka Event Source. 

It would be nice to have SASL authentication to a Kafka Event Source and Event Trigger, besides a TLS connection. In particular, I am using SASL_SSL and Plain text as SASL mechanism.

I am not a TLS expert, so I am not aware of a TLS-type solution to authenticate via SASL.

**Additional context**
Not applicable.

joshuajorel:
Checklist:

* [X] Fixes #703

Test Cases:
* [X] Test Invalid Credentials (EventSource and Trigger)
* [X] Test Invalid SASL Mechanism (EventSource and Trigger)
* [X] Test SASL PLAIN Mechanism

Notes:

- sarama library says that it only supports [SASL/PLAIN](https://github.com/Shopify/sarama/blob/a7eb541855c8e19965a30547eadc20f6bc6edea5/config.go#L60), but has different mechanisms available in its [codebase](https://github.com/Shopify/sarama/blob/a7eb541855c8e19965a30547eadc20f6bc6edea5/broker.go#L57). I added the other authentication mechanisms just in case
- This PR was only tested with the SASL PLAIN mechanism


pkg/apis/common/validate.go
```diff
@@ -30,3 +30,39 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
...
 }
+
+func ValidateSASLConfig(saslConfig *SASLConfig) error {
+	if saslConfig == nil {
+		return nil
+	}
+
+	var mechanismSet, userSet, passwordSet bool
+
+	if saslConfig.Mechanism != "" {
```
whynowy, line 41: Does not need `!=""` check

pkg/apis/common/validate.go
```diff
@@ -30,3 +30,39 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
...
+	if saslConfig.Password != nil {
+		passwordSet = true
+	}
+
+	if !userSet && !passwordSet {
+		return errors.New("invalid sasl config, please configure either User, and/or Password")
+	}
+
+	if !mechanismSet {
+		return errors.New("invalid sasl config, please configure Mechanism. Possible values are `OAUTHBEARER`, `PLAIN`, `SCRAM-SHA-256`, `SCRAM-SHA-512` and `GSSAPI`")
```
whynowy, line 64: It says `defaults to PLAIN` above.

pkg/apis/common/validate.go
```diff
@@ -30,3 +30,39 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
...
+
+	if saslConfig.User != nil {
+		userSet = true
+	}
+
+	if saslConfig.Password != nil {
+		passwordSet = true
+	}
+
+	if !userSet && !passwordSet {
```
whynowy, line 59: Can either of them be missing?

sensors/triggers/kafka/kafka.go
```diff
@@ -63,6 +63,24 @@ func NewKafkaTrigger(sensor *v1alpha1.Sensor, trigger *v1alpha1.Trigger, kafkaPr
 			config.Version = version
 		}
 
+		if kafkatrigger.SASL != nil {
+			config.Net.SASL.Enable = true
+			config.Net.SASL.Mechanism = sarama.SASLMechanism(kafkatrigger.SASL.Mechanism)
```
whynowy, line 68: Validating?

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -461,28 +461,31 @@ type KafkaEventSource struct {
 	Topic string `json:"topic" protobuf:"bytes,3,opt,name=topic"`
 	// Backoff holds parameters applied to connection.
 	ConnectionBackoff *apicommon.Backoff `json:"connectionBackoff,omitempty" protobuf:"bytes,4,opt,name=connectionBackoff"`
+	// SASL configuration for the kafka client
```
whynowy, line 464: BTW - we should keep the order of existing fields, otherwise it's a breaking change for proto files, we should reduce that kind of changes.

https://developers.google.com/protocol-buffers/docs/pythontutorial#extending-a-protocol-buffer

pkg/apis/common/validate.go
joshuajorel: This should have been an OR statement. Will fix.

sensors/triggers/kafka/kafka.go
joshuajorel: Checks whether the configuration is set. Enables SASL authentication if parameters are set in the YAML configs.

pkg/apis/common/common.go
```diff
@@ -128,6 +128,21 @@ type TLSConfig struct {
...
+	// Possible values: OAUTHBEARER, PLAIN (defaults to PLAIN).
+	Mechanism string `json:"mechanism,omitempty" protobuf:"bytes,1,opt,name=mechanism"`
+	// Version is the SASL Protocol Version to use
+	// Kafka > 1.x should use V1, except on Azure EventHub which use V0
+	// User is the authentication identity (authcid) to present for
+	// SASL/PLAIN or SASL/SCRAM authentication
+	User *corev1.SecretKeySelector `json:"user,omitempty" protobuf:"bytes,2,opt,name=user"`
+	// Password for SASL/PLAIN authentication
+	Password *corev1.SecretKeySelector `json:"password,omitempty" protobuf:"bytes,3,opt,name=password"`
+	// authz id used for SASL/SCRAM authentication
```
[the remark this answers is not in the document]
joshuajorel: Remove authz id. It wasn't supposed to be in there. Comments were lifted from the sarama library.

pkg/apis/common/common.go
```diff
@@ -128,6 +128,21 @@ type TLSConfig struct {
 	DeprecatedClientKeyPath string `json:"clientKeyPath,omitempty" protobuf:"bytes,6,opt,name=clientKeyPath"`
 }
 
+// SASLConfig refers to SASL configuration for a client
+type SASLConfig struct {
+	// SASLMechanism is the name of the enabled SASL mechanism.
+	// Possible values: OAUTHBEARER, PLAIN (defaults to PLAIN).
+	Mechanism string `json:"mechanism,omitempty" protobuf:"bytes,1,opt,name=mechanism"`
+	// Version is the SASL Protocol Version to use
```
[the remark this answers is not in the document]
joshuajorel: Removed comments.

pkg/apis/common/validate.go
joshuajorel: Removed `!=""` check

pkg/apis/common/validate.go
joshuajorel: Renamed error message to account for the only possible values.

sensors/triggers/kafka/kafka.go
joshuajorel: Added additional validation for empty strings.

pkg/apis/eventsource/v1alpha1/types.go
joshuajorel: Reverted the byte ordering and added the SASL configs to the end.

eventsources/sources/kafka/start.go
```diff
@@ -270,6 +270,30 @@ func getSaramaConfig(kafkaEventSource *v1alpha1.KafkaEventSource, log *zap.Sugar
 		config.Version = version
 	}
 
+	if kafkaEventSource.SASL != nil {
+		config.Net.SASL.Enable = true
+
+		if kafkaEventSource.SASL.Mechanism == "" {
```
whynowy, line 276: A function `func (s SASLConfig) GetMechanism()` would be helpful to save `if..else` everywhere.

joshuajorel: Will refactor.

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -499,20 +499,24 @@ type KafkaTrigger struct {
 	// Defaults to 500 milliseconds.
 	// +optional
 	FlushFrequency int32 `json:"flushFrequency,omitempty" protobuf:"varint,7,opt,name=flushFrequency"`
+	// SASL configuration for the kafka client
```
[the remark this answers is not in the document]
joshuajorel: Reverted order.

pkg/apis/common/validate.go
```diff
@@ -30,3 +30,36 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
 	}
 	return nil
 }
+
+func ValidateSASLConfig(saslConfig *SASLConfig) error {
+	if saslConfig == nil {
+		return nil
+	}
+
+	var mechanismSet, userSet, passwordSet bool
```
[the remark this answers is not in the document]
joshuajorel: Removed xxxSet flags. merged validating user and password into a single if statemetn.

eventsources/sources/kafka/start.go
joshuajorel: Refactored. By default, will set to "PLAINTEXT" unless otherwise specified.

pkg/apis/common/common.go
```diff
@@ -153,3 +167,13 @@ type Metadata struct {
 	Annotations map[string]string `json:"annotations,omitempty" protobuf:"bytes,1,rep,name=annotations"`
 	Labels      map[string]string `json:"labels,omitempty" protobuf:"bytes,2,rep,name=labels"`
 }
+
+func (s SASLConfig) GetMechanism() sarama.SASLMechanism {
```
whynowy, line 171: Overall everything looks good to me except this part.

1. Returning `sarama.SASLMechanism` makes `github.com/Shopify/sarama` get introduced to the models, which is not good. Ideally the dependencies in models are either from native GoLang or k8s related. I prefer to return a string here.
2. Another reason is, since this struct is put in `common`, it is intended to be used by not only Kafka. Assume it's going to be used by something else, `sarma.SASLMechanism` might not be recognized by it. Also, I made a little research on SASL auth, it looks like the mechanism is not unified, different product has different support (PLAIN might be the one everyone supports). Correct me if I'm wrong.

joshuajorel: 1. Apologies for the oversight. Removed the type reference and refactored again to reflect on the actual sensor and eventsource. 
2. I'm basing the authentication mechanism on the available library which is Sarama. The other authentication mechanisms might be something to look into in the future, but "PLAIN" auth is a good place to start.

whynowy (approved):
LGTM, thanks!

Issue #994, opened 2020-12-17
InvocationType for AWS Lambda trigger


mrkwtz:
**Is your feature request related to a problem? Please describe.**
I want to make use of AWS Lambdas [DeadLetterQueue feature](https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html#dlq) for errors while processing async Lambda invocations. But right now it's [not possible](https://github.com/argoproj/argo-events/blob/master/sensors/triggers/aws-lambda/aws-lambda.go#L108-L110) to do a async invocation while [it could be](https://github.com/aws/aws-sdk-go/blob/v1.30.7/service/lambda/api.go#L9455).

**Describe the solution you'd like**
I'd like to have a InvocationType parameter in the Lambda trigger. 

**Describe alternatives you've considered**
For this case it's all about error handling. When I execute a Lambda function and it fails I want to be able to act on it. It would be awesome if argo-events itself has this feature, but for now I don't see a way to handle failed triggers within argo-events.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

alexec:
Is this the case of adding an additional invoke input? Would you like to submit a PR?

VaibhavPage:
@mrkwtz can you explain your use-case when you say "acting on failed lambda"?

mrkwtz:
@VaibhavPage, sure. When you configure a DeadLetterQueue (DLQ) in a Lambda function, failed invocations go straight to this queue from where another process could pick them up and report them or try the same event again but maybe with a more resilient and failure tolerant second function. Unfortunately DLQs are only considered when you do an asynchronous invocation of a Lambda function. So the Lambda trigger has to be extended to support the additional parameter.

PR #1187, opened 2021-04-22
feat(lambda-trigger): expose Lambda invocation type. Closes: #994

whynowy:

Checklist:


pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -435,13 +435,25 @@ type AWSLambdaTrigger struct {
...
+	//    The API response includes the function response and additional data.
+	//
+	//    * Event - Invoke the function asynchronously. Send events that fail multiple
+	//    times to the function's dead-letter queue (if it's configured). The API
+	//    response only includes a status code.
+	//
+	//    * DryRun - Validate parameter values and verify that the user or role
+	//    has permission to invoke the function.
+	// +optional
+	InvocationType *string `json:"invocationType,omitempty" protobuf:"bytes,7,opt,name=invocationType"`
```
alexec, line 456: in empty string valid? if not, change to non-pointer type

alexec (approved):

pkg/apis/sensor/v1alpha1/types.go
whynowy: No, it's not. 

The AWS API uses `*string`, so the most convenient way is to use the same type.

alexec: I don't feel strongly.

whynowy: Why?

Precisely, we are not sure if empty string is valid or not, that depends on AWS server side implementation. But we do know `nil` is valid as default value.

So if we use `string`, we should do something as below:

```
        input := &lambda.InvokeInput{
		FunctionName:   &trigger.FunctionName,
		Payload:        payload,
	}
        if trigger.InvocationType != "" {
                input.InvocationType = &trigger.InvocationType
        }
```

But if we use `*string`, we only need to do:

```
        input := &lambda.InvokeInput{
		FunctionName:   &trigger.FunctionName,
		Payload:        payload,
		InvocationType: trigger.InvocationType,
	}
```

alexec: I read your second comment - and misunderstood - sounds like it needs to be *string

PR #1188, opened 2021-04-23
fix: file event source

cbishopvelti:


justinmchase (approved):

whynowy:
@justinmchase  - you also see the `/` makes a difference?

justinmchase:
I didn't test it out but it makes sense.

PR #1201, opened 2021-05-05
fix: Retry duration recognizes strings. Fixes #1200

### LINKED ISSUE #1200: retryStrategy duration cannot be a string or else it defaults to 0
When a string representation of a duration is used on the retryStrategy.Duration field it defaults to 0.

The following yaml will not work since https://github.com/argoproj/argo-events/blob/master/common/retry.go#L56 looks at the Int64Value() only for the duration field

```yaml
      retryStrategy:
        steps: 10
        duration: "2s"
```

This workaround can be used in the interim

```yaml
      retryStrategy:
        steps: 10
        duration: 2000000000
```

**To Reproduce**
Created two tests in retry_test.go
```go
func TestConnectDurationBug(t *testing.T) {

	start := time.Now()
	count := 4
	err := Connect(nil, func() error {
		if count == 0 {
			return nil
		} else {
			count--
			return fmt.Errorf("new error")
		}
	})
	end := time.Now()
	elapsed := end.Sub(start)
	assert.NoError(t, err)
	assert.Equal(t, 0, count)
	assert.True(t, elapsed >= 4 * time.Second) //Will fail as the duration defaults to 0
}

func TestConnectRetry(t *testing.T) {

	factor   := apicommon.NewAmount("1.0")
	jitter   := apicommon.NewAmount("1")
	duration := apicommon.FromInt64(1000000000) //This must be set as int64
	backoff := apicommon.Backoff{
		Duration: &duration,
		Factor:   &factor,
		Jitter:   &jitter,
		Steps:    5,
	}
	count := 4
	start := time.Now()
	err := Connect(&backoff, func() error {
		if count == 0 {
			return nil
		} else {
			count--
			return fmt.Errorf("new error")
		}
	})
	end := time.Now()
	elapsed := end.Sub(start)
	assert.NoError(t, err)
	assert.Equal(t, 0, count)
	assert.True(t, elapsed >= 4 * time.Second) //Will pass
}
```

**Expected behavior**
retries will backoff using a duration that is not 0

 - Kubernetes: v1.19.7
 - Argo Events: [e.g. v1.3.0]

whynowy:

Checklist:

### DIFF
--- a/common/retry.go
+++ b/common/retry.go
@@ -30,7 +30,7 @@ import (
 var (
 	defaultFactor   = apicommon.NewAmount("1.0")
 	defaultJitter   = apicommon.NewAmount("1")
-	defaultDuration = apicommon.FromString("5s")
+	defaultDuration = apicommon.FromString("1s")
 
 	DefaultBackoff = apicommon.Backoff{
 		Steps:    5,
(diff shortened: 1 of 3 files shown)


common/retry.go
```diff
@@ -53,10 +53,18 @@ func IsRetryableKubeAPIError(err error) bool {
...
-		result.Duration = time.Duration(backoff.Duration.Int64Value())
+	d := backoff.Duration
+	if d == nil {
+		d = &defaultDuration
+	}
+	if d.Type == apicommon.Int64 {
+		result.Duration = time.Duration(d.Int64Value())
 	} else {
-		result.Duration = time.Duration(defaultDuration.Int64Value())
+		parsedDuration, err := time.ParseDuration(d.StrVal)
```
alexec, line 63: suggestion - https://github.com/argoproj/pkg/blob/master/env/env.go

alexec (approved):

common/retry.go
whynowy: It does not lookup env here, but that's a good toolkit.

PR #1202, opened 2021-05-06
chore: build sensor image from scratch

whynowy:

Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,23 +45,33 @@ COPY dist/eventsource /bin/eventsource
 ENTRYPOINT [ "/bin/eventsource" ]
 
 ####################################################################################################
-# sensor
+# sensor-base
 ####################################################################################################
-FROM alpine:3.12.3 as sensor
+FROM alpine:3.12.3 as sensor-base
 RUN apk update && apk upgrade && \
     apk add --no-cache git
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -45,26 +45,28 @@ COPY dist/eventsource /bin/eventsource
...
 RUN wget -q https://github.com/argoproj/argo/releases/download/${ARGO_VERSION}/argo-linux-amd64.gz
 RUN gunzip argo-linux-amd64.gz
 RUN chmod +x argo-linux-amd64
 RUN mv ./argo-linux-amd64 /usr/local/bin/argo
 RUN argo version
 
+####################################################################################################
+# sensor
+####################################################################################################
+FROM scratch as sensor
```
alexec, line 65: Good for security.

Suggestion - if you're not confident about what the base image might need - try the distroless images for your bass. They are nearly as secure as scratch - but also have `/tmp` directory and other stuff. I've found that sometime you need this stuff, but it is hard to know if you'lle need it - just some days later you find a weird bug..  I find them more reliable and probably would choose them instead of scratch in the future:

https://github.com/GoogleContainerTools/distroless/blob/main/examples/go/Dockerfile

alexec (approved):

PR #1204, opened 2021-05-07
docs: Enriched github example

caueasantos:
Straightforward example showing `argo-events` power and how to interact with `Github` attributes


whynowy (approved):
Thanks!

PR #1207, opened 2021-05-07
fix: limit slack API calls

### LINKED ISSUE #1206: Slack trigger: Workspaces with large number of channels can trigger API rate limit
**Describe the bug**
We have a workspace with thousands of channels, many of which are archived. As the Slack API (inexplicably) provides no means of looking up a channel ID by name, the Slack trigger code attempts to iterate through paginated results of the `conversations.list` API method.

The iteration also does not stop when the channel is found, so the error always occurs in sufficiently large workspaces even when the channel ID is discovered prior to the rate limit enforcement.

This lookup is performed for each execution of the trigger, further compounding the problem.

**To Reproduce**
1. Create a slack trigger with a token attached to a workspace containing many channels (>1000 or more may be needed)
2. Send an event that triggers the sensor
3. Observe slack sensor pod logs for error

**Expected behavior**
Slack messages are delivered to channel reliably without excess API calls or rate limiting errors.

**Logs**

`
{"level":"error","ts":1620218197.853904,"logger":"argo-events.sensor","caller":"sensors/listener.go:271","msg":"failed to execute a trigger","sensorName":"zi-argo-slack-alerts","error":"failed to execute trigger: timed out waiting for the condition: failed to list channels: slack rate limit exceeded, retry after 3s","errorVerbose":"timed out waiting for the condition: failed to list channels: slack rate limit exceeded, retry after 3s\nfailed to execute trigger\ngithub.com/argoproj/argo-events/sensors.(*SensorContext).triggerOne\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:328\ngithub.com/argoproj/argo-events/sensors.(*SensorContext).triggerActions\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:269\ngithub.com/argoproj/argo-events/sensors.(*SensorContext).listenEvents.func1.3\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:181\nruntime.goexit\n\t/opt/hostedtoolcache/go/1.14.15/x64/src/runtime/asm_amd64.s:1373","triggerName":"slack-trigger","triggeredBy":["state-transition"],"triggeredByEvents":["36653033356137652d643161652d346365372d626338382d346339343864633632316438"],"stacktrace":"github.com/argoproj/argo-events/sensors.(*SensorContext).triggerActions\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:271\ngithub.com/argoproj/argo-events/sensors.(*SensorContext).listenEvents.func1.3\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:181"}
`

 - Argo Events: v1.3.1

---

tinkerborg:
- only attempt to list/join channels if postMessage call fails with
  not_in_channel
- respect rate limit 429 responses when iterating through paginated
  conversations.list result

fixes #1206

Checklist:


whynowy:
@shashwat-appdirect @AleksanderGondek - Could you review this PR since both of you made code change on it? I think you have better knowledge than me.

tinkerborg:
I'm also fairly new to Go, so definitely open to suggestions for a better solution for the rate limiting here. ISTM that something like [go-retryablehttp](https://github.com/hashicorp/go-retryablehttp) would implement it more generically for all the API methods here, but it wasn't clear for me, as a neophyte, how to utilize it here.

Apart from that, the other thing this PR does is implement a "try first, then ask for forgiveness" logic -- i.e., try to send the message and then try to join the channel only if that fails. The reasoning here is that the Slack API makes discovering a channel ID given its name for more complicated and expensive than it should be. Since the postMessage call doesn't require an ID, the ID is only needed during a join, so this logic defers that discovery until it's actually needed. Referring directly to the error text to determine this seems a bit brittle here, but I couldn't readily see another way to implement the logic.

shashwat-appdirect (approved):
The flow LGTM. Have you tested it both for public channel and private channel?

whynowy:
@shashwat-appdirect - could you help do a testing on the private channel stuff if @tinkerborg provides a testing image?

shashwat-appdirect:
> > The flow LGTM. Have you tested it both for public channel and private channel?
> 
> I did try this with a private channel, but it's not clear to me why/how the original code supported this. The private channel was not visible with my bot token and did not appear in the `conversations.list` output. The `chat.postMessage` call returns a `channel_not_found` error when trying to post to a private channel, rather than `not_in_channel`. What is the expected behavior here?

@tinkerborg you need to have proper permissions for your bot - channels:join, channels:read, chat:write, groups:read. Also, the slack app should be manually added to the private channel. Bot does not have permission to join a private channel.

shashwat-appdirect:
> @shashwat-appdirect - could you help do a testing on the private channel stuff if @tinkerborg provides a testing image?

Hi @whynowy , I think it will be easy for @tinkerborg to create a sensor image and test locally, since he will be having a setup. But still I could try to test.

tinkerborg:
I can test tomorrow morning.

shashwat-appdirect:
I am sure earlier `conversations.list` returned both public and private channels. To get private channel you need to pass `private_channel` in the field `types`, while calling `conversations.list` method. I tested just now using Slack API Tester - https://api.slack.com/methods/conversations.list/test and able to list all public and private channels.

tinkerborg:
> I am sure earlier `conversations.list` returned both public and private channels. To get private channel you need to pass `private_channel` in the field `types`, while calling `conversations.list` method. I tested just now using Slack API Tester - https://api.slack.com/methods/conversations.list/test and able to list all public and private channels.

That was my understanding as well, otherwise why would the option even exist? I'll do some digging and see if I can figure out what's going on.

tinkerborg:
@shashwat-appdirect Using my bot's token in the API tester, with `private_channel` argument, I can list private channels, but only the ones to which my bot has already been invited. So it's still not clear to me what the `isPrivateChannel` logic was doing.

shashwat-appdirect:
> @shashwat-appdirect Using my bot's token in the API tester, with `private_channel` argument, I can list private channels, but only the ones to which my bot has already been invited. So it's still not clear to me what the `isPrivateChannel` logic was doing.

`conversations.join` is not allowed for a private channel. Earlier if the channel was found in `conversations.list`, the `conversations.join` method was still called. But if the channel is private, this method fails with error - `method_not_supported_for_channel_type`. Hence, to avoid calling this method on private channels, `isPrivateChannel` check was used.
You could go through the old code flow, if still it is not clear :)

tinkerborg:
> > @shashwat-appdirect Using my bot's token in the API tester, with `private_channel` argument, I can list private channels, but only the ones to which my bot has already been invited. So it's still not clear to me what the `isPrivateChannel` logic was doing.
> 
> `conversations.join` is not allowed for a private channel. Earlier if the channel was found in `conversations.list`, the `conversations.join` method was still called. But if the channel is private, this method fails with error - `method_not_supported_for_channel_type`. Hence, to avoid calling this method on private channels, `isPrivateChannel` check was used.
> You could go through the old code flow, if still it is not clear :)

I didn't change this flow, just moved it to after trying `chat.postMessage`, and added the rate limiting. I don't see that it is doing anything though, since `conversations.list` only lists private channels if the bot is already joined. If you have a test case that the previous PR was addressing, I can try to reproduce and test with the code in this PR.

shashwat-appdirect:
Not any particular test case. I only tested if the message is posted in private channel. In your case you told that it is working after your code changes, so it should be fine. Don't think if anymore scenarios required.

sensors/triggers/slack/slack.go
```diff
@@ -114,54 +115,65 @@ func (t *SlackTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
...
+						}
+					}
+					for _, c := range channels {
+						if c.Name == channel {
+							channelID = c.ID
+							isPrivateChannel = c.IsPrivate
+							break
+						}
+					}
+					if nextCursor == "" || channelID != "" {
```
whynowy, line 149: 👍

sensors/triggers/slack/slack.go
```diff
@@ -114,54 +115,65 @@ func (t *SlackTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
...
+						break
+					}
+					params.Cursor = nextCursor
+				}
+				if channelID == "" {
+					return nil, errors.Errorf("failed to get channelID of %s", channel)
+				}
+				if isPrivateChannel {
```
whynowy, line 157: I actually not very sure what the logic for private channel should be, but here the logic seems to be changed.

Old: Skip "join conversation", still post message;
New: return an error.

sensors/triggers/slack/slack.go
```diff
@@ -114,54 +115,65 @@ func (t *SlackTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
...
-	params := &slack.GetConversationsParameters{
-		Limit: 200,
-		Types: []string{"public_channel", "private_channel"},
-	}
-	channelID := ""
-	isPrivateChannel := false
+	t.Logger.Infow("posting to channel...", zap.Any("channelName", channel))
 	for {
-		channels, nextCursor, err := api.GetConversations(params)
+		channelID, timestamp, err := api.PostMessage(channel, slack.MsgOptionText(message, false))
```
whynowy, line 120: With the code change, do we still need the for loop in line 119? I seems to be infinite.

sensors/triggers/slack/slack.go
tinkerborg: The error probably makes sense IMO, because if it's a private channel the bot can't join and can't send a message to it without joining. It needs to be invited.

sensors/triggers/slack/slack.go
tinkerborg: This is to implement retry logic, i.e. try to post message first, then if it fails due to not_in_channel error, join the channel and try again.

It shouldn't be infinite, as all paths of logic inside the loop lead to a return, either due to success or error. If the `api.JoinConversation` call on line 161 incorrectly returns a `nil` value for `err` when it actually did fail under the hood, this could loop infinitely. Not sure how likely that is, but I could limit the loop to max 2 iterations and add some sort of "unspecified error" logging after it.

whynowy: I see.

sensors/triggers/slack/slack.go
whynowy: Sounds good.

whynowy (approved):
LGTM, thanks!

PR #1209, opened 2021-05-08
docs(argo-events): Added helpful information to FAQ

alecrajeev:
I think it would be helpful to add this to the FAQ guide. The eventSourceName and eventName can be confusing at first.


whynowy (approved):
Thanks!

Issue #1190, opened 2021-04-26
[FEATURE REQUEST] Multi Arch support


ionutz89:
It will be great if we can have images compiled as multi arch like `arm64.`

A use case can be used on RPi 4.

whynowy:
https://github.com/argoproj/argo-events/pull/1218

PR #1213, opened 2021-05-13
chore: Dockerfile compatible with all Architecture builds

mickkael:
Signed-off-by: mickkael <19755421+mickkael@users.noreply.github.com>

Compiled successfully arm64 with this patched Dockerfile
https://hub.docker.com/r/mickkael/argo-events/tags

Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,19 @@
+ARG ARCH=amd64
 ####################################################################################################
 # base
 ####################################################################################################
 FROM alpine:3.12.3 as base
+ARG ARCH
 RUN apk update && apk upgrade && \
     apk add ca-certificates && \
     apk --no-cache add tzdata
 
 ENV ARGO_VERSION=v3.0.2
 
-RUN wget -q https://github.com/argoproj/argo/releases/download/${ARGO_VERSION}/argo-linux-amd64.gz
-RUN gunzip argo-linux-amd64.gz
-RUN chmod +x argo-linux-amd64
-RUN mv ./argo-linux-amd64 /usr/local/bin/argo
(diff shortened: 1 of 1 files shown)


whynowy:
Thanks for your PR!

Instead of specifying the ARCH, I would expect to see it support multi-arch build automatically (#1190). Please let me know if you would like to take it.

mickkael:
I'll work on the pipeline to make a multi-arch container. I do it already locally with an argo workflow and steps for arm64/arm/amd64, and mplatform/manifest-tool to have the multi-arch manifest. But we need this Dockerfile anyway as a starting point for each architecture build

whynowy:
> I'll work on the pipeline to make a multi-arch container. I do it already locally with an argo workflow and steps for arm64/arm/amd64, and mplatform/manifest-tool to have the multi-arch manifest. But we need this Dockerfile anyway as a starting point for each architecture build.
> 
> Hey @whynowy , what CI is used to build and push argo-events ? I can't find anything in the repo.

We used to support multi-arch build before, but there were some issues that pushing a large multi-arch image took too long and it often made the connection to image registry drop, so we removed that support. Now we have combined multiple images to one, I think probably we should get it back - but I'm not sure if the issue still exists, since current image is still large.

You could refer to https://github.com/argoproj/argo-events/pull/843.

The workflow to build images is https://github.com/argoproj/argo-events/blob/master/.github/workflows/release.yml.

mickkael:
I see. It's Github actions. It has only x86 hosts available, so everything else is emulated, and takes ages to compile. I'll dig and see if I can achieve something.
in the meantime, I have published the multmanifest image on my repo for anyone who needs
docker.io/mickkael/argo-events:latest

mickkael:
Close as #1218 supersedes it.

PR #1218, opened 2021-05-17
chore: Workflow to build all Arch

mickkael:
- Changed the Dockerfile to be compatible with different architecture (default remains amd64 if no arg provided)
- Changed "release" workflow for GH Actions
   - build the binaries for all architectures
   - publish the argo-events binaries as part of the release
   - push the docker image for all architectures
   - push a multi-manifest to have a multi-architecture tag
- Changed the makefile to reflect all the previous changes

Checklist:


mickkael:
I've updated the makefile for the test workflow to pass following the modification.
We should be all good now

mickkael:
Hi @whynowy , any comment,  anything to improve?
Thanks for the review!

Makefile
```diff
@@ -39,25 +39,41 @@ endif
...
 
 .PHONY: build
 image: dist/$(BINARY_NAME)-linux-amd64
-	DOCKER_BUILDKIT=1 docker build -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)  --target argo-events -f $(DOCKERFILE) .
+	DOCKER_BUILDKIT=1 docker build -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)  --target $(BINARY_NAME) -f $(DOCKERFILE) .
 	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION); fi
 
+## the command "argo version" has to be removed from dockerfile for this to work
+image-linux-multi: image-linux-amd64 image-linux-arm64
+	echo $(jq -c '. + { "experimental": "enabled" }' ${HOME}/.docker/config.json) > ${HOME}/.docker/config.json
```
whynowy, line 67: Using `DOCKER_CLI_EXPERIMENTAL=enabled` is better, which does not change any settings.

Makefile
```diff
@@ -39,25 +39,41 @@ endif
...
+	docker manifest create $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION) $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-arm64 $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-amd64
+	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker manifest push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION); fi
+image-linux-amd64: dist/$(BINARY_NAME)-linux-amd64
```
whynowy, line 70: Using wildcard `image-linnux-*` instead of listing all of them?

Makefile
```diff
@@ -39,25 +39,41 @@ endif
...
+	DOCKER_BUILDKIT=1 docker build -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-amd64  --target $(BINARY_NAME) -f $(DOCKERFILE) .
```
whynowy, line 71: ARCH not specified?

.github/workflows/release.yml
```diff
@@ -23,29 +28,167 @@ jobs:
...
+      - name: store artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: binaries
+          path: dist
 
-      - name: Release binaries
-        uses: softprops/action-gh-release@v1
-        if: startsWith(github.ref, 'refs/tags/')
+  build-push-linux-arm64:
```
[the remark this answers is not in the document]
whynowy: Also, can we just do `DOCKER_PUSH=true make image-linux-multi` to build and push images?

Makefile
```diff
@@ -39,25 +39,37 @@ endif
...
+	DOCKER_BUILDKIT=1 docker build -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)  --target $(BINARY_NAME) -f $(DOCKERFILE) .
 	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION); fi
 
+image-linux-%: dist/$(BINARY_NAME)-linux-$*
+	DOCKER_BUILDKIT=1 docker build --build-arg "ARCH=$*" -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-$* --platform "linux/$*" --target $(BINARY_NAME) -f $(DOCKERFILE) .
+	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-$*; fi
+
+image-multi: image-linux-amd64 image-linux-arm64
+	DOCKER_CLI_EXPERIMENTAL=enabled docker manifest create $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION) $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-arm64 $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-amd64
+	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker manifest push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION); fi
```
whynowy, line 71: Does `docker manifest push` need `DOCKER_CLI_EXPERIMENTAL=enabled`?

.github/workflows/release.yml
```diff
@@ -23,29 +28,86 @@ jobs:
...
         run: |
           tag=$(basename $GITHUB_REF)
           if [ $tag = "master" ]; then
             tag="latest"
           fi
-          make image IMAGE_TAG=${tag} IMAGE_NAMESPACE=${QUAYIO_ORG} DOCKER_PUSH=true
+          image_name="${QUAYIO_ORG}/${TARGET}:${tag}"
+          echo "::set-output name=TAG_LIST::$image_name"
+
+      - name: Container build and push
```
whynowy, line 83: Can we utilize `make image-multi` command to build and push images?

mickkael (commented):
still WIP

Makefile
mickkael: I rewrite with buildx that allows to push a multi image directly without playing with manifest

.github/workflows/release.yml
mickkael: it would not be efficient as is.
As we want to keep the container build short to avoid any quay or docker registry login timeout, we build the binaries in the previous step. it also allows to attach the binaries to the release.
But in the make image-multi we actually build the binaries as part of it.

What I could do is to have a switch in the make image-multi to optionally not build the binaries. I send it in few min

mickkael: done

Makefile
mickkael: rewritten differently with buildx

mickkael:
@whynowy , ready for a "hopefully" final review.
- the workflow builds using makefile and not github action
- the build and push of the container arm64/amd64 multi image takes 42s, so the login timeout is very unlikely
- the binary build takes 10 min and do amd64, arm64, arm, ppc64le, s390x . If needed some arch could be removed to speed it up.
- the binaries are added to the release, gzipped
- the code could be easily modified to have a multi registry push

I'm now fluent with github actions and makefile !

Makefile
```diff
@@ -39,25 +41,40 @@ endif
...
 image: dist/$(BINARY_NAME)-linux-amd64
-	DOCKER_BUILDKIT=1 docker build -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)  --target argo-events -f $(DOCKERFILE) .
+	DOCKER_BUILDKIT=1 docker build -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)  --target $(BINARY_NAME) -f $(DOCKERFILE) .
 	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION); fi
 
+image-linux-%: dist/$(BINARY_NAME)-linux-$*
+	DOCKER_BUILDKIT=1 docker build --build-arg "ARCH=$*" -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-$* --platform "linux/$*" --target $(BINARY_NAME) -f $(DOCKERFILE) .
+	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-$*; fi
+
+image-multi: set-qemu $(if $(DO_BUILD), dist/$(BINARY_NAME)-linux-arm64 dist/$(BINARY_NAME)-linux-amd64)
```
whynowy, line 71: When I run `make image-multi` on my laptop, I see:

```
docker pull tonistiigi/binfmt:latest
latest: Pulling from tonistiigi/binfmt
Digest: sha256:f52cfb6019e8c8d12b13093fd99c2979ab5631c99f7d8b46d10f899d0d56d6ab
Status: Image is up to date for tonistiigi/binfmt:latest
docker.io/tonistiigi/binfmt:latest
docker run --rm --privileged tonistiigi/binfmt:latest --install amd64,arm64
2021/06/02 23:05:36 installing: amd64 cannot write to /proc/sys/fs/binfmt_misc/register: write /proc/sys/fs/binfmt_misc/register: no such file or directory
2021/06/02 23:05:36 installing: arm64 qemu-aarch64 already registered
...
```

Is that error expected?

.github/workflows/release.yml
```diff
@@ -23,29 +28,97 @@ jobs:
...
+    name: Build & push linux/amd64 and linux/arm64
+    needs: [ build-binaries ]
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        target: [ argo-events ]
+    steps:
+      - uses: actions/checkout@v2
+
+#      - name: Set up QEMU
```
whynowy, line 49: clean up.

Makefile
```diff
@@ -17,6 +17,7 @@ DOCKER_PUSH?=false
 IMAGE_NAMESPACE?=quay.io/argoproj
 VERSION?=latest
 BASE_VERSION:=latest
+DO_BUILD?=true
```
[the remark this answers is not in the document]
mickkael: Yes.
In the github release workflow, we do the build separately to avoid the timeout of the quay/docker login, and also have the binaries handy to attach to the release page. In a manual build, we expect the image build order to build the binaries by default.

Makefile
mickkael: Yes, this is expected as you build on amd64. If you build on an arm64 platform, the error will be on the arm64 line. It doesn't stop the process and can be ignored

mickkael (commented):
clean up done, and comments replied

Makefile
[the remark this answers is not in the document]
whynowy: `DO_BUILD` is only used at:

```
image-multi: set-qemu $(if $(DO_BUILD), dist/$(BINARY_NAME)-linux-arm64 dist/$(BINARY_NAME)-linux-amd64)
```

Even `DO_BUILD` is true, if the binaries are already in place, it will not rebuild them. So to my understanding, it's useful at all. Correct me if I'm wrong.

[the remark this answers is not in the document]
mickkael: I did the test outside of github actions, you are right. So I've clean this.

mickkael:
I have added a set-buildx instruction in the makefile. I guess it's useful to have buildx setup automatically rather than being a prereq.

Makefile
```diff
@@ -35,29 +38,51 @@ VERSION=$(GIT_TAG)
...
+	@if [ "$(DOCKER_PUSH)" = "true" ]; then docker push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)-linux-$*; fi
+
+image-multi: set-buildx set-qemu dist/$(BINARY_NAME)-linux-arm64.gz dist/$(BINARY_NAME)-linux-amd64.gz
+	docker buildx build --tag $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION) --target $(BINARY_NAME) --platform linux/amd64,linux/arm64 --file ./Dockerfile ${PUSH_OPTION} .
+
+set-qemu:
+	docker pull tonistiigi/binfmt:latest
+	docker run --rm --privileged tonistiigi/binfmt:latest --install amd64,arm64
+
+set-buildx:
```
whynowy, line 82: I don't think this is needed. Let's assume buildx is available in either the desktop or github actions env.

I'm ok with rest parts, let's see what it performs.

whynowy (approved):
Conditional approve, please remove the `set-buildx` part.

Makefile
mickkael: set-buildx removed

PR #1220, opened 2021-05-18
fix: Kafka/SASL Verbiage cleanup and examples

cbuckley01:
To be consistent with the existing TLS configuration and be more descriptive, rename Kafka/SASL `user` configuration to `userSecret` and `password` to `passwordSecret`

Add Kafka SASL eventSource example

Update descriptions/error messages to be more accurate


pkg/apis/common/common.go
```diff
@@ -136,9 +136,9 @@ type SASLConfig struct {
 	Mechanism string `json:"mechanism,omitempty" protobuf:"bytes,1,opt,name=mechanism"`
 	// User is the authentication identity (authcid) to present for
 	// SASL/PLAIN or SASL/SCRAM authentication
-	User *corev1.SecretKeySelector `json:"user,omitempty" protobuf:"bytes,2,opt,name=user"`
+	UserSecret *corev1.SecretKeySelector `json:"user,omitempty" protobuf:"bytes,2,opt,name=user"`
```
whynowy, line 139: 1. If `json: ""user,xxxxx,name=user` is not changed, the spec is not changed.
2. A good practice would be adding 2 additional properties, and mark these existing 2 as `Deprecated`, and state they will be unsupported after 2 major releases (v1.6).  Take line 120-122 as an example.

cbuckley01: Doh! Missed #1, good catch, I was considering #2, but since its unreleased I wasn't sure if the cruft would be needed

whynowy: Oh... I forgot it's not released yet... then ignore the second one.

USERS.md
```diff
@@ -18,3 +18,4 @@ Organizations below are **officially** using Argo Events. Please send a PR with
 1. [Viaduct](https://www.viaduct.ai/)
 1. [Woolworths Group](https://www.woolworthsgroup.com.au/)
 1. [7shifts](https://www.7shifts.com)
+1. [ProPoint Solutions] (https://supersalon.com)
```
whynowy, line 21: Could you put it in alphabetical order (move `7Shifts` to the second as well)?

whynowy (approved):
LGTM, thanks!

PR #1203, opened 2021-05-07
chore: Combine binaries/images to one

whynowy:
Signed-off-by: Derek Wang <whynowy@gmail.com>

1. Combine images to one;
2. Switch to use `quay.io/argoproj`.

Checklist:

* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).

PR #1221, opened 2021-05-19
chore: Fix make command

svmaris:
target "all" was replaced by "build" in #1203

Checklist:


whynowy (approved):
Thanks for fixing it!

Issue #1177, opened 2021-04-12
GitHub Webhooks race condition


davidcollom:
**Describe the bug**
We have seen multiple times a race condition when GithubEventSoures are evicted, in which the new eventSource receives the hook configuration before it has been deleted when `DeleteHookOnFinish` is enabled.

**To Reproduce**
1. Have a github eventsource with `DeleteHookOnFinish` enabled
2. Evict/delete the pod running
3. Watch for the new pod to start
4. If done fast enough, the `DeleteHookOnFinish` action occurs AFTER the new pod has initilised 

**Expected behavior**
There should be a webhook registered to said repo.

**Environment (please complete the following information):**
 - Kubernetes: `v1.18.16-gke.302`
 - Argo: `v2.12.10`
 - Argo Events: `v1.2.1`

**Additional context**
It is not possible to set `maxUnavailable` / `maxAvailable` on the `Deployment` of the EventSource resource, nor is it possible within the scenario to ensure that a webhook is setup even after initialisation.

One suggestion/Solution would be that a reconcile loop/controller be implemented to ensure that the hooks are configured as expected and re-start the initialisation process if not.

Further context/discussion was raised in Argo Slack at: https://argoproj.slack.com/archives/CAM37QYMS/p1618248491217900

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
Could you give more detail about the reason of using `DeleteHookOnFinish`? I'm interested in the use case.

davidcollom:
Sure! Because argo-events manages the creation of the webhook, and the EventSource object for each repository required to be monitoring by our platform, it only makes sense that we use the `DeleteHookOnFinish` to clean up said hooks if we no longer need to monitor the repository.
If a repository no longer requires to be monitored we delete the EventSource and pragmatically remove this as a dependency on all relevant Sensor Objects.
This allows our platform to almost entirely be configured within kubernetes and doesn't require a third party or additional configuration steps for our end users, and they're able to subscribe/register their repositories as and when they required.

It's worth noting that in #1112 We only monitor for `push` events and allow filtering on our Sensors to the corresponding branch/reference.

Due to security requirements, we're not able to use a generic webhook endpoint as some of these repositories may not be owned/managed by the organisation therefore a unique endpoint/URI is generated for each repository.
[...]

PR #1223, opened 2021-05-20
feat: add eventsource support for Bitbucket Server. Fixes #891

### LINKED ISSUE #891: Add Support for Bitbucket Server Webhooks
Github and Gitlab are already supported as trigger sources by Argo Events, it would be nice if Bitbucket Server would be supported too. Actions like PR creations or pushes should be able to trigger PR builds via Argo Events and Workflows.

Add Bitbucket Server as an officially supported event source.
https://argoproj.github.io/argo-events/concepts/event_source/

General Webhooks could be used maybe.

**Additional context**
Here is the Bitbucket Webhook documentation: https://support.atlassian.com/bitbucket-cloud/docs/event-payloads/

---

svmaris:
Signed-off-by: Stephan van Maris <s.vanmaris@teamitg.com>

Checklist:


eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,274 @@
...
+	logger.With("hook-id", createdHook.ID).Info("hook succesfully registered")
+	return nil
+}
+
+// PostInactivate performs operations after the route is inactivated
+func (router *Router) PostInactivate() error {
+	bitbucketserverEventSource := router.bitbucketserverEventSource
+	route := router.route
+
+	if bitbucketserverEventSource.DeleteHookOnFinish {
```
whynowy, line 240: Race conditions will happen because of this, see https://github.com/argoproj/argo-events/issues/1177. Check github event source for the workaround.

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -112,6 +112,8 @@ type EventSourceSpec struct {
 	Generic map[string]GenericEventSource `json:"generic,omitempty" protobuf:"bytes,28,rep,name=generic"`
 	// Replicas is the event source deployment replicas
 	Replicas *int32 `json:"replicas,omitempty" protobuf:"varint,29,opt,name=replicas"`
+	// Bitbucket Server event sources
+	BitbucketServer map[string]BitbucketServerEventSource `json:"bitbucketserver,omitempty" protobuf:"bytes,30,rep,name=bitbucketserver"`
```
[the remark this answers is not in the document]
svmaris: This was done on purpose to allow for a bitbucketcloud eventsource to be added later. Bitbucket Cloud has a different API.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,274 @@
...
+		logging.LabelEndpoint, route.Context.Endpoint,
+		logging.LabelPort, route.Context.Port,
+		logging.LabelHTTPMethod, route.Context.Method,
+		"project-key", bitbucketserverEventSource.ProjectKey,
+		"repository-slug", bitbucketserverEventSource.RepositorySlug,
+		"base-url", bitbucketserverEventSource.BitbucketServerBaseURL,
+	)
+
+	logger.Info("retrieving the access token credentials...")
+	bitbucketCredentials, err := router.getCredentials(bitbucketserverEventSource.AccessToken)
```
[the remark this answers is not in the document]
svmaris: Correct me if I'm wrong, but isn't the point of a specific eventsource to be able to setup the webhook declaratively? If I don't want to create a webhook in Bitbucket Server automatically, users can create a Webhook eventsource in Argo Events and configure it manually from within Bitbucket

pkg/apis/eventsource/v1alpha1/types.go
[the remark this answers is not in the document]
whynowy: 👍

eventsources/sources/bitbucketserver/start.go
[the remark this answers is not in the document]
whynowy: Make sense.

eventsources/sources/bitbucketserver/start.go
svmaris: Please note that I copied the boilerplate for this EventSource from the Gitlab EventSource, so that probably means Gitlab is suffering from the same race conditions. The comments in the Github EventSource make it seem that it's a Github-specific issue ("Github can not handle race conditions well"), but in fact all EventSources that depend on external resources being created/destroyed suffer from the same issue (currently gitlab and awssns, and I believe the Stripe EventSource will create a webhook for every replica).

IMHO, adding the workaround is a bit too much of a kludge and the logic for managing external EventSource dependencies should move elsewhere. The current workaround causes the webhooks to be deleted and recreated for every replica and only works correctly when the number of replicas in a RollingUpdate stays the same (or is higher), otherwise the deletes "win" the race.

whynowy: > Please note that I copied the boilerplate for this EventSource from the Gitlab EventSource, so that probably means Gitlab is suffering from the same race conditions. The comments in the Github EventSource make it seem that it's a Github-specific issue ("Github can not handle race conditions well"), but in fact all EventSources that depend on external resources being created/destroyed suffer from the same issue (currently gitlab and awssns, and I believe the Stripe EventSource will create a webhook for every replica).
> 

This is true. I never tried Gitlab and others, not sure if they handle the race well.

> IMHO, adding the workaround is a bit too much of a kludge and the logic for managing external EventSource dependencies should move elsewhere. The current workaround causes the webhooks to be deleted and recreated for every replica and only works correctly when the number of replicas in a RollingUpdate stays the same (or is higher), otherwise the deletes "win" the race.

I don't like the webhook creation being handled by the eventsource, and wanted to deprecate it, https://github.com/argoproj/argo-events/issues/1112, but there were objections. The workaround works, because it keeps checking and creating.

svmaris: > I don't like the webhook creation being handled by the eventsource, and wanted to deprecate it, #1112, but there were objections. The workaround works, because it keeps checking and creating.

I really like the fact that the webhooks are set-up automatically. It's also shown in one of the [videos](https://www.youtube.com/watch?v=XNXJtxkUKeY) linked from the Argo-Events homepage.

I've now added the same race condition handling code to the bitbucketserver eventsource and it seems to work fine.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	go func() {
+		// Another kind of race conditions might happen when pods do rolling upgrade - new pod starts
+		// and old pod terminates, if DeleteHookOnFinish is true, the hook will be deleted from bitbucket.
+		// This is a workround to mitigate the race conditions.
+		logger.Info("starting bitbucket hooks manager daemon")
+		ticker := time.NewTicker(60 * time.Second)
+		defer ticker.Stop()
+		for {
```
whynowy, line 177: Putting this logic in `PostActive()` has a problem that `<- ctx.Done()` below never gets a chance to reach, try it by tailing the pod log and kill it to see if `exiting bitbucket hooks manager daemon` is printed. Correct me if I'm wrong.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,322 @@
...
+
+	if !isAlreadyExists {
+		apiResponse, err = router.bitbucketClient.DefaultApi.CreateWebhook(bitbucketserverEventSource.ProjectKey, bitbucketserverEventSource.RepositorySlug, localVarPostBody, []string{"application/json"})
+
+		if err != nil {
+			return errors.Errorf("failed to add webhook. err: %+v", err)
+		}
+
+	} else {
+		logger.Info("webhook already exists, updating it...")
```
whynowy, line 284: I'm not familiar with bitbucket, but have a feeling that calling the update API once (or multiple times if replicas > 1) a minute sounds not good.

eventsources/sources/bitbucketserver/start.go
VaibhavPage: 1. Why does the goroutine keep on creating webhooks? shouldn't it return after the first successful bitbucket webhook creation? 
2. Shouldn't the function return after line 164 if err == nil.
3. As @whynowy  mentioned, I am not sure if retrying webhook creation every 1 min makes sense. We can instead use exponential backoff and try to create the webhook as opposed to running a daemon.
4. Also if we go with the daemon, we should use signal.notify on sigterm instead of ctx.Done on line 179

svmaris: Thanks for the review. I'll clean this up. The current PR is a mix of the gitlab eventsource (that suffers from the same race conditions) and the fixes that have been done on the github eventsource. 

Like @whynowy mentioned before: I don't like the way you need to cope with these kind of race conditions in every eventsource that creates external resources.

There's obviously a demand for this use case (the Bitbucket eventsource is the nr1 voted issue currently), but I think a decision has to be made on how to implement this in a generic way.

I'm not that familiar with the design of Argo Events, but maybe this can be solved using a install/uninstall Job when creating/deleting the EventSource k8s resource, so it's not tied to the lifecycle of the individual Pods. Or maybe using a finalizer to clean up the external resource. 

This way the responsibility is moved to the eventsource-controller and the external resources can be left alone during RollingUpdates.

Any thoughts?

whynowy: Is `DeleteHookOnFinish` a must have feature? It is root cause of the race condition. If we abandon it, it avoids lots of problem. Question is how many ppl want it, I don't know.

My suggestion:
1. No `DeleteHookOnFinish` support.
2. If ppl do want `DeleteHookOnFinish`, use the workaround at this time.

svmaris: If you want to go "full gitops", the `DeleteHookOnFinish` is a must have feature. It shouldn't even be possible to *not* remove the webhook, because it doesn't make sense to keep the webhook around when there's no EventSource to process the events.

I think this feature is very powerful and in fact, the video of @vfarcic linked from the frontpage is what has drawn me towards the whole Argo "suite" of tools.

The point is not that it's *hard* to configure or cleanup the webhooks in Bitbucket (or Github, or Gitlab, etc.), the point is that it's a *manual* step I'd like to avoid as much as possible.

I think there's a big opportunity for Argo Events here, because there are a lot of systems out there that support some kind of webhooks that can be controlled using an API.

For now, I'll just fix the Bitbucket Server EventSource to handle the race conditions a bit better and keep it on par with the Github implementation.

Thanks!

whynowy: Sounds good, thanks!

whynowy:
@svmaris - could you continue to get it done?

svmaris:
> @svmaris - could you continue to get it done?

@whynowy Yes, I will make some time this weekend to get it done. Sorry for the delay.

aslafy-z:
@svmaris any news on this PR? any chance it gets finished soon?

svmaris:
Check again please!

svmaris:
Don't know what's wrong with the codegen check?

svmaris:
@whynowy Do you have a chance to look at this please?

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	logger.Info("retrieving the webhook secret...")
+	webhookSecret, err := router.getWebhookSecret(bitbucketserverEventSource.WebhookSecret)
+	if err != nil {
+		return errors.Errorf("failed to get bitbucketserver webhook secret. err: %+v", err)
+	}
+
+	newHook := bitbucketv1.Webhook{
+		Name:          "Argo Events",
```
whynowy, line 304: Does duplicate name matter?

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,354 @@
...
+		}
+
+		err = mapstructure.Decode(apiResponse.Values, &createdHook)
+		if err != nil {
+			return errors.Errorf("failed to convert API response to Webhook struct. err: %+v", err)
+		}
+		router.hookID = createdHook.ID
+
+		logger.With("hook-id", createdHook.ID).Info("hook succesfully registered")
+	}
```
whynowy, line 333: Adding a clean `return` will make it more readable.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,354 @@
...
+	if isAlreadyExists && (!reflect.DeepEqual(existingHook.Events, newHook.Events) || !reflect.DeepEqual(existingHook.Configuration, newHook.Configuration)) {
+		logger.Info("webhook already exists and configuration has changed. Updating webhook.")
+
+		apiResponse, err = bitbucketClient.DefaultApi.UpdateWebhook(bitbucketserverEventSource.ProjectKey, bitbucketserverEventSource.RepositorySlug, int32(existingHook.ID), localVarPostBody, []string{"application/json"})
+
+		if err != nil {
+			return errors.Errorf("failed to update webhook. err: %+v", err)
+		}
+
+		err = mapstructure.Decode(apiResponse.Values, &createdHook)
```
whynowy, line 345: It looks like this part can be skipped, as it's an update, and the hook id is already there.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,354 @@
...
+	controller = webhook.NewController()
+)
+
+// set up the activation and inactivation channels to control the state of routes.
+func init() {
+	go webhook.ProcessRouteStatus(controller)
+}
+
+// getCredentials retrieves credentials to connect to Bitbucket Server
+func (router *Router) getCredentials(keySelector *corev1.SecretKeySelector) (*cred, error) {
```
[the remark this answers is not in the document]
svmaris: I agree. There's another wrapper function and struct for fetching the Webhook secret just below this one. If you agree, I'll remove both the structs and wrapper functions and replace them with plain calls to `GetSecretFromVolume`. I'll keep the Github integration as is to keep this PR clean. I can open a new PR for Github if you'd like.

pkg/apis/common/common.go
```diff
@@ -71,6 +72,7 @@ var (
 		HDFSEvent,
 		FileEvent,
 		GenericEvent,
+		BitbucketServerEvent,
```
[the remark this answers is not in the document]
svmaris: Done

eventsources/sources/bitbucketserver/start.go
svmaris: Done, thanks!

eventsources/sources/bitbucketserver/start.go
svmaris: You're right. Fixed it. Thanks!

eventsources/sources/bitbucketserver/start.go
svmaris: It doesn't. Bitbucket Server doesn't care. Matching (to find the existing hook) is based on the Url and updating/deleting is done by Id. I don't like hard coded names like these and I tried to find another way of naming the hook so you can relate it to a specific EventSource, but I failed to find anything that I could use. If you have any suggestion for a better name, please let me know. Otherwise I'll just keep it like this.

svmaris:
@whynowy Thanks for the review. I've updated the PR. For some reason the codegen check keeps failing but I don't understand why. I looks like a small diff in the escaping of "#" in some URL's.

whynowy:
> @whynowy Thanks for the review. I've updated the PR. For some reason the codegen check keeps failing but I don't understand why. I looks like a small diff in the escaping of "#" in some URL's.

It's because some of your tools in $GOPATH/bin are out of date, no worries, you can ignore it and I'll fix it later.

eventsources/sources/bitbucketserver/start.go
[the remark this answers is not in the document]
whynowy: 👍

eventsources/sources/bitbucketserver/start.go
whynowy: I'm ok with it as long as bitbucket doesn't care.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,327 @@
...
+		Events:        bitbucketserverEventSource.Events,
+		Configuration: bitbucketv1.WebhookConfiguration{Secret: webhookSecret},
+	}
+
+	localVarPostBody, err := json.Marshal(newHook)
+	if err != nil {
+		return errors.Wrapf(err, "failed to marshal new webhook to JSON")
+	}
+
+	var createdHook *bitbucketv1.Webhook
```
whynowy, line 292: Minor: move it into `if` block below.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -0,0 +1,327 @@
...
+	if isAlreadyExists && (!reflect.DeepEqual(existingHook.Events, newHook.Events) || !reflect.DeepEqual(existingHook.Configuration, newHook.Configuration)) {
+		logger.Info("webhook already exists and configuration has changed. Updating webhook.")
+
+		_, err = bitbucketClient.DefaultApi.UpdateWebhook(bitbucketserverEventSource.ProjectKey, bitbucketserverEventSource.RepositorySlug, int32(existingHook.ID), localVarPostBody, []string{"application/json"})
+
+		if err != nil {
+			return errors.Errorf("failed to update webhook. err: %+v", err)
+		}
+
+		logger.With("hook-id", createdHook.ID).Info("hook succesfully updated")
```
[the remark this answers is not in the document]
svmaris: 🤦

eventsources/sources/bitbucketserver/start.go
svmaris: Thanks. That should've prevented me from making the previous mistake.

whynowy (approved):
LGMT, thanks for getting this done!

aslafy-z:
Any chance this gets released anytime soon? Thanks for getting this done @svmaris!

PR #1226, opened 2021-05-23
feat: Introduce expr filters

VaibhavPage:
Checklist:


whynowy:
Welcome back!

sensors/dependencies/filter.go
```diff
@@ -252,3 +256,60 @@ filter:
...
+	for _, filter := range filters {
+		parameters := map[string]interface{}{}
+		for _, field := range filter.Fields {
+			result := gjson.GetBytes(jsData, field.Path)
+			if !result.Exists() {
+				return false, fmt.Errorf("path %s does not exist", field.Path)
+			}
+			switch field.Type {
+			case v1alpha1.JSONTypeString:
+				parameters[field.Name] = result.Str
```
whynowy, line 291: Can you try `parameters[field.Name] = result.Value()`? I feel like that will work, and `fields` is not required.

whynowy (commented):
Do you think this can replace the `dataFilter`, especially if it does not need `fields`?

whynowy:
You also need to clean up some outdated, codegen related binaries in  {gopath}/bin/** to fix the codegen check error.

examples/sensors/filter-with-expressions.yaml
```diff
@@ -0,0 +1,70 @@
...
+  name: with-expression-filters
+spec:
+  dependencies:
+    - name: test-dep
+      eventSourceName: webhook
+      eventName: example
+      filters:
+      # If event payload passes either one of the following filters, the event is considered a valid event.
+        exprs:
+          - expr: a == "b" || c == 10
```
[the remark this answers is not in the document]
VaibhavPage: you'd need the fields because they represent the variables in the expression. e.g. `govaluate` won't understand field `d.e`

sensors/dependencies/filter.go
VaibhavPage: yeah that would work

VaibhavPage:
> Do you think this can replace the `dataFilter`, especially if it does not need `fields`?

Yep, I think we should deprecate dataFilters in a future release

VaibhavPage:
> You also need to clean up some outdated, codegen related binaries in {gopath}/bin/** to fix the codegen check error.

What's this step does "Ensure nothing has changed" in the pipeline?

whynowy:
> > You also need to clean up some outdated, codegen related binaries in {gopath}/bin/** to fix the codegen check error.
> 
> What's this step does "Ensure nothing has changed" in the pipeline?

That uses a fresh env to run `make codegen`, and check if there's any file changes. Here it fails which means the codegen files on your local env are different from a fresh set up env generated.

cbuckley01:
Hey guys, thanks for adding this, we are eagerly awaiting to give this a whirl, anything I can do to help it get merged in?

cbuckley01:
Thanks @whynowy !

PR #1260, opened 2021-06-22
chore: Upgrade to K8s V0.20.4 

uturunku1:
Checklist:

* [X ] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).

Hi! This is ready for review. This PR is a complement to this other argo-workflows PR https://github.com/argoproj/argo-workflows/pull/6143 to upgrade k8s dependencies

I ran `make test`, `make lint` and `make codegen`. When I ran `make codegen`, swagger complained about `The swagger spec at "api/openapi-spec/swagger.json" is invalid against swagger specification 2.0` and displayed multiple errors like this one:

```
- definitions.io.argoproj.eventsource.v1alpha1.WatchPathConfig.directory.default in body must be of type string: "null"
```

I was able to hide those errors by copying some of the logic I found in argo-workflows `hack` directory:  https://github.com/argoproj/argo-workflows/tree/master/hack/swagger

Let me know if there is a better or different approach to deal with this error that you'd like to suggest.

Thanks!


api/openapi-spec/filtered.swagger.json
```diff
@@ -0,0 +1,6742 @@
+{
```
whynowy, line 1: What is this file used to do?

uturunku1: Hi Derek! This `api/openapi-spec/filtered.swagger.json` is a copy of `api/openapi-spec/swagger.json`. After I upgraded the kube-openapi version, the swagger.json output file was including a new 
 "default" field with a value of an empty string or empty object causing the error that I mentioned in the PR description. To avoid that error, I filtered out those "default" fields from the original swagger.json output and the result of that is a new file, filtered.swagger.json (https://github.com/argoproj/argo-events/blob/72bdef995b759db1441856aa458b13b041086778/hack/update-swagger.sh#L30) that does no longer have those "default" fields.
I made the assumption that filtering out those "default" fields was okay because I noticed it had been previously done for Argo Workflows as well (https://github.com/argoproj/argo-workflows/blob/v3.0.7/hack/swagger/kubeifyswagger.go#L45-L53) @whynowy

whynowy: I see, thanks!

Could you:

1. Only keep the filtered one and name it `swagger.json`, put it at its original directory;
2. Merge `kubeifyswagger` into https://github.com/argoproj/argo-events/blob/master/hack/gen-openapi-spec/main.go?

uturunku1: Ok! I moved my kubeifyswagger logic into  https://github.com/argoproj/argo-events/blob/master/hack/gen-openapi-spec/main.go.  Let me know if those changes work for you or if you'd like me to make other adjustments @whynowy

whynowy (approved):
LGTM. Thanks, this will be released after workflow gets similar upgrade.

PR #1271, opened 2021-07-01
docs: Update 06-trigger-conditions.md

liyjcdl:
docs: Update 06-trigger-conditions.md
The last line seems unfinished, and missed a verb between "should following".  Proposed to change it from "You should following output, " to "You should get the following output." .

Checklist:


whynowy (approved):
Thanks for fixing it!

PR #1288, opened 2021-07-27
chore: Add pandoc to executable check in Makefile

blkperl:
Checklist:

With the patch applied. `make codegen` warns you to install pandoc.

Example output: 
```
make codegen
Makefile:41: *** "No pandoc in PATH".  Stop.
```

```
 brew install pandoc
```

```
make codegen
./hack/generate-proto.sh
[.....]
```


whynowy:
Adding this breaks other ci jobs like e2e, as they don't need `pandoc`

blkperl:
@whynowy do you have a suggestion on how to fix the tests? I'm not sure how to allow it for the other targets

whynowy:
Didn't you see `pandoc` check when you run `make codegen`? I thought it has been in place - https://github.com/argoproj/argo-events/blob/master/hack/update-api-docs.sh#L12

blkperl:
Yes when I ran `make codegen` it ran for a while and then failed when updating the docs. I figured it would be better to be told earlier in the process but we can also just close the PR if that's easier since it's just a minor inconvenience .

Issue #1311, opened 2021-08-14
Unable to use eventbus with istio.


antoniomo:
**Describe the bug**

When using the eventbus as created by the eventbus controller, with "native" strategy, the sensors can't connect to the eventbus if they have an istio-proxy sidecar. In organizations using istio, it's common to have the istio sidecar at the very least on the webhook event sources.

**To Reproduce**

Install a default eventbus, try to connect from an eventsource with istio-proxy sidecar.

---
**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

oursland:
This issue has prevented me from using Argo Events in my deployments.

The error noted on issue #855 by @lironrisk and observed by @ajoskowski and @wajdi-datalvo matches the one exhibited when deploying Argo Events in an Istio-enabled namespace.

The relevant line, pretty-printed:

```
[
   {
      "caller" : "driver/nats.go:102",
      "clientID" : "client-webhook-eventsource-4p9tf-5df7969b6b-nvqhm-629",
      "error" : "read tcp 10.42.0.40:42834->10.42.2.28:4222: i/o timeout",
      "eventSourceName" : "webhook",
      "level" : "error",
      "logger" : "argo-events.eventsource",
      "msg" : "Failed to connect to NATS server",
[...]

PR #1312, opened 2021-08-14
fix(eventbus): allow clients with istio sidecar. Fixes: #1311

antoniomo:
Istio-related issues affecting this behaviour:

- https://github.com/nats-io/nats-operator/issues/88
- https://github.com/istio/istio/issues/28623

The fix is very simple. I added the related issues to the code comment,
let me know if that's undesirable or there's something else to clarify.

Checklist:


antoniomo:
Fixes: #1311

controllers/eventbus/installer/nats.go
```diff
@@ -414,7 +414,11 @@ func (i *natsInstaller) buildStanService() (*corev1.Service, error) {
 		Spec: corev1.ServiceSpec{
 			ClusterIP: corev1.ClusterIPNone,
 			Ports: []corev1.ServicePort{
-				{Name: "client", Port: clientPort},
+				// Prefix tcp- to enable clients to connect from
+				// an istio-enabled namespace, following:
+				// https://github.com/nats-io/nats-operator/issues/88
+				// https://github.com/istio/istio/issues/28623
+				{Name: "tcp-client", Port: clientPort},
```
whynowy, line 421: Have you tested if this will impact upgrading from an existing EventBus?

antoniomo: Hi!

Good point, I need to test with and without Istio (not just with). I will have time in the weekend to do more thoroughly testing, any particular scenario you would like to see checked?

- Without Istio, run pre-patch, apply post-patch, everything should still work.
- With Istio, run pre-patch (doesn't work), apply post-patch, everything should work.
- Fresh installs on both cases.

Does this cover it?

whynowy: Sorry for the late response, just came back from vacation.

I'm more concerned about if the port name change is allowed on an existing service.

antoniomo: Got it! I'll check that as well, some delays getting time on my side.

whynowy: I tested it, upgrading has no issue, thanks!

whynowy (approved):

controllers/eventbus/installer/nats.go
antoniomo: Awesome! Thank you very much! I was unable to get time for this, deployments "season" at work have kept me busy and I kept postponing it :/

Thanks again!

PR #1317, opened 2021-08-20
fix(trigger): correct field name in error message

Windfarer:
The field name `generatedName` in the error of argo_workflow trigger is wrong. The correct name is `generateName`.
ref: https://github.com/kubernetes/apimachinery/blob/02cfb53916346d085a6c6c7c66f882e3c6b0eca6/pkg/apis/meta/v1/unstructured/unstructured.go#L258

Checklist:


whynowy (approved):
Thanks for fixing it!

Issue #1322, opened 2021-08-27
Missing steps in the getting started guide


natalefalco:
**Describe the bug**
On the main getting started guide there is a missing service account creation prequirement/step.
I would add either a requiement warning or extend the existing yaml to create the service acocunt, roles and rolebindings too

**To Reproduce**
1. Go to https://argoproj.github.io/argo-events/quick_start/
2. Follow the instructions

**Expected behavior**
The verification succeed in the end, leaving a better first impression

**Screenshots**

**Environment (please complete the following information):**
 - Kubernetes: v1.21.3+k3s1
 - Argo: v3.1.8
 - Argo Events: 1.4.0

**Additional context**
In the introduction there is a mention to the service users that are missing 
https://argoproj.github.io/argo-events/tutorials/01-introduction/

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

daniel-codefresh:
Here's the PR: #1324

PR #1324, opened 2021-08-28
fix: added service account creation step in quick start docs

daniel-codefresh:

Checklist:


daniel-codefresh:
fixes #1322

natalefalco:
@daniel-codefresh thanks for the account creation step
I think an another role and rolebinding is missing from the default account described here: https://github.com/argoproj/argo-events/blob/master/examples/README.md. Without that I also had no success.
Can you please take a look?

whynowy:
The yaml file can not be put into `sensor` folder, as there's a unit test to validate all the files under that folder to be valid Sensor objects. You probably can put it under `examples`.

natalefalco:
> The yaml file can not be put into `sensor` folder, as there's a unit test to validate all the files under that folder to be valid Sensor objects. You probably can put it under `examples`.

@whynowy isn't it already under the examples ?

whynowy:
> > The yaml file can not be put into `sensor` folder, as there's a unit test to validate all the files under that folder to be valid Sensor objects. You probably can put it under `examples`.
> 
> @whynowy isn't it already under the examples ?

It's mentioned in the README under examples, not a standalone file. I'm ok to have a file for `kubectl apply` quick reference.

natalefalco:
> > > The yaml file can not be put into `sensor` folder, as there's a unit test to validate all the files under that folder to be valid Sensor objects. You probably can put it under `examples`.
> > 
> > 
> > @whynowy isn't it already under the examples ?
> 
> It's mentioned in the README under examples, not a standalone file. I'm ok to have a file for `kubectl apply` quick reference.

I meant in the commit https://github.com/argoproj/argo-events/pull/1324/files
sorry for my misunderstanding

whynowy:
> > > > The yaml file can not be put into `sensor` folder, as there's a unit test to validate all the files under that folder to be valid Sensor objects. You probably can put it under `examples`.
> > > 
> > > 
> > > @whynowy isn't it already under the examples ?
> > 
> > 
> > It's mentioned in the README under examples, not a standalone file. I'm ok to have a file for `kubectl apply` quick reference.
> 
> I meant in the commit https://github.com/argoproj/argo-events/pull/1324/files
> sorry for my misunderstanding

I meant `examples/xxx.yaml` instead of `examples/sensors/xxx.yaml`.

daniel-codefresh:
> The yaml file can not be put into `sensor` folder, as there's a unit test to validate all the files under that folder to be valid Sensor objects. You probably can put it under `examples`.

Sure np, will do

daniel-codefresh:
> @daniel-codefresh thanks for the account creation step
> I think an another role and rolebinding is missing from the default account described here: https://github.com/argoproj/argo-events/blob/master/examples/README.md. Without that I also had no success.
> Can you please take a look?

sure I'll take a look, but actually I had success without it. I wonder which argo-workflows installation are you using? and are you using the same sensor from the example? or have you modified it?

natalefalco:
> > @daniel-codefresh thanks for the account creation step
> > I think an another role and rolebinding is missing from the default account described here: https://github.com/argoproj/argo-events/blob/master/examples/README.md. Without that I also had no success.
> > Can you please take a look?
> 
> sure I'll take a look, but actually I had success without it. I wonder which argo-workflows installation are you using? and are you using the same sensor from the guide? or have you modified it?

Indeed I have just used everything from the guide I have made no modifications on it.
As I in the current state (with the added operate-workflow-sa role ) the workflow is getting created. Since there is not enough permission is given to the user of the workflow (default) in the argo-events namespace, it will hang in a pending state.
I think it the reason is coming from here https://github.com/argoproj/argo-workflows/blob/master/docs/workflow-rbac.md 
@daniel-codefresh did your workflow succeed to create the pod?

daniel-codefresh:
> > > @daniel-codefresh thanks for the account creation step
> > > I think an another role and rolebinding is missing from the default account described here: https://github.com/argoproj/argo-events/blob/master/examples/README.md. Without that I also had no success.
> > > Can you please take a look?
> > 
> > 
> > sure I'll take a look, but actually I had success without it. I wonder which argo-workflows installation are you using? and are you using the same sensor from the guide? or have you modified it?
> 
> Indeed I have just used everything from the guide I have made no modifications on it.
> As I in the current state (with the added operate-workflow-sa role ) the workflow is getting created. Since there is not enough permission is given to the user of the workflow (default) in the argo-events namespace, it will hang in a pending state.
> I think it the reason is coming from here https://github.com/argoproj/argo-workflows/blob/master/docs/workflow-rbac.md
> @daniel-codefresh did your workflow succeed to create the pod?

Yes, with or without the workflow rbac that you've mentioned, I still get the same result: a workflow pod is being created and eventually gets to a "Completed" state, and also the workflow itself ends up in a "Succeeded" status. The workflow logs also seem to be fine. I have tested both namespace and cluster installations of Argo Workflows and had no issues with both. Btw, I used a docker-desktop cluster to test this, which one do you use?

In any case, I don't mind adding it to the guide as well if you're sure that this is the root cause of your workflows being stuck in "pending" state. If possible, please verify it and let me know.

daniel-codefresh:
@whynowy Happy to help:)

natalefalco:
Thanks @daniel-codefresh  :) 
I did a new scratch install and you were right the workflow rbac is not needed something was off with my install. My bad.

PR #1342, opened 2021-09-09
feat: allow insecure GitHub Enterprise instance

aslafy-z:
Checklist:

### DIFF
--- a/eventsources/sources/github/start.go
+++ b/eventsources/sources/github/start.go
@@ -17,6 +17,7 @@ package github
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"math/rand"
@@ -255,6 +256,12 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 			return errors.Wrap(err, "failed to set up auth transport for http client")
 		}
 
+		logger.Info("setting up client for GitHub...")
+		if !githubEventSource.APISkipTLSVerify {
+			authTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+		}
+		client := gh.NewClient(&http.Client{Transport: authTransport})
+
 		logger.Info("configuring GitHub hook...")
 		formattedURL := common.FormattedURL(githubEventSource.Webhook.URL, githubEventSource.Webhook.Endpoint)
 		hookConfig := map[string]interface{}{
@@ -271,9 +278,6 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 		if router.hookSecret != "" {
 			hookConfig["secret"] = router.hookSecret
 		}
-
-		logger.Info("setting up client for GitHub...")
-		client := gh.NewClient(&http.Client{Transport: authTransport})
 
 		logger.Info("setting up base url for GitHub client...")
 		if githubEventSource.GithubBaseURL != "" {
@@ -292,6 +296,7 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 			}
 			client.UploadURL = uploadURL
 		}
+
 		router.githubClient = client
 		router.repoHookIDs = make(map[string]int64)
 		router.orgHookIDs = make(map[string]int64)
--- a/pkg/apis/eventsource/v1alpha1/types.go
+++ b/pkg/apis/eventsource/v1alpha1/types.go
@@ -689,11 +689,14 @@ type GithubEventSource struct {
 	// APIToken refers to a K8s secret containing github api token
 	// +optional
 	APIToken *corev1.SecretKeySelector `json:"apiToken,omitempty" protobuf:"bytes,6,opt,name=apiToken"`
+	// Whether to skip TLS verification when connecting to GitHub for webhook creation, useful for GitHub Enterprise use cases.
+	// +optional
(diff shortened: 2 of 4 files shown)


eventsources/sources/github/start.go
```diff
@@ -197,6 +198,11 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 			Token: apiTokenCreds.secret,
 		}
 
+		httpClient := PATTransport.Client()
```
whynowy, line 201: Minor - could you move it down after `logger.Info("setting up client for GitHub...")`?

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -714,6 +714,9 @@ type GithubEventSource struct {
 	// Repositories holds the information of repositories, which uses repo owner as the key,
 	// and list of repo names as the value
 	Repositories []OwnedRepositories `json:"repositories,omitempty" protobuf:"bytes,15,rep,name=repositories"`
+	// GitHub Insecure TLS verification (for GitHub Enterprise)
+	// +optional
+	GithubInsecure bool `json:"githubInsecure,omitempty" protobuf:"varint,16,opt,name=githubInsecure"`
```
whynowy, line 719: Maybe `SkipVerifyTLS`? And give comments like below:

Whether to skip TLS verification when connecting to Github for webhook creation, useful for Github Enterprise use case.

whynowy (commented):
You need to run `make codegen` after changing models.

aslafy-z:
I applied the reviews and applied the codegen, I have two points:
- HTML and Markdown files were generated with newlines between all lines (due to a pandoc wrong version maybe, I'm running with )
- `manifests/base/crds/argoproj.io_eventbuses.yaml` was generated by the codegen steps, but `manifests/base/crds/argoproj.io_eventbus.yaml` existed at the time. I deleted that file since it looks like it's not valid anymore.
Let me know if I can do something to solve these points.

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -714,6 +714,9 @@ type GithubEventSource struct {
 	// Repositories holds the information of repositories, which uses repo owner as the key,
 	// and list of repo names as the value
 	Repositories []OwnedRepositories `json:"repositories,omitempty" protobuf:"bytes,15,rep,name=repositories"`
+	// Whether to skip TLS verification when connecting to GitHub for webhook creation, useful for GitHub Enterprise use case.
+	// +optional
+	GithubSkipVerifyTLS bool `json:"githubSkipVerifyTLS,omitempty" protobuf:"varint,16,opt,name=githubSkipVerifyTLS"`
```
whynowy, line 719: Could we just. use `SkipVerifyTLS`? It sounds redundant since the spec is already under a field named `github`.

manifests/base/crds/argoproj.io_eventbuses.yaml
```diff
@@ -2,13 +2,13 @@
 apiVersion: apiextensions.k8s.io/v1
```
whynowy, line 2: Please revert it, we don't use `eventbuses` but `eventbus`. We have `plural exceptions` at https://github.com/argoproj/argo-events/blob/master/hack/update-codegen.sh#L41, but somehow your local code-generator version screwed it.

pkg/apis/eventsource/v1alpha1/types.go
aslafy-z: This struct already has `Insecure` which refers to the target GitHub Webhook configuration.

I through that GitHubSkipVerifyTLS would ease the understanding of which field does what. What do you think?

whynowy: Can we do `APISkipVerifyTLS` instead, considering we have `APIToken` as well.

```yaml
spec:
  github:
    example:
      apiToken: xxxx
      apiSkipTLSVerify: true
```

aslafy-z:
Un-stale, will edit and rebase this PR soon.

alexec (changes requested):
looks good so far, please see requested changes

aslafy-z:
@alexec @whynowy I still have the same kind of issue with the `codegen` make target:
- HTML and Markdown files were generated with newlines between all lines
- `go:build` tags are removed in generated go files
Any idea how I can tackle these points?

whynowy:
> @alexec @whynowy I still have the same kind of issue with the `codegen` make target:
> 
> * HTML and Markdown files were generated with newlines between all lines
> * `go:build` tags are removed in generated go files
> 
> Any idea how I can tackle these points?

@aslafy-z - ignore the `codegen` failure, just fix the conflicts and we will take care of the rest, thanks!

aslafy-z:
@whynowy I updated the code without running the codegen. I had no time to test it tho.

eventsources/sources/github/start.go
```diff
@@ -255,6 +256,12 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 			return errors.Wrap(err, "failed to set up auth transport for http client")
 		}
 
+		logger.Info("setting up client for GitHub...")
+		if !githubEventSource.APISkipTLSVerify {
+			authTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
whynowy, line 261: It would be better to do a type assertion check here.

PR #1343, opened 2021-09-09
feat: support for PubSub emulator

avanderveen:
The rest of my application runs locally and uses the PubSub emulator, and I'd like to be able to run Argo Events locally as well.

My application triggers Argo Workflows via the GCP PubSub event source. However, when pointing the PubSub source at the emulator (by setting the `PUBSUB_EMULATOR_HOST` environment variable), the source errors out on the permissions test, since the PubSub emulator doesn't support that gRPC call.

This PR makes the GCP PubSub event source skip the permissions check when the `PUBSUB_EMULATOR_HOST` environment variable is set.

I've tested this locally on my machine, and events are being pulled from the PubSub emulator with these changes.

### DIFF
--- a/eventsources/sources/gcppubsub/start.go
+++ b/eventsources/sources/gcppubsub/start.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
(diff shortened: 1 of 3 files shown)


eventsources/sources/gcppubsub/start.go
```diff
@@ -233,23 +234,28 @@ func (el *EventListener) prepareSubscription(ctx context.Context, logger *zap.Su
...
-		case codes.OK:
-			client.Close()
-			return nil, nil, errors.Errorf("you lack permission to pull from %s", subscription)
-		case codes.NotFound:
-			// OK, maybe the subscription doesn't exist yet, so create it later
-			// (it possibly means project itself doesn't exist, but it's ok because we'll see an error later in such case)
-		default:
-			client.Close()
-			return nil, nil, errors.Wrapf(err, "failed to test permission for subscription %s", subscription)
+	subscExists := false
```
whynowy, line 237: Will this always being `false` in the emulator case matter?

whynowy (commented):
Please also add something to mention this in https://github.com/argoproj/argo-events/blob/master/docs/eventsources/gcp-pubsub.md.

eventsources/sources/gcppubsub/start.go
avanderveen: Right. This should probably still check if the subscription exists.

I've just added that check in [the latest commit](https://github.com/argoproj/argo-events/pull/1343/commits/a4827aad3331cf3d2f0f82aae5fa00769070933b).

avanderveen:
> Please also add something to mention this in https://github.com/argoproj/argo-events/blob/master/docs/eventsources/gcp-pubsub.md.

Updated the docs in [6389819](https://github.com/argoproj/argo-events/pull/1343/commits/6389819d5659f18314d6c720193d243516ad50f3).

whynowy (approved):
LGTM, thanks for your contribution!

PR #1344, opened 2021-09-10
docs: Add missing webhook secret key used for GitHub EventSource

terrytangyuan:
Signed-off-by: Yuan Tang <terrytangyuan@gmail.com>

### DIFF
--- a/docs/eventsources/setup/github.md
+++ b/docs/eventsources/setup/github.md
@@ -32,11 +32,11 @@ Example event-source yaml file is [here](https://github.com/argoproj/argo-events
 1. Create an API token if you don't have one. Follow [instructions](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) to create a new GitHub API Token.
    Grant it the `repo_hook` permissions. 
 
-1. Base64 encode your api token key.
+1. Base64 encode your API token key.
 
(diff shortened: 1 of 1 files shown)


examples/event-sources/github.yaml
```diff
@@ -49,7 +49,7 @@ spec:
 #        # Name of the K8s secret that contains the hook secret
 #        name: github-access
 #        # Key within the K8s secret whose corresponding value (must be base64 encoded) is hook secret
-#        key: secret
+#        key: token
```
whynowy, line 52: Why?

terrytangyuan: Because the `github-access` Secret only has `token` as the key. See step 3 in this document: https://argoproj.github.io/argo-events/eventsources/setup/github/

`webhookSecret` having the incorrect key leads to errors in the event source pod (something like cannot find the `argo-events/secret/github-access`).

whynowy: I see. In that case, we should change the secret to contain a `secret` key, `apiToken` is the token to access github api, `webhookSecret` is the secret used for github webhook configuration, they are different.

terrytangyuan: Sounds good. Just updated the PR to add that instead.

whynowy (approved):
Thanks!

PR #1376, opened 2021-10-13
fix: (docs) update an example sensor to include group, resource, and version keys

canthonyscott:
This adds the following fields to the example yaml file
* group: argoproj.io
* version: v1alpha1
* resourceL workflows

These additions were necessary for Argo events to be able to create the workflow using the operate-workflow-sa service account that was setup following the examples within the Argo Events docs. Without them I received kubernetes permission errors on the sensor pod logs.


canthonyscott:
Oh this is great to know! I will be upgrading. Thanks! I will close this PR

PR #1394, opened 2021-10-25
feat: Sensor/AWSLambdaTrigger: allow using AWS IAM Role specified in Service Account

kromanow94:
### Add possibility to use AWSLambdaTrigger without access and secret keys.

---

This is useful when using ServiceAccount with AWS IAM Role assigned.

I've updated unit tests to cover 2 scenarios:
- minimalistic AWSLambdaTrigger def without aws credentials
- full AWSLambdaTrigger def with credentials, roleARN, payload

---

Checklist:

* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md). **(N/A)**


kromanow94:
I see this didn't make to latest release. Any ETA on adding this to release?

whynowy:
This will go with next patch release, which will happen in around 1-2 weeks.

24601:
It seems this isn't in 1.5.4, @whynowy - any idea on when you think it'll make it into a release?

whynowy:
@24601 @kromanow94 - sorry about that, I remember I had ever committed to include one PR when I was releasing v1.5.4, but wasn't able to figure out which one was that - will patch  v1.5.5 by the end of this week!

Issue #1308, opened 2021-08-11
GitHub EventSource should be able to create webhooks at GitHub organization level.


YesemKebede:
For use cases that involve listening in for organizational events such as repository creation and deletion, there is no good way to create webhooks from GitHub EventSource. It looks like GitHub EventSource definition expects repository name to be defined in order to create a webhook. It would be awesome if Github EventSource also supports creating webhooks at organization level.

As a workaround I tried setting up a generic webhook and it works but starts failing when I add authSectret. See  [#1289 ](https://github.com/argoproj/argo-events/issues/1289)for detail

daniel-codefresh:
I can take a look at this one, @whynowy can you assign it to me please?

whynowy:
Of course! @daniel-codefresh

daniel-codefresh:
@whynowy Heres the related PR: #1412

PR #1412, opened 2021-11-05
feat: Added org level webhook support for github es

daniel-codefresh:
Checklist:


daniel-codefresh:
Related issue: #1308

daniel-codefresh:
@whynowy Do you think we should have an additional validation to block the usage of Repositories and Organizations fields simoatainisly in a single eventsource? One thing that bothers me about that is that both will share the same Events field, but there are few github events that are only supported on Orgs level and not on Repos level, so potentially a user might have both repos and orgs on the same eventsource with an org only event which will cause a failure while creating the repo hooks... So should we create a clear separation between the two? WDYT?

whynowy:
> @whynowy Do you think we should have an additional validation to block the usage of Repositories and Organizations fields simoatainisly in a single eventsource? One thing that bothers me about that is that both will share the same Events field, but there are few github events that are only supported on Orgs level and not on Repos level, so potentially a user might have both repos and orgs on the same eventsource with an org only event which will cause a failure while creating the repo hooks... So should we create a clear separation between the two? WDYT?

To make it simple, I think it's ok to only allow one of them configured.

daniel-codefresh:
@whynowy Ok great, I've added another validation to cover it

whynowy (approved):
LGTM, high quality as always!

whynowy:
Some test cases need to be fixed.

daniel-codefresh:
@whynowy Done:)

PR #1420, opened 2021-11-09
docs: Add example service that selects event source pods with selector

terrytangyuan:
Currently, there's only documentation on the use of `service` in event source spec. Adding this example so that users don't have to search for the selector key for event source pods.

Signed-off-by: Yuan Tang <terrytangyuan@gmail.com>

Checklist:


whynowy (approved):
Thanks for refining the doc, that looks better!

Issue #1366, opened 2021-09-30
Template parameters empty for multi dependencies sensors


dan1el-k:
**Describe the bug**
When using multiple dependencies with "or" condition in a Sensor Object, template parameters get only resolved if they come as a last block in the "parameters" section. Meaning, if each dependency has it's own template parameters, bounded to the respective dependency, one will work (the later block), one will not work.

**To Reproduce**
1. Create 2 dependency in a Sensor Object
2. Create for each dependency a template parameter
3. Trigger the sensor
4. Check the logs of the Sensor object and see, that only 1 dependency has values resolved, the other not.
5. Re-order the Parameters, to put this time the other dependency parameter at the top.
6. Trigger the sensor
7. Check the logs of the Sensor object and see this time, that the other dependency object has values resolved.

**Expected behavior**
All dependency object which are called based on the condition, should have there template parameters be correctly resolved, independently of the order in where they are put.

**Screenshots**
![image](https://user-images.githubusercontent.com/34611072/135529806-e6bf79fd-8ecb-4a1a-9ac4-d8597a729482.png)

**Environment (please complete the following information):**
 - Kubernetes: v1.19 (Openshift 4.6)
 - Argo: v.3.0.7
 - Argo Events: v1.4.0

**Additional context**
Sensor Object used:

```apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: ref-opened
  namespace: argo
spec:
  template:
    serviceAccountName: argo
  dependencies:
  - name: manifest-create-branch
    eventSourceName: github-pipeline
    eventName: manifest-ref
    filters:
      data:
      - path: body.X-GitHub-Event # Eventsource already filters for create but just in case ...
        type: string
        value:
        - create
      - path: body.ref_type
        type: string
        value:
        - branch
      - path: body.master_branch
        type: string
        value:
        - master
  - name: application-create-branch
    eventSourceName: github-pipeline
    eventName: application-ref
    filters:
      data:
      - path: body.X-GitHub-Event # Eventsource already filters for create but just in case ...
        type: string
        value:
        - create
      - path: body.ref_type
        type: string
        value:
        - branch
      - path: body.master_branch
        type: string
        value:
        - master
  triggers:
  - template:
      conditions: application-create-branch || manifest-create-branch
      name: github-workflow-trigger
      argoWorkflow:
        operation: submit
        source:
          resource:
            apiVersion: argoproj.io/v1alpha1
            kind: Workflow
            metadata:
[...]

PR #1433, opened 2021-11-17
fix: Eval of params with missing src dep and no default value

daniel-codefresh:

Checklist:

### DIFF
--- a/sensors/triggers/params.go
+++ b/sensors/triggers/params.go
@@ -24,7 +24,6 @@ import (
 
 	sprig "github.com/Masterminds/sprig/v3"
 	"github.com/ghodss/yaml"
-	"github.com/pkg/errors"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


daniel-codefresh:
Fixes: #1366

sensors/triggers/params.go
```diff
@@ -159,60 +162,71 @@ func renderEventDataAsJSON(event *v1alpha1.Event) ([]byte, error) {
...
-			return *src.Value, nil
+			return src.Value, nil
 		}
-		return string(value), nil
+
+		strPayload := string(eventPayload)
+		return &strPayload, nil
 	}
-	return "", errors.Wrapf(err, "unable to resolve '%s' parameter value", src.DependencyName)
+	return nil, errors.Wrapf(err, "unable to resolve '%s' parameter value", src.DependencyName)
```
whynowy, line 229: It seems `err` is always `nil`, shall this be changed to:

```
return nil, fmt.Errorf("unable to resolve '%s' parameter value", src.DependencyName)
```

daniel-codefresh: Yes 100%, just forgot to change this line

whynowy (approved):
LGTM, thanks!

Issue #1432, opened 2021-11-17
cannot create k8s resource ingress


coawlion:
**Describe the bug**
```
- template:
        conditions: "create-editorx-dep"
        name: ingress-creation
        k8s:
          operation: create
          source:
            resource:
              apiVersion: networking.k8s.io/v1beta1
              kind: Ingress
              metadata:
                name: editorx-ingress-coawlion
                namespace: editorx
...
```
When using the above sensor trigger, it failed to create ingress.

```
{"level":"error","ts":1637112348.0356252,"logger":"argo-events.sensor","caller":"sensors/listener.go:302","msg":"failed to execute a trigger","sensorName":"webhook","error":"failed to execute trigger: timed out waiting for the condition: ingresss.networking.k8s.io is forbidden: User \"system:serviceaccount:argo-events:editorx-k8s-resource-sa\" cannot create resource \"ingresss\" in API group \"networking.k8s.io\" in the namespace \"editorx\"","errorVerbose":"timed out waiting for the condition: ingresss.networking.k8s.io is forbidden: User \"system:serviceaccount:argo-events:editorx-k8s-resource-sa\" cannot create resource \"ingresss\" in API group \"networking.k8s.io\" in the namespace \"editorx\"\nfailed to execute trigger\ngithub.com/argoproj/argo-events/sensors.(*SensorContext).triggerOne\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:355\ngithub.com/argoproj/argo-events/sensors.(*SensorContext).triggerWithRateLimit\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:300\nruntime.goexit\n\t/opt/hostedtoolcache/go/1.17.1/x64/src/runtime/asm_amd64.s:1581","triggerName":"ingress-creation","triggeredBy":["create-editorx-dep"],"triggeredByEvents":["65633634313066362d303134372d343733322d393963392d353636633636313836393538"],"stacktrace":"github.com/argoproj/argo-events/sensors.(*SensorContext).triggerWithRateLimit\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:302"}

```

Notice that, it tries to create resource `ingresss` instead of `ingresses`

**To Reproduce**
1. Use webhook trigger as event-source
2. Trigger k8s resource ingress

**Expected behavior**
Ingress created as a result of trigger

**Environment (please complete the following information):**
 - Kubernetes: v1.20.6
 - Argo: NA
[...]

PR #1434, opened 2021-11-17
fix: Creation of ingress resource with k8s trigger

daniel-codefresh:

Checklist:

### DIFF
--- a/sensors/triggers/triggers.go
+++ b/sensors/triggers/triggers.go
@@ -17,21 +17,19 @@ limitations under the License.
 package triggers
 
 import (
-	"strings"
-
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/gengo/namer"
+	"k8s.io/gengo/types"
 )
 
 func GetGroupVersionResource(obj *unstructured.Unstructured) schema.GroupVersionResource {
 	gvk := obj.GroupVersionKind()
-	lowercaseKind := strings.ToLower(gvk.Kind)
-	var resource string
-	if strings.HasSuffix(lowercaseKind, "s") {
-		resource = lowercaseKind + "es"
-	} else {
-		resource = lowercaseKind + "s"
-	}
+	resource := namer.NewAllLowercasePluralNamer(nil).Name(&types.Type{
+		Name: types.Name{
+			Name: gvk.Kind,
+		},
+	})
 
 	return schema.GroupVersionResource{
 		Group:    gvk.Group,
--- a/sensors/triggers/triggers.go
+++ b/sensors/triggers/triggers.go
@@ -17,7 +17,6 @@ limitations under the License.
 package triggers
 
 import (
-	"fmt"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -26,9 +25,17 @@ import (
 
 func GetGroupVersionResource(obj *unstructured.Unstructured) schema.GroupVersionResource {
 	gvk := obj.GroupVersionKind()
+	lowercaseKind := strings.ToLower(gvk.Kind)
+	var resource string
+	if strings.HasSuffix(lowercaseKind, "s") {
+		resource = lowercaseKind + "es"
+	} else {
+		resource = lowercaseKind + "s"
+	}
+
 	return schema.GroupVersionResource{
 		Group:    gvk.Group,
 		Version:  gvk.Version,
-		Resource: fmt.Sprintf("%ss", strings.ToLower(gvk.Kind)),
+		Resource: resource,
 	}
 }
--- a/go.mod
+++ b/go.mod
@@ -71,7 +71,7 @@ require (
 	k8s.io/apimachinery v0.21.1
 	k8s.io/client-go v0.21.1
 	k8s.io/code-generator v0.20.4
-	k8s.io/gengo v0.0.0-20201113003025-83324d819ded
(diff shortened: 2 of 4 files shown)


daniel-codefresh:
Fixes: #1432

sensors/triggers/triggers.go
```diff
@@ -26,9 +25,17 @@ import (
 
 func GetGroupVersionResource(obj *unstructured.Unstructured) schema.GroupVersionResource {
 	gvk := obj.GroupVersionKind()
+	lowercaseKind := strings.ToLower(gvk.Kind)
```
whynowy, line 28: Can we use this?

https://github.com/kubernetes/gengo/blob/master/namer/plural_namer.go#L54

daniel-codefresh: Yes, good point, changing now

whynowy (approved):
Excellent!

daniel-codefresh:
@whynowy BTW, I just noticed we are using `eventbus` as the plural form of the EventBus CRD, is there any particular reason for this? It got me thinking that maybe we should have an optional `resource` field after all
https://github.com/argoproj/argo-events/blob/7ad602783e590898909ac4c6682c5a66c6b21f0e/hack/crds.go#L32

whynowy:
> @whynowy BTW, I just noticed we are using `eventbus` as the plural form of the EventBus CRD, is there any particular reason for this? It got me thinking that maybe we should have an optional `resource` field after all
> 
> https://github.com/argoproj/argo-events/blob/7ad602783e590898909ac4c6682c5a66c6b21f0e/hack/crds.go#L32

Per @alexec (who is from England) - plural of `bus` is still `bus`...

daniel-codefresh:
@whynowy Oh I see.. Well I'm certainly not an English expert but from a quick Google search, it seems like `buses` is also an acceptable plural form of `bus`:
https://www.oxfordlearnersdictionaries.com/definition/english/bus_1
https://www.collinsdictionary.com/dictionary/english/bus
https://www.merriam-webster.com/words-at-play/plural-of-bus
And more importantly, I just verified that the `k8s.io/gengo/namer` library that was used in this PR is actually returning `eventbuses` as the plural form of `eventbus` (which will cause a failure if a user tries to CRUD an `eventbus` through a k8s-trigger), so I guess we would have to support that by one way or another... WDYT?

whynowy:
> @whynowy Oh I see.. Well I'm certainly not an English expert but from a quick Google search, it seems like `buses` is also an acceptable plural form of `bus`: https://www.oxfordlearnersdictionaries.com/definition/english/bus_1 https://www.collinsdictionary.com/dictionary/english/bus https://www.merriam-webster.com/words-at-play/plural-of-bus And more importantly, I just verified that the `k8s.io/gengo/namer` library that was used in this PR is actually returning `eventbuses` as the plural form of `eventbus` (which will cause a failure if a user tries to CRUD an `eventbus` through a k8s-trigger), so I guess we would have to support that by one way or another... WDYT?

I think it would be weird to support both - an extra CRD for `buses`. I would recommend to keep it as is. If there's really somebody want to use that to CRUD an `eventbus`, we should add an exception when using that lib. I remember `code-nenerator` also used to support plural `exceptions`, now sure why it stopped working, and I added that workaround in the shell.

daniel-codefresh:
@whynowy Yes, definitely, an extra CRD would be weird. I actually meant either adding an exception like you suggested, or changing the plural form of the existing CRD back to `eventbuses` so we can stick to the conventions and also get rid of that extra logic that was added to the codegen process, of transforming `eventbuses` to `eventbus`

daniel-codefresh:
> > @whynowy Oh I see.. Well I'm certainly not an English expert but from a quick Google search, it seems like `buses` is also an acceptable plural form of `bus`: https://www.oxfordlearnersdictionaries.com/definition/english/bus_1 https://www.collinsdictionary.com/dictionary/english/bus https://www.merriam-webster.com/words-at-play/plural-of-bus And more importantly, I just verified that the `k8s.io/gengo/namer` library that was used in this PR is actually returning `eventbuses` as the plural form of `eventbus` (which will cause a failure if a user tries to CRUD an `eventbus` through a k8s-trigger), so I guess we would have to support that by one way or another... WDYT?
> 
> I think it would be weird to support both - an extra CRD for `buses`. I would recommend to keep it as is. If there's really somebody want to use that to CRUD an `eventbus`, we should add an exception when using that lib. I remember `code-nenerator` used to support plural `exceptions`, now sure why it stopped working, and I added that workaround in the shell.

@whynowy Yes, definitely, an extra CRD would be weird. I actually meant either adding an exception like you suggested, or changing the plural form of the existing CRD back to eventbuses so we can stick to the conventions and also get rid of that extra logic that was added to the codegen process, of transforming eventbuses to eventbus

whynowy:
@daniel-codefresh - I prefer to add exceptions, changing CRD definition sometimes brings weird issues during upgrade. Code-gen used to support [exceptions](https://github.com/argoproj/argo-events/blob/v1.3.1/hack/update-codegen.sh#L41), not sure why it is not supporting now. I didn't spent time on the research.

daniel-codefresh:
> @daniel-codefresh - I prefer to add exceptions, changing CRD definition sometimes brings weird issues during upgrade. Code-gen used to support [exceptions](https://github.com/argoproj/argo-events/blob/v1.3.1/hack/update-codegen.sh#L41), not sure why it is not supporting now. I didn't spent time on the research.

@whynowy Oh yea, now I see what you meant by codegen exceptions.. I was thinking about using a plural exception for `eventbus` together with `k8s.io/gengo/namer ` lib so we can be covered in case a user wants to CRUD eventbus resource as well. Heres the change I was talking about: #1440

PR #1455, opened 2021-12-09
fix: Ack handled messages in pulsar eventsource

arnarg:
Currently consumed messages are not acknowledged by argo-events so pulsar keeps them around. This results in the whole queue flooding in when the pulsar eventsource is restarted.
Consumers should acknowledge all messages consumed. https://pulsar.apache.org/docs/en/concepts-messaging/#acknowledgement

This pull request sends acknowledgement to pulsar if the eventsource could handle the message successfully and send it to the eventbus.

Checklist:


whynowy (approved):
Thanks for fixing it!

PR #1468, opened 2021-12-14
docs: fix metrics names for event source

mweibel:
all names are singular for the event source.


mweibel:
nevermind :wink: all good with the names!

PR #1484, opened 2021-12-22
feat: introducing namespaced flag for admission webhooks

tkennes:
Signed-off-by: Tom Kennes <tomkennes@skyworkz.nl>

Checklist:


webhook/webhook.go
```diff
@@ -125,8 +128,15 @@ func (ac *AdmissionController) Run(ctx context.Context) error {
 // Register registers the validating admission webhook
 func (ac *AdmissionController) register(
 	ctx context.Context, client clientadmissionregistrationv1.ValidatingWebhookConfigurationInterface, caCert []byte) error {
+	var scope admissionregistrationv1.ScopeType
 	failurePolicy := admissionregistrationv1.Ignore
 
+	if ac.Options.Namespaced {
```
whynowy, line 134: I don't think `scope` is used to specify if it's namespaced installation, but if it's namespaced resource. For example `namespace` is a cluster scoped resource, `pod` is a namespace scoped one.

To do namespace scoped installation for the validating webhook:

1. Use `namespaceSelector` in `ValidatingWebhookConfiguration` to match the installation namespace;
2. Check if there's a `ValidatingWebhookConfiguration` existing, if yes, add an entry to `webhooks`, otherwise create a new one.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector

tkennes: You are right, and this made me realize that the ValidatingWebhook is not a namespaced resource anyway, so this would still mean that you need to configure cluster-wide resources if you want to set up a fully namespaced version of argo events.

Nonetheless, I have attempted to make those changes. Let me know what you think!

whynowy: @tkennes - I think more work is needed to get this done. Is this very important for you?

tkennes: Agreed. It is not very important to me at the moment. Also, thinking about it again, it might perhaps be easier to expand the documentation and cover a case with a Validating Webhook which applies only to resources in a specific namespace, rather than incorporate this like this.

whynowy: 👍 Appreciate if you could do that.

If you don't mind, I'd like to close this PR.

Issue #1449, opened 2021-12-05
Filter issues


whynowy:
1. the [Filter documentation](https://github.com/argoproj/argo-events/blob/master/docs/tutorials/07-filters.md#multiple-paths):
    - is really incomplete
    - examples are too simple (it would be good to have a real example based for example on GitHub and GitLab events)
    - should have a dedicated section and not be placed only under "tutorials", as imo it's a really important feature of Argo Events
2. I suggest to add an option to change the evaluation process of the list of filters and conditions under each filter, offering possibility to specify if evaluated in `AND` or in `OR`, because:
    - there is no option to have multiple conditions under a filter evaluated in `OR`. I'm forced to create multiple sensors, increasing complexity (more yaml manifests) and costs of maintainability and resources in the cluster (more sensors = more resource consumption)
    - `Multiple paths` could fast become really complex and it's completely not flexible as `Expr` (see examples of differences below)
    - declaring or-conditions in `Multiple paths` is complex and limited, because based on regex

The Filter doc is really incomplete because:
- Nowhere is documented that "the list of filters is evaluated in AND condition"
- Nowhere is documented that "the list of conditions under a filter is evaluated in AND condition"
- There are **4** types of filter, not 3. `Expr filter` is completely missing and it's one of the most important and flexible
- Data filter Multiple path example is not really clear and probably wrong
- There is no clear example for Context filter (like for Data filter)
- There is no clear example for Time filter (like for Data filter)
- There is no real examples or use case, for example filtering an event coming from GitHub or GitLab webhook

"Data filter Multiple paths" is not comparable to "Expr filter", for example:
- declaring multiple conditions within a regex becomes fast really complex and error prone
- declaring something like `b != ""` is not so easy together with other conditions
- it's not possible to declare conditions in `||` (OR)
- it's not possible to declare a condition like `b > 2.5` as number or `b == true` as bool
- no "Parenthesis to control order of evaluation" available (based on https://github.com/Knetic/govaluate)
- no "Arrays" available (based on https://github.com/Knetic/govaluate)
For these reasons (and others I didn't list), I don't consider "Data filter Multiple paths" a valid alternative to "Expr filter". And for sure also not a valid, easy, fast alternative to have conditions under a filter evaluated in `OR`.

@bygui86

bygui86:
@whynowy great issue. This is mainly to improve the documentation of one of the most important feature of ArgoEvents in my opinion :)

bygui86:
[...]

PR #1490, opened 2021-12-30
feat: Improved Sensors filtering feature (aka Sensors filtering v2)

bygui86:
**Description**
As discussed with @whynowy recently in a call, this is a PR to improve Sensors filtering as described in issue #1449 .
As in my opinion this feature is one of the most important but not properly enhanced (especially in the documentation, just mentioned in a tutorial), I worked to improve it.

**List of changes and improvements**

I added a new field named `FiltersLogicalOperator` of type `LogicalOperator` on `EventDependency` in `pkg/apis/sensor/v1alpha1/types.go`. FiltersLogicalOperator defines how different filters are evaluated together, for example: time AND data, data OR expr, etc. FiltersLogicalOperator available values are `and` (&&) and `or` (||). FiltersLogicalOperator is optional and if left blank treated as `and` (&&).

I added a new field named `LogicalOperator` of type `LogicalOperator` on `EventDependencyFilter` in `pkg/apis/sensor/v1alpha1/types.go`. LogicalOperator defines how multiple Data filters (if defined) or Exprs filters (if defined) are evaluated together, for example: data filter #1 AND data filter #2, expr filter #1 OR expr filter #2. LogicalOperator available values are `and` (&&) and `or` (||). LogicalOperator is optional and if left blank treated as `and` (&&).

To summarise: FiltersLogicalOperator is valid across different filter types (time, ctx, data, expr), LogicalOperator is valid across different internal filters (between data filters or between expr filters).

So the Sensor is able to evaluate filters in 2 ways:
- all must be TRUE to let the event pass (AND)
- either one must be TRUE to let the event pass (mutual exclusive, OR)

I tried to improve filter testing to cover as many use cases as possible, splitting tests in different files to avoid end up having a file of 4k lines.

Moreover I added some examples in `examples/sensors/filter-*` to help a better understanding of filters in general and new features as well.

In a second PR I will also improve and enhance Sensors filtering related documentation ;) waiting for feedback and discussion I will work on this.

Looking forward your feedback!!

Checklist:
- [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).


bygui86:
@whynowy fixed tests, please allow pipeline

VaibhavPage:
@bygui86  A more flexible way would be to define a circuit of filters.

For filters of the same type, user can optionally name the filters as below,
```
logicalOperator:  filter1 || (filter2 && filter3)
data:
  - path: "a"
    type: "bool"
    value:
      - "true"
    name: filter1 # optional, should only be defined if logicalOperator is defined.
  - path: "b.c"
    type: "number"
    value:
      - "3.14"
    name: filter2
  - path: "b.d"
    type: "string"
    value:
      - "hello,world"
      - "hello, world"
    name: filter3
```

For filters of different types, the `filtersLogicalOperator` can have a value like `data || expr || context`.

bygui86:
@VaibhavPage thanks for the suggestion! That's a good idea and I would keep it for "filtering v3".

The missing of base flexibility (basically not allowing filters in "OR") is limiting a lot the usage of the filtering feature.
So in this PR I started adding some flexibility and improving testing without completely changing everything and causing a disruption (following semantic versioning: +1 MINOR).

Trying to avoid disruption, I didn't change the structure of the filter itself. In fact `logicalOperator` is valid for `both data and expr filters` (if specified).
What you suggest `logicalOperator:  filter1 || (filter2 && filter3)` is really good, but requires a disruptive change in the structure of `filter.data` and `filter.expr` replacing a simple list of filters with a slightly more complex struct like:
```yaml
  filtersLogicalOperator: data || expr   # optional, default to ALL AND
  filters:
    exprs:
      logicalOperator: exprFactor1 || exprFactor2   # optional, default to ALL AND
      factors:
        - id: exprFactor1   # optional, default to empty
          expr: a == "b" || c == 10
          fields:
            - name: a
              path: a
            - name: c
              path: c
        - id: exprFactor2
          expr: e == false
          fields:
            - name: e
              path: d.e
    data:
      logicalOperator: dataFactor1 && (dataFactor2 || dataFactor3)
      factors:
        - id: dataFactor1
          path: "a"
          type: "string"
          value:
            - "b"
        - id: dataFactor2
          path: "c"
          type: "number"
          comparator: ">"
          value:
            - "20"
        - id: dataFactor3
          path: "d.e"
          type: "bool"
          value:
            - "false"
```

I suggest to evolve filtering to that level in a second step, (following semantic versioning: +1 MAJOR) maybe argo-events v2.0.0?

bygui86:
@VaibhavPage if you want to run a further and deeper discussion, you can find me on slack in CNCF space ;)

bygui86:
@VaibhavPage tests are fine and PR is ready for detailed review.
CI pipe fails on `Ensure nothing changed` but I don't know if and how I can fix it.

VaibhavPage:
@bygui86 LGTM, mind fixing the merge conflicts?

bygui86:
@VaibhavPage of course no problem, I will fix conflicts

bygui86:
@VaibhavPage conflicts fixed, please let the CI go through :)

VaibhavPage:
@bygui86 pipelines are still failing

bygui86:
@VaibhavPage I'm sorry :( locally I had no problem before committing all conflicts fixes.

The issue is in the `go.sum`:
`github.com/fasthttp/websocket v1.4.2 h1:hTnl0DRQ9Tj4uVc4QcZCGp+mgF5mVKnijR1eNyn2waY=` this line works locally
`github.com/fasthttp/websocket v1.4.2 h1:AU/zSiIIAuJjBMf5o+vO0syGOnEfvZRu40xIhW/3RuM=` this line works in GitHub Actions

I don't know which library uses `fasthttp/websocket` behind the scenes but googling I found that the maintainer republished version 1.4.2 many times, that's why there is this issue...

I reverted the references that seems to work on GitHub Actions. If this won't work, I need your help to solve this...

fentonfentonfenton:
I have a conundrum here:

Given block:

```
{ "name": "jack",
"name": "jill"}
```
Is it possible to do an event filter that says 

```
path: #.name
does_not_include: bob
```

I.e. the event is valid if 'bob' is not in the array of names?

If I can't do this, is it possible to add this feature?

Thanks.

bygui86:
@fentonfentonfenton if I understand correctly your question, you can already do that sort of filtering using `Expr filter`.
Unfortunately this is not documented, but after this PR I will improve also the documentation including it.

fentonfentonfenton:
> @fentonfentonfenton if I understand correctly your question, you can already do that sort of filtering using `Expr filter`. Unfortunately this is not documented, but after this PR I will improve also the documentation including it.

I'll try some things in `expr` and if I get a solution will pass an example back to you.


What might work, too, is if you add the logical operator NOT, if that's possible.

thanks for responding!

bygui86:
> What might work, too, is if you add the logical operator NOT, if that's possible.

This is already offered by `expr filter` itself.
This PR is to extend the way different types of filters interact together, for example: time AND data, ctx OR expr. In the current version, only AND is supported.
This PR is also to extend the way different value in data or expr filter interact together, for example: data-filter-1 AND data-filter-2, data-filter-1 OR data-filter-2.

If this is not what you are searching for, please open another issue so you can make a better example and we can discuss about it.

bygui86:
@whynowy I updated the PR according to what discussed in our call, splitting logicalOperator in dataLogicalOperator and exprLogicalOperator and adding some further tests to verify there are no disruptive changes.

Let me know :)

whynowy (approved):
LGTM, thanks!

bygui86:
@whynowy great!

In the next days I will improve the documentation about filtering!! Please expect another PR about this :)

fentonfentonfenton:
Thanks @bygui86 . great work!

If you want any help from a user when you submit the documentation PR, pls tag me! cheers

whynowy:
Thank both of you guys! @bygui86 @fentonfentonfenton

Issue #1426, opened 2021-11-12
Support for bitbucket cloud as an eventsource


RixTmobilender:
**Is your feature request related to a problem? Please describe.**
There is currently no first-class support for bitbucket **cloud** as an event source. Only for the [soon to be deprecated](https://www.atlassian.com/migration/assess/journey-to-cloud) **server** version.

**Describe the solution you'd like**
First-class support of bitbucket cloud as an eventsource.

**Describe alternatives you've considered**
Implementing the generic "webhook" event source but I'd like to avoid this as I'd require to use an external tool (probably terraform) to be able to automatically manage webhooks as opposed to from within argo events, which would be ideal.

**Additional context**
[Bitbucket cloud's REST API](https://developer.atlassian.com/cloud/bitbucket/rest/intro/)
There is a stale PR on this subject that never got merged https://github.com/argoproj/argo-events/pull/917

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
We already have bitbucket server eventsource but missing documents for it, just created an issue for [tracking](https://github.com/argoproj/argo-events/issues/1443).

Would you like to contribute on the bitbucket cloud eventsource? @RixTmobilender

daniel-codefresh:
@whynowy I'll do it  👍  I'm assigning this one to myself

daniel-codefresh:
Related PR: #1493

PR #1493, opened 2022-01-02
Bitbucket Cloud Eventsource

daniel-codefresh:
Added new eventsource for Bitbucket Cloud

Checklist:


daniel-codefresh:
Related issue: #1426

common/util.go
```diff
@@ -404,3 +404,47 @@ func uniqueVolumeMounts(mounts []v1.VolumeMount) []v1.VolumeMount {
 	}
 	return rMounts
 }
+
+// AreSlicesEqual returns true if the two provided string slices are equal.
+func AreSlicesEqual(first []string, second []string) bool {
+	if len(first) == 0 && len(second) == 0 {
```
VaibhavPage, line 410: Can we not sort and apply reflect.DeepEqual?

common/util.go
```diff
@@ -404,3 +404,47 @@ func uniqueVolumeMounts(mounts []v1.VolumeMount) []v1.VolumeMount {
...
+
+	tmp := make(map[string]int)
+	for _, str := range first {
+		tmp[str] = 1
+	}
+
+	for _, str := range second {
+		v, ok := tmp[str]
+		if !ok || v == -1 {
+			tmp[str] = -1
```
VaibhavPage, line 425: we can return false here.

eventsources/sources/bitbucket/basic_auth_strategy.go
```diff
@@ -0,0 +1,51 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bitbucket
+
+import (
+	bitbucketv2 "github.com/ktrysmt/go-bitbucket"
```
VaibhavPage, line 19: lets run go import to organize imports for all files

eventsources/sources/bitbucket/start.go
```diff
@@ -0,0 +1,344 @@
...
+			logger.Errorw("failed to update existing webhook", zap.Error(err))
+			return errors.Wrap(err, "failed to update existing webhook")
+		}
+
+		logger.Info("successfully updated the existing webhook")
+		return nil
+	}
+
+	logger.Info("webhook doesn't exist yet, creating a new webhook...")
+	newWebhook, err := router.createWebhook(formattedWebhookURL)
```
VaibhavPage, line 234: shouldn't we check for just `if isFound` before creating a new one?

eventsources/sources/bitbucket/start.go
```diff
@@ -0,0 +1,344 @@
...
+	return err
+}
+
+// listWebhooks gets a list of all existing webhooks in target repository
+func (router *Router) listWebhooks() ([]WebhookSubscription, error) {
+	es := router.bitbucketEventSource
+	hooksResponse, err := router.client.Repositories.Webhooks.Gets(&bitbucketv2.WebhooksOptions{
+		Owner:    es.Owner,
+		RepoSlug: es.RepositorySlug,
+		Active:   true,
```
VaibhavPage, line 295: Should we not list all webhooks, active or not? If someone manually deactivates the webhook registered on the BitBucket and the event source has `deleteOnFinish` set to `false`, then the code will try to register it as a fresh webhook when eventsource pod restarts. Instead, we just need to update the existing one.

eventsources/sources/bitbucket/start.go
```diff
@@ -0,0 +1,344 @@
...
+// findWebhook searches for a webhook in a list by its URL and returns the webhook if its found
+func (router *Router) findWebhook(hooks []WebhookSubscription, targetWebhookURL string) (*WebhookSubscription, bool) {
+	var existingHookSubscription *WebhookSubscription
+	isFound := false
+	for _, hook := range hooks {
+		if hook.Url == targetWebhookURL {
+			isFound = true
+			existingHookSubscription = &hook
+			router.hookID = hook.Uuid
+			break
```
VaibhavPage, line 332: Instead of `break` and defining variables `existingHookSubscription` and `isFound`, we can just return &WebhookSubscription{}, true on 332 and nil, false on 336.

Also, what's the purpose of `isFound`? can the caller not just check whether the pointer to WebhookSubscription is nil or not?

eventsources/sources/bitbucket/types.go
```diff
@@ -0,0 +1,81 @@
...
+
+type WebhookSubscription struct {
+	// Uuid holds the webhook's ID
+	Uuid string `json:"uuid,omitempty"`
+	// The Url events get delivered to.
+	Url string `json:"url,omitempty"`
+	// Description holds a user-defined description of the webhook.
+	Description string `json:"description,omitempty"`
+	// Subject holds metadata about the subject of the webhook (repository, etc.)
+	Subject *interface{} `json:"subject,omitempty"`
```
VaibhavPage, line 70: should the type for Subject be just interface{}?

eventsources/sources/bitbucket/types.go
```diff
@@ -0,0 +1,81 @@
...
+	client *bitbucketv2.Client
+	// hookID holds the the uuid of the webhook that is created in Bitbucket
+	hookID string
+	// bitbucketEventSource is the event source that holds information to consume events from Bitbucket
+	bitbucketEventSource *v1alpha1.BitbucketEventSource
+}
+
+type WebhookSubscription struct {
+	// Uuid holds the webhook's ID
+	Uuid string `json:"uuid,omitempty"`
```
VaibhavPage, line 64: Are all of these fields optional? I'd assume fields like Uuid to be required, if so, we can remove omitempty tags.

VaibhavPage (changes requested):

common/util.go
daniel-codefresh: No, the slices must be sorted if we want to apply `reflect.DeepEqual`. Also, as far as I recall `reflect.DeepEqual` is much slower than a loop in this case.

common/util.go
daniel-codefresh: Actually I didn't write that logic, I only moved this function to a more general place (from github eventsource utils) so it can be reused in more places, so I didn't want to introduce any unnecessary changes there. But anyway - sure, its NP I can change that.

eventsources/sources/bitbucket/basic_auth_strategy.go
daniel-codefresh: Hmm not sure I follow.. seems pretty organized to me, or am I missing something?

eventsources/sources/bitbucket/start.go
daniel-codefresh: Yes 100% good catch, thanks! I actually added the `shouldUpdateWebhook` check only later and forgot to handle this case.
Pushing a fix right now.

eventsources/sources/bitbucket/start.go
daniel-codefresh: Yes, totally right. Actually the `Gets` method only uses `Owner` and `RepoSlug` for the query, I just wanted to test something and forgot to remove it.
Fixing now.

eventsources/sources/bitbucket/start.go
daniel-codefresh: Yes of course you could replace the break, I just wanted to avoid an extra `return` statement if it's unnecessary, and thought it would be more readable this way. Same regarding `isFound` - just thought it would be more readable than comparing the pointer to nil. Just personal taste I guess

eventsources/sources/bitbucket/types.go
daniel-codefresh: Hmm nope you're right, `map[string]interface{}` makes more sense, just forgot to modify this field as its not really used anywhere ATM.
Pushing a fix now.

eventsources/sources/bitbucket/types.go
daniel-codefresh: Yes 100%, will remove them now

eventsources/sources/bitbucket/basic_auth_strategy.go
VaibhavPage: The go imports will group core libs and external libs into 2 sections, right now the imports have unnecessary extra lines between them. Not important but just housekeeping stuff.

VaibhavPage (approved):
LGTM!

whynowy (approved):
LGTM, thanks!

Issue #1479, opened 2021-12-20
Allow AMQP arguments in AMQPQueueDeclare instead of always passing nil (Enables Quorum Queues and other functionality)


jlee-idbydna:
**Is your feature request related to a problem? Please describe.**
We would like to use argo-events to listen to our ingress queue which is a RabbitMQ quorum queue.  However, doing so requires the x-queue-type argument to be set upon queue declaration.  https://www.rabbitmq.com/quorum-queues.html#declaring

**Describe the solution you'd like**
A simple solution may be to allow a key/value map in AMQPQueueDeclareConfig for the AMQP arguments, which can be passed through to amqlib.QueueDeclare instead of the hard coded nil value: https://github.com/argoproj/argo-events/blob/f85b8d01b631b9ae92e5e16435e9c4a50354455a/eventsources/sources/amqp/start.go#L257

**Describe alternatives you've considered**
There are no reasonable alternatives for us without introducing another queue between the quorum queue and argo-events, which defeats the resiliency goal of the quorum queue

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

pdaBlueVoyant:
Funny this is here, I was about to write the exact same request.  To meet our goals I may have to create a local version unless this can get added in the next 3 weeks.  If we cannot find a work around and I add this I'll create a PR.

Capable-Disk4147:
I forked the repo to start this work for @pdaBlueVoyant (he's my architect).

I know Go well, but I haven't used much protobuf, Argo (either workflow or events), or RabbitMQ so while the code change is probably small, I'm not sure on some of the things I'm seeing. I didn't write tests yet because it's only a couple lines of code, and my entire premise might be off. Therefore, I'm opening a PR to make some questions.

whynowy:
Any comments on the spec like following:

```yaml
     exchangeName: test
     exchangeType: fanout
     routingKey: hello
     queueDeclare:
       name: xxx
       durable: xxxxx
       arguments: |-
         a: b
         c:
           d: e
           f: g
```

Where `arguments` is a string in yaml format. @jlee-idbydna, @pdaBlueVoyant

Capable-Disk4147:
hm, I don't entirely know what that Yaml syntax means until I test things today. Are you saying you suspect that it is all one escaped string?

whynowy:
[...]

PR #1501, opened 2022-01-07
feat(eventsource): add optional arguments for amqp eventsource

Capable-Disk4147:
**This is a question PR requiring a little guidance to fulfill (enhancement) issue #1479** before I add a test to this small code change and make sure it's good. I'm not very familiar with protobuf, argo, or the amqp, but I think it should be pretty straight-forward. 

**Changes**

1. First, I added `Arguments []byte `json:"arguments,omitempty" protobuf:"bytes,6,rep,name=arguments"`` to `AMQPQueueDeclareConfig` in types.go in order to allow optional arguments to be passed to a queue declare. In our case, we want this enhancement in order to allow argo events to support quorum style queues.
2. It looks like we need a type of Table in eventsources/sources/amqp/start.go so I tried telling protobuf to use bytes and I unmarshall those to the Table type in getDelivery.

**Questions**

1. Before I work on setting up infrastructure to test this live, does this change make sense in general?
2. Is this the most appropriate protobuf type?
3. At some point the codegen made a lot of changes that appear to be to very unrelated things. Is this something that has an obvious cause for anyone else?


pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -417,6 +417,9 @@ type AMQPQueueDeclareConfig struct {
 	// NowWait when true, the queue assumes to be declared on the server
 	// +optional
 	NoWait bool `json:"noWait,omitempty" protobuf:"varint,5,opt,name=noWait"`
+	// Arguments of a queue (also known as "x-arguments") used for optional features and plugins
+	// +optional
+	Arguments []byte `json:"arguments,omitempty" protobuf:"bytes,6,rep,name=arguments"`
```
whynowy, line 422: Any reason of using `[]byte`? Are you intended to get it as a base64 encoded string?

whynowy:
@Capable-Disk4147 - thanks for submitting the PR, could you commit with `-s` so that `DCO` does not complain?

PaluMacil:
@whynowy sure, I will fix that up when I get answers to the questions since I'm sure I'm not going to be using these specific commits once I know what to do

pkg/apis/eventsource/v1alpha1/types.go
Capable-Disk4147: I am not familiar with protobuf. I know Go well enough to be able to run with things after some guidance, this is a broad question PR which I should be able to fix up with a little guidance.

I spent a fair bit of time trying to figure out what would be most appropriate for the type here. I tried to use Table directly (which I see is a `map[string]interface{}`) which I need to then pass into the `QueueDeclare` method of `Channel` in `amqp/start.go`. This added an import from github.com/streadway/amqp that looked out of place in this types file. Also, the codegen step didn't seem to recognize it.

The `Table` needed in the optional arguments of a queue_declare can be (looking at the pika library in Python) a string key to anything if the typings in Python are correct. That makes sense because these arguments are used for plugins and optional features. It appears I can use a byte slice, which would allow this arbitrary data. Then I can defer the serializing into a Table in the queue declare of the channel.

I'm of course eager to follow any advice on other data types or places to put things.

Capable-Disk4147:
@whynowy I signed off even though I'm uncertain on the data type. I can always sign again

pkg/apis/eventsource/v1alpha1/types.go
Capable-Disk4147: In regards to the message being base64, no, not that I'm aware. I think it is constrained to nearly arbitrary json, except with the top level being a object, or as it's defined in github.com/streadway/amqp, `Table`, which is a `map[string]interface{}`.

[Talking to some people on Slack, starting here](https://gophers.slack.com/archives/C0FSEE1UJ/p1641839086038800) (if you have access to the Gopher Slack), it sounds like getting arbitrary data in protobuf via `[]byte` is probably about the best I can do since the AMQP protocol makes this field maximally dynamic:

> Optional queue arguments, also known as "x-arguments" because of their field name in the AMQP 0-9-1 protocol, is a map (dictionary) of arbitrary key/value pairs that can be provided by clients when a queue is declared.

Then I can serialize it into `Table`, which the AMQP library wants, in the start.go file as I'm doing. There isn't much more I can do with it. A string would cost a slight bit more performance for no real gain.

whynowy: The `arguments` in amqp API is almost an arbitrary map structure, so I'd prefer to put the `arguments` field in the spec as a string in json or yaml format (yaml might be easier):

 ```yaml
      exchangeName: test
      exchangeType: fanout
      routingKey: hello
      queueDeclare:
        name: xxx
        durable: xxxxx
        arguments: |-
          a: b
          c:
            d: e
            f: g
```

Comment (Capable-Disk4147) [reply pkg/apis/eventsource/v1alpha1/types.go:422]: @whynowy I need a couple more keywords to know what you're asking for, I think...
> The arguments in amqp API is almost an arbitrary map structure
Yep, sounds right to me
> I'd prefer to put the arguments field in the spec as a string in json or yaml format
- spec? Like the Argo types file? Meaning you just want the Arguments type on 422 to be a string here in types.go?
- What are you asking me to do with "in json or yaml format"?
  - Are you asking me to change how the swagger docs are generated? How do I do that?
  - Or are you asking me to change the format sent to Argo from the libraries that I'm using (in my case, Pika in Python)? Or perhaps it's already Yaml... I'm getting the RabbitMQ, argo, and argo-events all set up today to test this out and see if I understand it well enough to make the changes needed.

Comment (whynowy) [reply pkg/apis/eventsource/v1alpha1/types.go:422]: Please use `string` instead of `[]byte` as below.

```
Arguments string `json:"arguments,omitempty" protobuf:"bytes,6,opt,name=arguments"`
```

File: eventsources/sources/amqp/start.go [hunk 2]
```diff
@@ -19,6 +19,7 @@ package amqp
 import (
 	"context"
 	"encoding/json"
+	"github.com/ghodss/yaml"
```
whynowy, line 22: Please use `sigs.k8s.io/yaml` which is a better implementation forked from `ghodss` one. The `ghodss` one has some issues.

whynowy (commented):
Did you test it?

whynowy:
Could you also run `go mod tidy`?

whynowy:
Please fix the CI mentioned errors:

```
verifying github.com/fasthttp/websocket@v1.4.2: checksum mismatch
	downloaded: h1:AU/zSiIIAuJjBMf5o+vO0syGOnEfvZRu40xIhW/3RuM=
	go.sum:     h1:hTnl0DRQ9Tj4uVc4QcZCGp+mgF5mVKnijR1eNyn2waY=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.
```

Capable-Disk4147:
> Please fix the CI mentioned errors:
> 
> ```
> verifying github.com/fasthttp/websocket@v1.4.2: checksum mismatch
> 	downloaded: h1:AU/zSiIIAuJjBMf5o+vO0syGOnEfvZRu40xIhW/3RuM=
> 	go.sum:     h1:hTnl0DRQ9Tj4uVc4QcZCGp+mgF5mVKnijR1eNyn2waY=
> 
> SECURITY ERROR
> This download does NOT match an earlier download recorded in go.sum.
> The bits may have been replaced on the origin server, or an attacker may
> have intercepted the download attempt.
> ```

Any idea on how this happened? I don't use this dependency in my changes. I just pushed a unit test, but this appears to be a dependency from unrelated e2e tests. Go mod tidy has no effect. I manually changed the hash back to the original from before my work. Hopefully that's okay. I must have a very weird problem in my mod cache.

Even though this might resolve the immediate issue and we have the test passing, we didn't have time today to test this in the real environment with argo events and rabbit mq, so we will probably need to do that next week now.

pkg/apis/eventsource/v1alpha1/types.go
Capable-Disk4147: done

Capable-Disk4147:
I'm not super familiar with signing. I commit with -s and then DCO fails, so I follow what it recommends. Now it looks like I forgot a commit that wasn't the most recent, and i haven't been able to figure out signing further back in history

whynowy (approved):
LGTM, thanks!

PR #1511, opened 2022-01-11
fix: Sensor triggers an Argo Workflow with wrong name

daniel-codefresh:

### DIFF
--- a/sensors/triggers/params.go
+++ b/sensors/triggers/params.go
@@ -166,6 +166,7 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
 	var eventPayload []byte
 	var key string
 	var tmplt string
+	var resultValue string
 
 	event, eventExists := events[src.DependencyName]
 	switch {
@@ -188,7 +189,8 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
 		}
 	case src.Value != nil:
 		// Use the default value set by the user in case the event is missing
-		return src.Value, nil
+		resultValue = *src.Value
+		return &resultValue, nil
 	default:
 		// The parameter doesn't have a default value and is referencing a dependency that is
 		// missing in the received events. This is not an error and may happen with || conditions.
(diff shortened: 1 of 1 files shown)


daniel-codefresh:
Fixes: #1511

daniel-codefresh:
Context:
In some cases `src.Value` pointer was returned from `ResolveParamValue` function and the `ApplyParams` function was dereferencing this pointer and changing its value: 
https://github.com/argoproj/argo-events/blob/f91e05905ae55094912ff1153b80fd3736db7727/sensors/triggers/params.go#L112
Due to this reason, the original default value that was set by the user was overridden which of course effected the next actions that were triggered by the same sensor.

whynowy:
> Context: In some cases `src.Value` pointer was returned by `ResolveParamValue` and then the `ApplyParams` function was dereferencing this pointer and changing its value:
> 
> https://github.com/argoproj/argo-events/blob/f91e05905ae55094912ff1153b80fd3736db7727/sensors/triggers/params.go#L112
> 
> 
> Due to this reason, the original default value that was set by the user was overridden which of course effected the next actions that were triggered by the same sensor.

Thanks @daniel-codefresh ! Do you mean this issue only happens when operation is set to append?

daniel-codefresh:
> > Context: In some cases `src.Value` pointer was returned by `ResolveParamValue` and then the `ApplyParams` function was dereferencing this pointer and changing its value:
> > https://github.com/argoproj/argo-events/blob/f91e05905ae55094912ff1153b80fd3736db7727/sensors/triggers/params.go#L112
> > 
> > Due to this reason, the original default value that was set by the user was overridden which of course effected the next actions that were triggered by the same sensor.
> 
> Thanks @daniel-codefresh ! Do you mean this issue only happens when operation is set to append?

Prepend as well:
https://github.com/argoproj/argo-events/blob/f91e05905ae55094912ff1153b80fd3736db7727/sensors/triggers/params.go#L114

whynowy:
> > > Context: In some cases `src.Value` pointer was returned by `ResolveParamValue` and then the `ApplyParams` function was dereferencing this pointer and changing its value:
> > > https://github.com/argoproj/argo-events/blob/f91e05905ae55094912ff1153b80fd3736db7727/sensors/triggers/params.go#L112
> > > 
> > > Due to this reason, the original default value that was set by the user was overridden which of course effected the next actions that were triggered by the same sensor.
> > 
> > 
> > Thanks @daniel-codefresh ! Do you mean this issue only happens when operation is set to append?
> 
> Prepend as well:
> 
> https://github.com/argoproj/argo-events/blob/f91e05905ae55094912ff1153b80fd3736db7727/sensors/triggers/params.go#L114

👍

whynowy (approved):

PR #1512, opened 2022-01-11
Add SNSEvent to RecreateStrategyEventSources

svmaris:
The SNS EventSource creates (and removes) a subscription to the SNS Topic in AWS. With the default RollingUpdate strategy, the subscription is removed by the terminating Pod after the new EventSource starts, preventing the EventSource from working.

Checklist:

### DIFF
--- a/pkg/apis/common/common.go
+++ b/pkg/apis/common/common.go
@@ -73,6 +73,7 @@ var (
 		HDFSEvent,
 		FileEvent,
 		GenericEvent,
+		SNSEvent
 	}
 )
 
--- a/pkg/apis/common/common.go
+++ b/pkg/apis/common/common.go
@@ -73,7 +73,7 @@ var (
 		HDFSEvent,
 		FileEvent,
 		GenericEvent,
-		SNSEvent
+		SNSEvent,
 	}
 )
 


whynowy:
@svmaris - Please note right now `RecreateStrategyEventSources` is not honored for deployment update strategy, but just for HA strategy (active - active or active - standby), making this change means there will be only one pod serving traffic actively when you use HA. Do you think this will solve the issue?

We still can make deployment the deploy honor `RecreateStrategyEventSources` though.

svmaris:
@whynowy Thanks for looking at this. Unfortunately, that probably won't fix my problem. It seems that it's not possible to manually override the `StrategyType` of the generated `Deployment` as they all default to `RollingUpdate`. And even if we could, it won't completely fix this issue, because there's a chance that events are missed during the deployment. Setting the `StrategyType` to `Recreate` only makes sure that the subscription is set-up correctly after the deployment is finished.

It seems that the only way to fix this currently is to create a daemon goroutine in the `EventSource` that periodically checks if the subscription is still there, just like the Github/Gitlab/Bitbucket. And even then you can't guarantee that the subscription is configured at all times, so there's still the chance that you'll miss events.

Is there really no way to move the Subscription/Webhook setup/teardown code directly to the lifecycle of the `EventSource` resource instead of the individual `Pods` of the `Deployment`? As long as the `EventSource` exists, the webhooks/subscriptions should too, regardless of the status of the `Pods`.

whynowy:
> @whynowy Thanks for looking at this. Unfortunately, that probably won't fix my problem. It seems that it's not possible to manually override the `StrategyType` of the generated `Deployment` as they all default to `RollingUpdate`. And even if we could, it won't completely fix this issue, because there's a chance that events are missed during the deployment. Setting the `StrategyType` to `Recreate` only makes sure that the subscription is set-up correctly after the deployment is finished.
> 
> It seems that the only way to fix this currently is to create a daemon goroutine in the `EventSource` that periodically checks if the subscription is still there, just like the Github/Gitlab/Bitbucket `EventSources`. And even then you can't guarantee that the subscription is configured at all times, so there's still the chance that you'll miss events.
> 
> Is there really no way to move the Subscription/Webhook setup/teardown code to the lifecycle of the `EventSource` resource instead of the individual `Pods` of the `Deployment`? As long as the `EventSource` exists, the webhooks/subscriptions should too, regardless of the status of the individual `Pods`.

This is actually the same issue as `github` and `gitlab` have - they need to set up the webhook when the pod starts and clean it up afterwards. Right now they use the workaround that having a daemon goroutine to do that as you mentioned above.

The proper approach might be using a k8s job to create those resources at the time eventsource is created, and clean up when it is deleted.

Issue #1277, opened 2021-07-14
Allow SSL connections to Kafka, even when not using TLS authentication


signaleleven:
**Is your feature request related to a problem? Please describe.**

The kafka library used by argo-events supports SASL_SCRAM authentication over an SSL connection. See for instance https://github.com/Shopify/sarama/blob/master/examples/sasl_scram_client/main.go

However, when configuring an EventSource for kafka like in the snippet below, the Eventsource pod attempts to connect to Kafka in Plaintext.

```yaml
 kafka:
    cdm:
      url: blahblah-msk-endpoint.eu-central-1.amazonaws.com:9096
      topic: test
      jsonBody: true
      # partition id
      partition: "1"
      # optional backoff time for connection retries.
      # if not provided, default connection backoff time will be used.
      connectionBackoff:
        # duration in nanoseconds, or strings like "3s", "2m". following value is 10 seconds
        duration: 1000000000 # "10s" is reported invalid here, but that's another issue
        # how many backoffs
        steps: 5
        factor: 2
        jitter: 0.2
     # tls: TLS IS NOT the answer. TLS enables an SSL connection, but it's for client authentication using SSL. Something that MSK supports only with a very expensive private CA. AWS, amirite? 🤷 
      sasl:
        mechanism: SCRAM-SHA-512
        passwordSecret:
          key: password
          name: kafkauser
        userSecret:
          key: user
          name: kafkauser
```
```
namespace=orekit-fetcher, eventSourceName=kafka, eventSourceType=kafka, eventName=cdm, level=info, time=2021-07-14T16
:14:23Z, msg=start kafka event source...
namespace=orekit-fetcher, eventSourceName=kafka, eventSourceType=kafka, eventName=cdm, level=info, time=2021-07-14T16
:14:23Z, msg=connecting to Kafka cluster...
namespace=orekit-fetcher, eventSourceName=kafka, eventSourceType=kafka, eventName=cdm, level=error, time=2021-07-14T1
6:14:23Z, msg=failed to start listening eventsource
namespace=orekit-fetcher, eventSourceName=kafka, level=error, time=2021-07-14T16:14:23Z, msg=Erroring out, no active
event server running
namespace=orekit-fetcher, eventSourceName=kafka, level=info, time=2021-07-14T16:14:23Z, msg=exiting eventbus connecti
on daemon...
namespace=orekit-fetcher, eventSourceName=kafka, level=error, time=2021-07-14T16:14:23Z, msg=failed to start
[...]

PR #1528, opened 2022-01-13
Feat/enable kafka eventsource ssl with tls

usamaB:
Adds option to create TLS without certs so Kafka(MSK) with SASL_SSL can work. 
#1277


USERS.md
```diff
@@ -32,3 +32,4 @@ Organizations below are **officially** using Argo Events. Please send a PR with
 1. [Viaduct](https://www.viaduct.ai/)
 1. [WooliesX](https://wooliesx.com.au/)
 1. [Woolworths Group](https://www.woolworthsgroup.com.au/)
+1. [Mobimeo GmbH](https://mobimeo.com/en/home/)
```
whynowy, line 35: Alphabetical order please.

pkg/apis/common/common.go
```diff
@@ -132,6 +132,8 @@ type TLSConfig struct {
 	ClientCertSecret *corev1.SecretKeySelector `json:"clientCertSecret,omitempty" protobuf:"bytes,2,opt,name=clientCertSecret"`
 	// ClientKeySecret refers to the secret that contains the client key
 	ClientKeySecret *corev1.SecretKeySelector `json:"clientKeySecret,omitempty" protobuf:"bytes,3,opt,name=clientKeySecret"`
+	// If set to true, skips creation of TLSConfig with certs and creates an empty TLSConfig
+    InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty" protobuf:"varint,4,opt,name=insecureSkipVerify"`
```
whynowy, line 136: Could you let them have same indents?

whynowy (commented):
I assume the PR is still in progress.

pkg/apis/common/common.go
usamaB: indented

common/util.go
```diff
@@ -238,6 +246,11 @@ func GetTLSConfig(config *apicommon.TLSConfig) (*tls.Config, error) {
 		}
 	}
 
+	if !config.InsecureSkipVerify && len(caCertPath)+len(clientCertPath)+len(clientKeyPath) == 0 {
```
whynowy, line 249: This is redundant. When the code is executed here, `InsecureSkipVerify` must be false, and line 250 already has the check there.

pkg/apis/common/validate.go
```diff
@@ -9,7 +9,11 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
 	if tlsConfig == nil {
 		return nil
 	}
-	var caCertSet, clientCertSet, clientKeySet bool
+	var caCertSet, clientCertSet, clientKeySet, insecureSkipVerify bool
+
+	if !tlsConfig.InsecureSkipVerify {
```
whynowy, line 14: redundant.

pkg/apis/common/validate.go
```diff
@@ -23,6 +27,10 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
 		clientKeySet = true
 	}
 
+	if !caCertSet && !clientCertSet && !clientKeySet && !insecureSkipVerify {
```
whynowy, line 30:
```
if !caCertSet && !clientCertSet && !clientKeySet && !tlsConfig.InsecureSkipVerify {
```

common/util.go
usamaB: yes, just wanted to create a separate test for it but since the default value is false, doesn't make much sense. Need to remove the test as well. 
Done!

pkg/apis/common/validate.go
usamaB: done!

pkg/apis/common/validate.go
usamaB: done!

pkg/apis/common/validate.go
```diff
@@ -23,6 +23,10 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
 		clientKeySet = true
 	}
 
+	if !caCertSet && !clientCertSet && !clientKeySet && !tlsConfig.InsecureSkipVerify {
```
whynowy, line 26: Should validate the case that `InsecureSkipVerify` is true. Also missing a test case for that.

whynowy:
Also, please run `make codegen`.

pkg/apis/common/common.go
```diff
@@ -132,6 +132,8 @@ type TLSConfig struct {
 	ClientCertSecret *corev1.SecretKeySelector `json:"clientCertSecret,omitempty" protobuf:"bytes,2,opt,name=clientCertSecret"`
 	// ClientKeySecret refers to the secret that contains the client key
 	ClientKeySecret *corev1.SecretKeySelector `json:"clientKeySecret,omitempty" protobuf:"bytes,3,opt,name=clientKeySecret"`
+	// If true, skips creation of TLSConfig with certs and creates an empty TLSConfig. Optional default is false
+	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty" protobuf:"varint,4,opt,name=insecureSkipVerify"`
```
whynowy, line 136: The comments format matter.
```
// xxxxx description goes here. xxxxxx
// +optional
InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty" protobuf:"varint,4,opt,name=insecureSkipVerify"`
```

usamaB: changed and added the code-gen files. The tests are passing locally.

pkg/apis/common/validate.go
usamaB: added test for the case `InsecureSkipVerify` is true

pkg/apis/common/validate.go
```diff
@@ -23,6 +23,14 @@ func ValidateTLSConfig(tlsConfig *TLSConfig) error {
 		clientKeySet = true
 	}
 
+	if tlsConfig.InsecureSkipVerify {
```
whynowy, line 26: Is this right? When `InsecureSkipVerify` is true, we should let it pass and the rest does not matter right?

usamaB: yes! that's the point of it. It's in TLS documentation as well. Currently argo-events doesn't allow to connect with MSK with SASL,SSL. I can only find this way to make it provide that. 
https://github.com/argoproj/argo-events/issues/1277

usamaB: > Also, please run `make codegen`.

Do you know why is the codegen failing? it doens't fail on my side

usamaB:
> 

Do you know why is the codegen failing? it doens't fail on my side

usamaB:
> Also, please run `make codegen`.

Do you know why is the codegen failing? it doens't fail on my side

pkg/apis/common/validate.go
whynowy: I mean, your current logic is when `InsecureSkipVerify` is true, it returns an error, this doesn't seem to be right.

go.mod
```diff
@@ -88,183 +88,25 @@ require (
 	github.com/Azure/azure-amqp-common-go/v3 v3.1.0 // indirect
 	github.com/Azure/azure-sdk-for-go v52.6.0+incompatible // indirect
 	github.com/Azure/go-amqp v0.13.6 // indirect
-	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
```
whynowy, line 91: Something is wrong with `go.mod`? Cleaning up so much shouldn't be right.

pkg/apis/common/validate.go
usamaB: No, that's only in the Validation function because the validate function has an `error` return type. If I use the `Fmt.Sprint` I'd have to change to `string` and all the tests would have to change as well. Then we'd have this question again for the other validation conditions(fields) whether it's better for them to return error or string.

go.mod
usamaB: true! could it be dependent on the go version? mine is go1.16.3 darwin/amd64
I can revert this file and push

pkg/apis/common/validate.go
whynowy: Validation function returns an error means it will not go through. If the spec is valid, simply return nil.

usamaB: Yeah! Others tests have it as well. I was thinking of just having Assert.Nil() would be less significant. Done!

go.mod
usamaB: I have reverted this file but if I run make codegen it makes these changes in go.mod

MaorDavidzon:
+1 We are facing the exact same issue with confluent cloud :)

usamaB:
@whynowy You can review it now. I have built, deployed, and tested this in my cluster. It works as expected. It's running in our cluster. 
Argo-events is now able to connect and consume from with KAFKA with SASL_SCRAM mechanisms and with MSK. 
Previously it was only able to connect with Kafka with PLAINTEXT or with or no auth. 

I haven't tested producing to Kafka yet but it'll do that soon as well.

whynowy:
> 

will do.

pkg/apis/common/validate.go
usamaB: That is for when it's true. The false is insignificant so I'll remove the test case as well it's redundant.

usamaB: done

whynowy (approved):
LGTM, thanks!

PR #1529, opened 2022-01-13
fix(docs): Fix doc about event transform release date

### LINKED ISSUE #1527: Event Transforms not working in v1.5.6
[Documentation](https://argoproj.github.io/argo-events/sensors/transform/) says that event transforms are available after version 1.5.5 but they are not.

Seems like transforms were implemented in #1492 . It got merged to master but its not there in v1.5.6

hamza3202:
Fixes #1527

### DIFF
--- a/docs/sensors/transform.md
+++ b/docs/sensors/transform.md
@@ -1,6 +1,6 @@
 # Event Transformation
 
-> Available after v1.5.5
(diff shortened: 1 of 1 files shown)


whynowy (approved):
Thanks!

PR #1535, opened 2022-01-16
fix: Removed "limitation" from outdated doc

daniel-codefresh:


daniel-codefresh:
Context:
Theres no real limitation because we start a new http server for a specific port only if it doesn’t exist already (if exists, the new endpoint will be assigned to the same server instance):
Screen Shot 2022-01-14 at 11 45 22

Using the same port multiple times in a single eventsource works fine:
Screen Shot 2022-01-14 at 11 14 30 (1)

whynowy (approved):

Issue #1452, opened 2021-12-08
Multiple repository support for Bitbucketserver Event Sources


serainville:
**Is your feature request related to a problem? Please describe.**
Environments with a large number of Bitibucket repositories require an equal number of bitbucketserver event sources. There should not be a  1:1 relationship with Bitbucket repositories and an event source.

**Describe the solution you'd like**
The Bitbucketserver event source should be modelled after the Gitlab and Github event sources, which permit events from any number of projects or repos. Also, management of a Bitbucket repository's webhook should be optional.  Not all environments permit Bitbucket webhooks to be managed by Argo Events.

Simplify the event source to accept events from a Bitbucket server for any project or repository, with optional authorization support using webhook secrets

**Describe alternatives you've considered**
There are no alternatives for Bitbucket Webhooks where HMAC is required for workflow authorization.

**Additional context**

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

daniel-codefresh:
@serainville @whynowy Related PR: #1540

PR #1540, opened 2022-01-19
Support multiple repos bitbucketserver

daniel-codefresh:


daniel-codefresh:
Related issue: #1452

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -865,16 +865,21 @@ type BitbucketServerEventSource struct {
 	// Webhook holds configuration to run a http server
 	Webhook *WebhookContext `json:"webhook,omitempty" protobuf:"bytes,1,opt,name=webhook"`
 	// ProjectKey is the key of project for which integration needs to setup
-	ProjectKey string `json:"projectKey" protobuf:"bytes,2,opt,name=projectKey"`
+	// +optional
+	ProjectKey string `json:"projectKey,omitempty" protobuf:"bytes,2,opt,name=projectKey"`
```
whynowy, line 869: Are we deprecating this? If yes, it would be better to mention it in the comments, also give a version when it will be unsupported. For this one, we can stop supporting it in v1.8.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -190,26 +192,34 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+	s1 := rand.NewSource(time.Now().UnixNano())
+	r1 := rand.New(s1)
+	time.Sleep(time.Duration(r1.Intn(2000)) * time.Millisecond)
+	createWebhooks()
+
 	go func() {
 		// Another kind of race conditions might happen when pods do rolling upgrade - new pod starts
 		// and old pod terminates, if DeleteHookOnFinish is true, the hook will be deleted from Bitbucket.
-		// This is a workround to mitigate the race conditions.
+		// This is a workaround to mitigate the race conditions.
```
whynowy, line 222: Similar workaround has been existing in several event sources, do you think it would be better if we let the controller to create a K8s job to create the hook when there's a new or updated event source object, and another job to delete the hook when the event source object is deleted?

daniel-codefresh: Thats an interesting idea actually, but what if the user decides to update the webhook configuration in the event source object? Wouldn't the controller need to create a Job for deleting the old hook as well? Because if so, I think we would still face race conditions with this solution as well

pkg/apis/eventsource/v1alpha1/types.go
daniel-codefresh: Yes I was wondering if we should deprecate it - v1.8 sounds good to me! Will add the comments right now

eventsources/sources/bitbucketserver/start.go
whynowy: I think the biggest issue for updating is, we don't know what is the old hook, because the controller only knows the updated object.....

whynowy (approved):
LGTM, thanks!!!!

eventsources/sources/bitbucketserver/start.go
daniel-codefresh: Hmm yes thats a good point. I mean we can always store some state in the event source objects (etcd) and then use it on update events by reading it from the old object state, but this would definitely introduce a whole lot of new complexity to the system..

whynowy: No.. When an updating happens, the event source object is updated in etcd before the controller sees it.

daniel-codefresh: But wouldn't the controller be able to see the old object as well as the new object on update events?

whynowy: It is swallowed by controller-runtime. https://github.com/kubernetes-sigs/controller-runtime/issues/399

PR #1544, opened 2022-01-24
chore: develop with k3d cluster

whynowy:

Checklist:

### DIFF
--- a/controllers/eventsource/resource.go
+++ b/controllers/eventsource/resource.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"os"
 
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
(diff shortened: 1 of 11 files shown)


controllers/eventsource/resource.go
```diff
@@ -281,9 +282,13 @@ func buildDeployment(args *AdaptorArgs, eventBus *eventbusv1alpha1.EventBus) (*a
 }
 
 func buildDeploymentSpec(args *AdaptorArgs) (*appv1.DeploymentSpec, error) {
+	imgPullPolicy := corev1.PullAlways
```
VaibhavPage, line 285: can we have a method under common package called getImagePolicy() String {}?

manifests/base/sensor-controller/sensor-controller-deployment.yaml
```diff
@@ -19,6 +19,7 @@ spec:
       containers:
         - name: sensor-controller
           image: quay.io/argoproj/argo-events:latest
+          imagePullPolicy: Always
```
alexec, line 22: This changes the default behaviour? Which was - `Always` if `latest` and `IfNotPresent` otherwise.

alexec (approved):

sarabala1979 (approved):

manifests/base/sensor-controller/sensor-controller-deployment.yaml
whynowy: When `imagePullPolicy` is missing, it defaults to `Always`.

PR #1581, opened 2022-02-01
feat: add a default location to look for .kube/config in home directory

juliev0:
adding a default location to look for .kube/config in home directory if the environment variable isn't defined


whynowy (approved):
LGTM, thanks!

PR #1585, opened 2022-02-02
fix: conditions reset does not work if the service is down at the triggering time

juliev0:
This fix uses the cron syntax specified in the Condition Reset to determine the last time the trigger should have fired prior to application start up. If any messages arrive with a timestamp prior to that time, they are acknowledged.


common/cronutil_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	cronlib "github.com/robfig/cron/v3"
+)
+
+func TestPrevCronTime(t *testing.T) {
```
whynowy, line 11: It's nice to see all these testing cases!

eventbus/driver/nats.go
```diff
@@ -255,23 +258,15 @@ func (n *natsStreaming) processEventSourceMsg(m *stan.Msg, msgHolder *eventSourc
...
-		// ACK all the old messages after conditions meet
-		if m.Timestamp <= msgHolder.latestGoodMsgTimestamp {
-			if depName != "" {
-				msgHolder.reset(depName)
-			}
-			msgHolder.ackAndCache(m, event.ID())
-			return
-		}
+	// Acknowledge any old messages that occurred either before the last reset
+	if m.Timestamp/1e9 <= msgHolder.getLastResetTime() {
```
whynowy, line 262: Much cleaner, love it!

eventbus/driver/nats.go
```diff
@@ -392,15 +383,14 @@ func newEventSourceMessageHolder(dependencyExpr string, dependencies []Dependenc
 	}
 
 	return &eventSourceMessageHolder{
-		lastResetTime:          int64(0),
-		latestGoodMsgTimestamp: int64(0),
```
whynowy, line 396: 👍

whynowy (commented):
Great work!

common/cronutil_test.go
juliev0: Thank you. I'll admit that I used the test cases from the cron gomodule as examples. :)

eventbus/driver/nats.go
juliev0: thanks!

juliev0:
Here's a description of how I tested this: 

1. created a Sensor with a dependency of A && B && C and a Condition Reset and tested the following cases:
   1a. per the bug, I scaled the deployment down to 0, waited for the CR time to pass, then scaled it back up to 1 to verify that the reset occurred for all dependencies
   1b. standard Condition Reset logic: cron fired and reset occurred for all dependencies (to verify I didn't break this)
   1c. standard logic: all 3 dependencies were sent (to verify I didn't break this)
2. also, to verify I didn't break anything I tried creating a Sensor with just one dependency and triggered it

-

eventbus/driver/nats.go
```diff
@@ -439,8 +446,9 @@ func (mh *eventSourceMessageHolder) ackAndCache(m *stan.Msg, id string) {
 func (mh *eventSourceMessageHolder) reset(depName string) {
 	mh.parameters[depName] = false
 	delete(mh.msgs, depName)
+
 	if mh.isCleanedUp() {
-		mh.setLastResetTime(0)
+		mh.resetTimeout = 0
```
whynowy, line 451: This should be protected by a lock, e.g. we could utilize `setLastResetTime()`.

eventbus/driver/nats.go
```diff
@@ -451,7 +459,7 @@ func (mh *eventSourceMessageHolder) resetAll() {
 	for k := range mh.parameters {
 		mh.parameters[k] = false
 	}
-	mh.setLastResetTime(0)
+	mh.resetTimeout = 0
```
whynowy, line 462: Same here.

eventbus/driver/nats.go
juliev0: okay, fixed this

eventbus/driver/nats.go
```diff
@@ -411,9 +413,30 @@ func (mh *eventSourceMessageHolder) getLastResetTime() int64 {
...
 
 func (mh *eventSourceMessageHolder) setLastResetTime(t int64) {
-	mh.lock.Lock()
-	defer mh.lock.Unlock()
-	mh.lastResetTime = t
+	{
+		mh.lock.Lock() // since this can be called asyncronously as part of a ConditionReset, we neeed to lock this code
+		defer mh.lock.Unlock()
+		mh.lastResetTime = t
+	}
```
whynowy, line 420: Can we share the same lock and simply call `setLastResetTime(0)` instead of `setRestTimeout(0)`? It should be safe to set `lastResetTime` to 0, right?

```
mh.lock.Lock()
defer mh.lock.Unlock()
mh.lastResetTime = t
if t == 0 {
  mh.resetTime = 0
} else {
  mh.resetTime = t + 60
}
```

juliev0: No, one of the changes I made as part of this whole thing is that I never set lastResetTime to 0. lastResetTime now always represents an actual time. This ended up being necessary in a certain scenario.

whynowy: Certain scenario such as?

juliev0: Say you have condition A && B && C,
Say A and B both occur 
Say the pod goes down while the condition reset time occurs
So, we want to be able to ack both A and B when they arrive but we don't: this is because if A arrives, we call reset(), which ends up doing this:
       `if mh.isCleanedUp() {
		mh.setLastResetTime(0)
	}`
That is because the pod went down and B is no longer stored in the eventSourceMessageHolder's memory. 
Then when B arrives we fail to ack it, because we would skip the logic which checks lastResetTime.

whynowy: Good Point!

whynowy (approved):
LGTM, great work!

PR #1601, opened 2022-02-04
docs: create filtering.md

dpadhiar:
Created documentation to detail how to use event source filtering and to update with examples and use cases.


whynowy (approved):
LGTM, thanks @dpadhiar !

PR #1637, opened 2022-02-15
fix: adds SigningCertURL validation for SNS messages

aweis89:
Given the SigningCertURL is part of the HTTP request body, I don't see where we're validating that the cert URL is indeed valid and from AWS before retrieving the cert [here](https://github.com/argoproj/argo-events/blob/master/eventsources/sources/awssns/start.go#L294). [Here's](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/SimpleNotificationService/Custom/Util/Message.cs#L244) an example of this being done in the .net SDK but I don't see that happening in the Golang SDK anywhere.

This PR adds that validation check explicitly.

Checklist:

### DIFF
--- a/eventsources/sources/awssns/start.go
+++ b/eventsources/sources/awssns/start.go
@@ -25,6 +25,7 @@ import (
 	"encoding/pem"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


whynowy (approved):
LGTM, thanks for adding this!

Issue #1649, opened 2022-02-21
Can't install the Validating Admission Webhook on Openshift


danielm-codefresh:
**Describe the bug**
At the moment it is not possible to install the argo-events validating admission webhook on Openshift.

**To Reproduce**
1. apply the validating admission webhook manifest:
`oc apply -f https://raw.githubusercontent.com/argoproj/argo-events/v1.6.0/manifests/install-validating-webhook.yaml`
2. Look at the events-webhook pod

**Expected behavior**
The events-webhook pod is running and the webhook is registered successfully.

**Screenshots**
The events-webhook pod is stuck in `Error` status with a bunch of errors regarding lack of permissions to set finalizers on some resources:
<img width="1704" alt="Screen Shot 2022-02-21 at 17 37 04" src="https://user-images.githubusercontent.com/94960579/154986202-4b57563c-6bfc-41c1-9dd7-f6141499c1f0.png">

**Environment (please complete the following information):**
 - Kubernetes: 1.21.6
 - Argo Events: v1.6.0

**Additional context**
* In order for the the validating admission controller to work, we need to add a few additional permissions in the argo-events-webhook cluster role. Specifically update permission for finalizers on resources that will be attached as ownerReferences (since K8s might add finalizers on owner references).
* This permission is necessary if the [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) plugin is enabled in your cluster. (which is enabled by default for Openshift)
---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #1650, opened 2022-02-21
fix: Openshift support for the Validating Admission Webhook

danielm-codefresh:
Signed-off-by: danielm-codefresh <daniel.maizel@codefresh.io>

Checklist:

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -26,9 +26,26 @@
     
                 kubectl create clusterrolebinding YOURNAME-cluster-admin-binding --clusterrole=cluster-admin --user=YOUREMAIL@gmail.com
            
-         * On Openshift, make sure to grant `anyuid` scc to the service account.
+         * On Openshift: 
+             - Make sure to grant `anyuid` scc to the service account.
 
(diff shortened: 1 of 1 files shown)


danielm-codefresh:
Fixes #1649

manifests/extensions/validating-webhook/rbac/argo-events-webhook-cluster-role.yaml
```diff
@@ -27,9 +27,11 @@ rules:
       - apps
     resources:
       - deployments
+      - deployments/finalizers
```
whynowy, line 30: 1. I think this would be better to go with a notice in a doc for OpenShift installation.
2. From the document, it only requires update privilege on `xxx/finalizers`, not the  update privilege on the whole `xxx`.

danielm-codefresh: 1. Alright, I'll add it to the docs.
2. You're right

whynowy (approved):
Thanks!

PR #1657, opened 2022-02-22
chore: Allow dependabot to open two prs

blkperl:
Checklist:

@alexec is okay with allowing two prs now.


whynowy (approved):
thanks!

PR #1664, opened 2022-02-23
fix: read kafka error channel. Fixes #1656

### LINKED ISSUE #1656: Kafka trigger: read kafka producer's error channel
When the kafka cluster brokers are restarted, I notice that the kafka-sensor fails to produce the messages. And no kafka-related error messages printed in log. I have to manually restart the sensor pod, then it works again.

argo-event version: 1.3.1

According to the comment of sarama's AsyncProducer:
```
// Errors is the error output channel back to the user. You MUST read from this
// channel or the Producer will deadlock when the channel is full. Alternatively,
// you can set Producer.Return.Errors in your config to false, which prevents
// errors to be returned.
```
However, there is no such option to specify the Producer.Return.Errors to be False in kafka-trigger yaml template.

After checking the source code of [kafka trigger](https://github.com/argoproj/argo-events/blob/master/sensors/triggers/kafka/kafka.go) , I found that the producer's error channel is not read.

Therefore I believe the most straightforward way to fix is by reading error channel, for example

``` Go
// inside NewKafkaTrigger function, add below code

go func() {
	
	for err := range producer.Errors() {
		log.Println(err)
	}
}()
```

**Additional context**
The root cause is potentially related to 
https://github.com/Shopify/sarama/pull/2133 
https://github.com/Shopify/sarama/issues/1430 
Ideally kafka async producer should refresh metadata and re-connect to brokers automatically, but seems that is not completely fixed by Shopify. 

---

csimplestring:
Checklist:

This tiny PR is to fix this [issue](https://github.com/argoproj/argo-events/issues/1656). The kafka async producer's error channel must be read otherwise the producer will be blocked. This case happened, e.g., when the brokers restarted.

### DIFF
--- a/sensors/triggers/kafka/kafka.go
+++ b/sensors/triggers/kafka/kafka.go
@@ -123,14 +123,12 @@ func NewKafkaTrigger(sensor *v1alpha1.Sensor, trigger *v1alpha1.Trigger, kafkaPr
 		kafkaProducers[trigger.Template.Name] = producer
 	}
 
-	kafkaTrigger := &KafkaTrigger{
+	return &KafkaTrigger{
 		Sensor:   sensor,
 		Trigger:  trigger,
 		Producer: producer,
 		Logger:   triggerLogger,
-	}
-
-	return kafkaTrigger, nil
+	}, nil
 }
 
 // GetTriggerType returns the type of the trigger
--- a/sensors/triggers/kafka/kafka.go
+++ b/sensors/triggers/kafka/kafka.go
@@ -47,6 +47,7 @@ type KafkaTrigger struct {
 // NewKafkaTrigger returns a new kafka trigger context.
 func NewKafkaTrigger(sensor *v1alpha1.Sensor, trigger *v1alpha1.Trigger, kafkaProducers map[string]sarama.AsyncProducer, logger *zap.SugaredLogger) (*KafkaTrigger, error) {
 	kafkatrigger := trigger.Template.Kafka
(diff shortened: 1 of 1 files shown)


sensors/triggers/kafka/kafka.go
```diff
@@ -115,12 +115,21 @@ func NewKafkaTrigger(sensor *v1alpha1.Sensor, trigger *v1alpha1.Trigger, kafkaPr
...
+	kafkaTrigger := &KafkaTrigger{
 		Sensor:   sensor,
 		Trigger:  trigger,
 		Producer: producer,
 		Logger:   logger.With(logging.LabelTriggerType, apicommon.KafkaTrigger),
-	}, nil
+	}
+
+	//must read from the Errors() channel or the async producer will deadlock.
+	go func() {
```
whynowy, line 126: Will this goroutine exit after `Execute()`?

csimplestring: No, this goroutine will run until the sensor pod exit.

whynowy: This block of code should go into the section above line 116, as for now, every time when the trigger executes, we run `NewKafkaTrigger()` - good thing is, we reuse the same producer.

Going forward, we should change it to only run `NewKafkaTrigger()` one time.

csimplestring: ok, got it. Thanks for explaining. Done.

whynowy (approved):
LGTM, thanks for submitting the PR!

PR #1670, opened 2022-02-24
fix: remove string replacement on expressions for event filtering

### LINKED ISSUE #1667: Eventsource filtering modifies the filter expression at runtime -> doesn't work as expected
**Describe the bug**
The new EventSource filtering feature modifies the filters at runtime (replacing `-` with `_` in values), making it impossible (at least to my knowledge) to apply filters for values containing `-`.

**To Reproduce**
1. Create an Event Source with a filter like this
```yaml
kind: EventSource
metadata:
  name: kafka-eventsource-test
spec:
  kafka:  
    my-test-event:
[...]
       expression: "(body.someUUIDfield == '7e577e57-7e57-7e57-7e57-7e577e577e57')"
```
2. Write a message on the kafka topic (but I suppose this is the same for other eventsources) like this one
```json
{ "_id": "whatever",
  "someUUIDfield": "7e577e57-7e57-7e57-7e57-7e577e577e57"
 }
```
3. Check the Eventsource logs: the event is received but NOT published on the eventbus (filtered)
4. Write a message on the kafka topic like this one (note the `_` instead of `-`)
```json
{ "_id": "whatever",
  "someUUIDfield": "7e577e57_7e57_7e57_7e57_7e577e577e57"
 }
```
5. Check the Eventsource logs: the event is received AND published on the eventbus (not filtered) despite clearly NOT matching the filter.

**Expected behavior**
The filter expression is applied as written, instead of being modified.

 - Kubernetes: v1.22
 - Argo Events: v1.6.0

**Additional context**
The reason for this (and for my very specific procedure to reproduce above) is most likely this line:
https://github.com/argoproj/argo-events/blob/df18152bb54cd55573990f69c1dfad60a645dedf/eventsources/eventing.go#L568

I see that the filters are processed to turn `-`s into `_`s. I assume that was done for a reason*, but I can imagine filtering for string values containing - can be important. It is important to me at least 😄  And I cannot figure out how to do it. 

*: @dpadhiar can you comment on this? Maybe you only meant to do this for the keys, not for the values, in order to fit the pairs into a map? I'm neither a go nor an expr expert, so I'm just thinking out loud... btw, **thank you** for this feature, it got merged pretty much on the day I realized I needed it (Sensor filters is ok functionally, but my metrics sampled on the eventsources are wrong, as they count events that then get filtered by the sensor. Your feature solves my problem, minus this bug)

---

dpadhiar:
Signed-off-by: Dillen Padhiar <dpadhiar99@gmail.com>

Fixes #1667 

Previously, expressions were filtered to replace `-` with `_` similar to how params for their keys of their pairs. However this causes unintended behavior and makes filters fail on messages which it shouldn't. 

Example yaml from issue:

```
kind: EventSource
metadata:
  name: kafka-eventsource-test
spec:
  kafka:  
    my-test-event:
[...]
       expression: "(body.someUUIDfield == '7e577e57-7e57-7e57-7e57-7e577e577e57')"
```

Sample JSON message:

```
{"someUUIDfield": "7e577e57-7e57-7e57-7e57-7e577e577e57",
  "id": 1}
```

We expect this message to be published since it will evaluate true with our expression.

Logs for this message being sent:
```
2022-02-24T18:14:12.584Z        INFO    argo-events.eventsource kafka/start.go:217      dispatching event on the data channel...        {"eventSourceName": "kafka", "eventSourceType": "kafka", "eventName": "example", "partition-id": "0"}
2022-02-24T18:14:12.607Z        INFO    argo-events.eventsource eventsources/eventing.go:512    succeeded to publish an event   {"eventSourceName": "kafka", "eventName": "example", "eventSourceType": "kafka", "eventID": "kafka:example:kafka-broker:9092:topic-2:0:12"}
```

We also test a JSON message with all `_` instead to show that we have removed string replacement from the expression as previously this would evaluate to `true`.

```
{"someUUIDfield": "7e577e57_7e57_7e57_7e57_7e577e577e57",
  "id": 1}
```

Logs for this message being sent:
```
2022-02-24T18:06:16.493Z        INFO    argo-events.eventsource kafka/start.go:217      dispatching event on the data channel...        {"eventSourceName": "kafka", "eventSourceType": "kafka", "eventName": "example", "partition-id": "0"}
2022-02-24T18:06:16.550Z        DEBUG   argo-events.eventsource eventsources/eventing.go:477    Do not publish event, filter condition not met  {"eventSourceName": "kafka"}
```

### DIFF
--- a/eventsources/eventing.go
+++ b/eventsources/eventing.go
@@ -565,5 +565,5 @@ func filterEvent(data []byte, filter *v1alpha1.EventSourceFilter) (bool, error) 
 		params[strings.ReplaceAll(key, "-", "_")] = value
 	}
 	env := expr.GetFuncMap(params)
-	return expr.EvalBool(strings.ReplaceAll(filter.Expression, "-", "_"), env)
+	return expr.EvalBool(filter.Expression, env)
 }
--- a/common/expr/eval_test.go
+++ b/common/expr/eval_test.go
@@ -41,6 +41,7 @@ func TestEvalBool(t *testing.T) {
 		"email":      "johndoe@intuit.com",
 		"gender":     "Male",
 		"dept":       "devp",
+		"uuid":       "test-case-hyphen",
 	}
 
 	pass, err := EvalBool("(id == 1) && (last_name == 'Doe')", env)
@@ -54,4 +55,9 @@ func TestEvalBool(t *testing.T) {
 	pass, err = EvalBool("invalidexpression", env)
 	assert.Error(t, err)
 	assert.False(t, pass)
+
+	// expr with '-' evaluate the same as others
+	pass, err = EvalBool("(uuid == 'test-case-hyphen')", env)
+	assert.NoError(t, err)
+	assert.True(t, pass)
 }


eventsources/eventing.go
```diff
@@ -565,5 +565,5 @@ func filterEvent(data []byte, filter *v1alpha1.EventSourceFilter) (bool, error)
 		params[strings.ReplaceAll(key, "-", "_")] = value
 	}
 	env := expr.GetFuncMap(params)
-	return expr.EvalBool(strings.ReplaceAll(filter.Expression, "-", "_"), env)
+	return expr.EvalBool(filter.Expression, env)
```
alexec, line 568: and a test?

dpadhiar: Added a test case for EvalBool for expressions with `-` and added output for example for original issue to PR.

alexec (approved):

Issue #1694, opened 2022-03-01
Release 1.6.0 breaks AMQP event source and loses messages


pdaBlueVoyant:
**Describe the bug**
We recently upgraded from 1.5.6 to 1.6.0 and found that several messages are getting lost in the event bus.  We have confirmed via logs that all the queued messages appear in the event source, but 4 out of 5 never make it to the event sensor.

We use a RabbitMQ (amqp) event source.  The loss happens when we queue 5 messages at about the same time.  I noticed some changes to EventBus and EventSource (forget which) involved timestamps and the NATS event bus and wonder if there is a problem when events come in that have more-or-less identical time stamps.  Our setup uses several event source bindings to different queues.  Not sure that is important as problem happens using just 1 of the queues.  Things work fine if we send messages 1 at time (human speed).

**To Reproduce**
Steps to reproduce the behavior (rough outline):
1. Set up an event source using AMQP and and event sensor to process things.
2. Queue a bunch of messages into RabbitMQ quickly (we could repro this w/ 5 messages at once)
3. Notice only 1 message is processed; the others are lost (bad, very bad)

**Expected behavior**
Every message that is sent to a Rabbit MQ queue that is bound to an event source should be processed.

**Environment (please complete the following information):**
 - Kubernetes: 1.21.5
 - Argo: 3.3.0 rc6
 - Argo Events: 1.6.0

**Additional context**
We backed out EventSensor and EventBus components to use 1.5.6 and everything worked fine.  We kept EventSource at 1.6.0 as we require the newly released feature for quorum queues (made by one of my teammates and I).

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

tooptoop4:
@pdaBlueVoyant do you have logs like https://github.com/argoproj/argo-events/issues/1547 ?

pdaBlueVoyant:
No, the EventSource logs were clean.  There were no errors, if I recall.  There were 5 log messages in Event source, one for each queued message.  Looking for the corresponding logs in Event sensor, there were no errors, just 1 message instead of 5.  I will try to dig up the logs and post them...will most likely take until tomorrow.

juliev0:
[...]

PR #1695, opened 2022-03-02
fix: changing lastResetTime to a time.Time from int64 seconds so time comparison can occur at a finer granularity

juliev0:
Checklist:

### DIFF
--- a/eventbus/driver/nats.go
+++ b/eventbus/driver/nats.go
@@ -472,6 +472,7 @@ func (mh *eventSourceMessageHolder) resetAll() {
 	for k := range mh.msgs {
 		delete(mh.msgs, k)
 	}
+
 	for k := range mh.parameters {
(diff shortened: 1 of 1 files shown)


juliev0:
This is in reference to this bug: https://github.com/argoproj/argo-events/issues/1694

juliev0:
I have tested this with the following conditions:
1. similar to bug: just one dependency in the condition and multiple webhooks curled immediately one after the other: verifying each is triggered in sensor
2.  conditional expression involving && and both webhooks are curled - verify sensor is triggered
3. standard condition reset case with conditional expression involving &&: verifying condition reset occurs
4. condition reset case where pod is down at the cron time

whynowy (approved):
LGTM, thanks for the quick fix!

tooptoop4:
when is next release with this fix?

PR #1701, opened 2022-03-02
fix: Quick fix for Issue #1694

juliev0:
I believe this would take care of the case of messages actually received in order by the sensor, which would I think be the majority of the time: basically right now we're comparing a publish time (m.Timestamp) to a receive time (lastResetTime). Instead, lastResetTime should be set to publish time as well, so we can compare apples to apples.

Checklist:


tooptoop4:
keen for new release with this 🥇

Issue #1564, opened 2022-01-26
Combine controllers to 1


whynowy:
Right now we have `eventsource-controller`, `eventbus-controller` and `sensor-controller` installed in `argo-events` namespace, which can be combined to 1.

PR #1705, opened 2022-03-03
feat: jetstream eventbus controller implementation

whynowy:

This PR is part of the effort of JetStream EventBus implementation.

A JetStream EventBus spec looks like below, where `version` is a required field in the spec.

```yaml
kind: EventBus
metadata:
  name: default
spec:
  jetstream:
    version: 2.7.3
```

TLS enabling is not done in this PR, it should be implemented before we do an official release.

Current `Nats Streaming` image configuration is also moved into a configmap, together with JetStream image configs. This configmap is now only used by `eventbus-controller`. As one of the follow-up tasks, all of the 3 controllers will be merged to one #1564.

Checklist:


common/hash.go
```diff
@@ -0,0 +1,21 @@
+package common
```
whynowy, line 1: We should refactor all these utilities under common.

pkg/apis/eventbus/v1alpha1/jetstream_eventbus.go
```diff
@@ -0,0 +1,96 @@
...
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty" protobuf:"bytes,13,opt,name=affinity"`
+	// ServiceAccountName to apply to the StatefulSet
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty" protobuf:"bytes,14,opt,name=serviceAccountName"`
+	// JetStream configuration, if not specified, global settings in controller-config will be used.
+	// See https://docs.nats.io/running-a-nats-service/configuration#jetstream.
+	// Only configure "max_memory_store" or "max_file_store", do not set "store_dir" as it has been hardcoded.
+	// +optional
+	Settings *string `json:"settings,omitempty" protobuf:"bytes,15,opt,name=settings"`
```
juliev0, line 73: This is nice and generic. I'll update my design document to reflect this.

pkg/apis/eventbus/v1alpha1/jetstream_eventbus.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (j JetStreamBus) GetReplicas() int {
+	if j.Replicas == nil {
+		return 3
+	}
+	if *j.Replicas < 3 {
```
juliev0, line 80: oh, interesting. We don't let them have less than 3. Ideally maybe we'd warn them but not critical.

juliev0 (approved):
Nice. Left a few comments but didn't really see any issues.

pkg/apis/eventbus/v1alpha1/jetstream_eventbus.go
whynowy: Cluster setting has a different configuration from single node setting, it's not worth supporting that for us.

For a cluster setting, it needs at least 2 nodes to make it work. JetStream uses a RAFT algorithm for clustering, which does not make sense to use an even number. So a reasonable cluster node number is 3 or 5.

```
In order to ensure data consistency across complete restarts, a quorum of servers is required. A quorum is ½ cluster size + 1. This is the minimum number of nodes to ensure at least one node has the most recent data and state after a catastrophic failure. So for a cluster size of 3, you’ll need at least two JetStream enabled NATS servers available to store new messages. For a cluster size of 5, you’ll need at least 3 NATS servers, and so forth.
```

https://docs.nats.io/running-a-nats-service/configuration/clustering/jetstream_clustering#the-quorum

juliev0: Okay. Just wondering if we might want to log something so it's not a silent change. But I'm good with whatever - probably rare that somebody would set replicas to 1 or 2.

manifests/base/eventbus-controller/controller-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+          # Only configure "max_memory_store" or "max_file_store", do not set "store_dir" as it has been hardcoded.
+          # e.g. 1G. -1 means no limit, up to 75% of available memory
+          max_memory_store: -1
+          # e.g. 20G. -1 means no limit, Up to 1TB if available
+          max_file_store: 1TB
+        versions:
+        - version: 2.7.3
+          natsImage: nats:2.7.3
+          metricsExporterImage: natsio/prometheus-nats-exporter:0.9.1
+          startCommand: /nats-server
```
daniel-codefresh, line 26: shouldn't it be `nats-server` (without the `/`) ?

daniel-codefresh (approved):
Overall looks great! left few small comments/questions

manifests/base/eventbus-controller/controller-config.yaml
whynowy: This image is built from `scratch`, the binary is put under `/`,  so we have to use `/nats-server`.

controllers/config.go
```diff
@@ -0,0 +1,87 @@
...
+func (g *GlobalConfig) GetNatsStreamingVersion(version string) (*NatsStreamingVersion, error) {
+	if g.EventBus.NATS == nil || len(g.EventBus.NATS.Versions) == 0 {
+		return nil, fmt.Errorf("nats streaming configuration not found")
+	}
+	for _, r := range g.EventBus.NATS.Versions {
+		if r.Version == version {
+			return &r, nil
+		}
+	}
+	return nil, fmt.Errorf("no nats streaming configuration found for %q", version)
```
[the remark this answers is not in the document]
whynowy: Good point!

controllers/eventbus/installer/jetstream.go
```diff
@@ -0,0 +1,615 @@
...
+		old.Spec = spec
+		if err := r.client.Update(ctx, old); err != nil {
+			return fmt.Errorf("failed to update jetstream statefulset, err: %w", err)
+		}
+		r.logger.Info("updated jetstream statefulset successfully")
+	}
+	return nil
+}
+
+func (r *jetStreamInstaller) buildStatefulSetSpec(jeVersion *controllers.JetStreamVersion) appv1.StatefulSetSpec {
```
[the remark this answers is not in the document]
whynowy: Having a default version will cause future upgrade issue - if `version` is not specified, we use a default version, that will prevent us from updating the default version later, otherwise if the default one is changed, those installation with default versions will be automatically updated. This is a problem for current nats streaming implementation.

pkg/apis/eventbus/v1alpha1/jetstream_eventbus.go
whynowy: Adding a log sounds good.

whynowy: It seems to be awkward to add a log in the model, skip it for now.

controllers/eventbus/installer/jetstream.go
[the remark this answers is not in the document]
daniel-codefresh: Got it, thanks for the clarification

controllers/eventbus/validate.go
```diff
@@ -1,29 +1,36 @@
...
+				return fmt.Errorf("\"spec.nats.exotic.clusterID\" is missing")
+			}
+			if e.URL == "" {
+				return fmt.Errorf("\"spec.nats.exotic.url\" is missing")
+			}
+		}
+	}
+	if x := eb.Spec.JetStream; x != nil {
+		if x.Version == "" {
+			return fmt.Errorf("invalid spec: a version for jetstream needs to be specified")
```
sarabala1979, line 33: Can we add `replicas` validation here?

whynowy: sounds good, done.

sarabala1979 (approved):
LGTM

PR #1719, opened 2022-03-09
feat: enabling Jetstream publishing/subscribing

juliev0:
In order to incorporate Jetstream I needed to separate out our existing driver interface into multiple interfaces, since the driver interface didn't support everything I needed.

The problems I was trying to address as part of this:
- I needed some of my Jetstream logic to occur outside of one individual Trigger (i.e. the cleanup effort in the K/V store that happens on the Sensor level (removing old Triggers, etc))
- I wanted Jetstream-specific code and NATS Streaming-specific code to be able to determine things like: client ID, group name, and subject name since the two are different in that regard.

I've now created new interfaces in addition to the original ones (driver.Driver and driver.Connection):

- sourceEventBus.Driver: for setting up connections for the EventSource
- sensorEventBus.Driver: for setting up connections for the Sensor
- SourceConnection: for publishing Events for an individual Source
- TriggerConnection: for subscribing to Events for an individual Trigger

If it helps I can provide a class/interface diagram.

Checklist:


whynowy:
I like the idea to have different drivers and connections for event source and sensor. One suggestion is grouping them into a better structure under centralized `eventbus` dir. For example:

```
- interface.go
- driver.go
|- jetstream
   |- jet stream impl files
|- nats
   |- nats impl files
```

interface.go
```golang
type Connection interface {
	Close() error

	IsClosed() bool

	ClientID() string
}

type EventSourceConnection interface {
	Connection

	Publish(subject string, data []byte) error
}

type SensorConnection interface {
	Connection

	subscribe()
}

type EventSourceDriver interface {
	Connect() EventSourceConnection
}

type SensorDriver interface {
	Connect() SensorConnection
}
```

driver.go (or you name it)
```golang
func GetEventSourceDriver() EventSourceDriver {
	return nil
}

func GetSensorDriver() SensorDriver {
	return nil
}
```

juliev0:
> I like the idea of having different drivers and connections for event source and sensor. One suggestion is grouping them into a better structure under centralized `eventbus` dir. For example:
> 
> ```
> - interface.go
> - driver.go
> |- jetstream
>    |- jet stream impl files
> |- nats
>    |- nats impl files
> ```
> 
> interface.go
> 
> ```go
> type Connection interface {
> 	Close() error
> 
> 	IsClosed() bool
> 
> 	ClientID() string
> }
> 
> type EventSourceConnection interface {
> 	Connection
> 
> 	Publish(subject string, data []byte) error
> }
> 
> type SensorConnection interface {
> 	Connection
> 
> 	Subscribe()
> }
> 
> type EventSourceDriver interface {
> 	Connect() EventSourceConnection
> }
> 
> type SensorDriver interface {
> 	Connect() SensorConnection
> }
> ```
> 
> driver.go (or you name it)
> 
> ```go
> func GetEventSourceDriver() EventSourceDriver {
> 	return nil
> }
> 
> func GetSensorDriver() SensorDriver {
> 	return nil
> }
> ```

Sure, thanks. That might make it easier to find things.

juliev0:
> This is not the right way to do it, we should not have 2 sets of same files for testing different eventbus implementations.
> 
> The right way to do it is, to support reading an evn variable which is used to indicate what type of event bus will be used, and run same testing for each of them. For example, to run e2e test with jetstream driver, run following command:
> 
> ```
> EventBusDriver=jetstream make test-functional
> ```
> 
> When we run e2e test with github action for different eventbus, we will not use only one CI to do it, instead, we will run 1 for each.

Good. I was thinking about that kind of thing but wasn't sure. I didn't like having copies of those files, but it was the line in the .yaml file for the "eventBusName" that was different between the two. Is your idea to somehow incorporate that environment variable into the .yaml file like "eventBusName: ${EventBusDriver}" (not sure if kustomize allows that), or will the code manipulate it some other way?

eventsources/eventing.go
```diff
@@ -390,13 +390,17 @@ func (e *EventSourceAdaptor) run(ctx context.Context, servers map[apicommon.Even
...
 	clientID := generateClientID(e.hostname)
-	driver, err := eventbus.GetDriver(ctx, *e.eventBusConfig, e.eventBusSubject, clientID)
+	driver, err := eventbus.GetEventSourceDriver(ctx, *e.eventBusConfig, e.eventSource.Name, e.eventBusSubject)
 	if err != nil {
 		logger.Errorw("failed to get eventbus driver", zap.Error(err))
 		return err
 	}
 	if err = common.Connect(&common.DefaultBackoff, func() error {
-		e.eventBusConn, err = driver.Connect()
+		err = driver.Initialize()
```
juliev0, line 399: oops, this shouldn't be here...need to move outside of the trigger logic

juliev0:
Closing this PR so I can open a new one and see if that takes care of the issue of tests not being run.

Issue #1747, opened 2022-03-17
Event loss by Event Source/Sensors after enabling HA (High Availability) 


PremanandPatil:
**Describe the bug**
HA was enabled on Event Source, So there were two event source were running, one is master(Leader) and another is follower.
Meanwhile we have upgraded Event Bus Configuration, it took little longer time than usual, in mean time both ES becomes follower. 

We were expecting that after one ES will be become Leader after Event Bus updation, but both ES remains follower and non of them become Leaser, which cause that our events were not consumed and we miss our dependency due to this.  

1. Create ES with HA 
2. Create Sensor which will be triggered by ES on some condition
3. Stop Event Bus or upgrade it's configuration
4. both ES will become follower ("I am not Leader")
5. bring back Event Bus
6. both ES will stay as follower ("I am not Leader")

**Expected behavior**
Expected behavior one of ES should become Leader and start consuming message from event bus. 

**Screenshots**
N/A

**Environment (please complete the following information):**
 - Kubernetes: [e.g. v1.18.6]
 - Argo: [e.g. v2.9.4]
 - Argo Events: 1.5.6

**Additional context**
This issue has been identified when Event Bus upgradation was happening for more than 20 min 

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
Max reconnection to Nats server defaults to 60, needs to change it to retry forever.

PR #1748, opened 2022-03-18
fix(leaderelection): never give up connection retry to nats server

whynowy:
Fixes: #1747

1. The max auto reconnection to Nats server for Leader Election defaults to 60, changed to retry forever;
2. If EventBus auth file is changed (e.g. the auth secret is deleted by mistake, the controller will regenerate a secret), let the eventsource and sensor pods restart to pick up the new auth file.

Checklist:


sarabala1979 (approved):

tooptoop4:
would be great to have a new release with this, helps fix https://github.com/argoproj/argo-events/issues/1547

PR #1774, opened 2022-03-28
build: Make `make lint` faster

alexec:
Signed-off-by: Alex Collins <alex_collins@intuit.com>

Checklist:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,5 @@
 run:
+  timeout: 5m
   skip-dirs:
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -29,4 +28,3 @@ linters:
     - unused
```
alexec, line 28: I felt like there were a lot of linters, but lets see how it works

whynowy (approved):

Issue #1794, opened 2022-03-31
Jetstream - to enhance "exactly once" add a cache of previous messages for the Sensor


juliev0:
Argo Events doesn't promise "exactly once" 100% of the time, but we should aim for almost all of the time. To better enhance that we can add a cache of previous messages in the Sensor processing logic.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #1802, opened 2022-03-31
feat: Jetstream - Get closer to "exactly once" by maintaining a cache of message IDs in the Trigger processing code

juliev0:
Fixes: #1794 

Signed-off-by: Julie Vogelman <julie_vogelman@intuit.com>

Checklist:

### DIFF
--- a/eventbus/jetstream/sensor/trigger_conn.go
+++ b/eventbus/jetstream/sensor/trigger_conn.go
@@ -28,11 +28,11 @@ type JetstreamTriggerConn struct {
 	evaluableExpression  *govaluate.EvaluableExpression
 	deps                 []eventbuscommon.Dependency
 	sourceDepMap         map[string][]string // maps EventSource and EventName to dependency name
-	recentMsgsByID       map[string]*Msg     // prevent re-processing the same message as before (map of msg ID to time)
-	recentMsgsByTime     []*Msg
+	recentMsgsByID       map[string]*msg     // prevent re-processing the same message as before (map of msg ID to time)
+	recentMsgsByTime     []*msg
 }
 
-type Msg struct {
+type msg struct {
 	time  int64
 	msgID string
 }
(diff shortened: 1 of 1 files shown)


eventbus/jetstream/sensor/trigger_conn.go
```diff
@@ -28,6 +28,13 @@ type JetstreamTriggerConn struct {
 	evaluableExpression  *govaluate.EvaluableExpression
 	deps                 []eventbuscommon.Dependency
 	sourceDepMap         map[string][]string // maps EventSource and EventName to dependency name
+	recentMsgsByID       map[string]*Msg     // prevent re-processing the same message as before (map of msg ID to time)
+	recentMsgsByTime     []*Msg
+}
+
+type Msg struct {
```
whynowy, line 35: `msg`?

eventbus/jetstream/sensor/trigger_conn.go
```diff
@@ -227,6 +237,14 @@ func (conn *JetstreamTriggerConn) processMsg(
 		return
 	}
 
+	// De-duplication
+	// In the off chance that we receive the same message twice, don't re-process
+	_, alreadyReceived := conn.recentMsgsByID[event.ID()]
```
whynowy, line 242: If same ID message comes every 2 mins, it will be blocked here, right?

eventbus/jetstream/sensor/trigger_conn.go
juliev0: Oh, I see: lower case it to keep it only accessible within the package? sure, good idea

eventbus/jetstream/sensor/trigger_conn.go
juliev0: I believe we resolved this offline

whynowy (approved):

PR #1845, opened 2022-04-13
Fixes related to making Sensor (and Eventsource) more robust if one or more Jetstream Pods goes down

juliev0:
With this branch of code, I have verified that the Sensor functions well when:
1. All EventBus pods are scaled down to 0 and then back up again
2. Individual EventBus pods are deleted - I have a script which randomly delete 10 Jetstream Pods one after the other.
After these events, I have verified that the Sensor leader is able to a) receive subsequent messages on the original durable consumer, and b) make use of the original K/V store

Checklist:

### DIFF
--- a/eventsources/eventing.go
+++ b/eventsources/eventing.go
@@ -435,7 +435,7 @@ func (e *EventSourceAdaptor) run(ctx context.Context, servers map[apicommon.Even
 				logger.Info("exiting eventbus connection daemon...")
 				return
 			case <-ticker.C:
-				if e.eventBusConn == nil || e.eventBusConn.IsInterfaceValueNil() || e.eventBusConn.IsClosed() {
+				if e.eventBusConn == nil || e.eventBusConn.IsClosed() {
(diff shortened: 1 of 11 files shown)


eventsources/eventing.go
```diff
@@ -435,7 +435,7 @@ func (e *EventSourceAdaptor) run(ctx context.Context, servers map[apicommon.Even
 				logger.Info("exiting eventbus connection daemon...")
 				return
 			case <-ticker.C:
-				if e.eventBusConn == nil || e.eventBusConn.IsClosed() {
+				if e.eventBusConn == nil || e.eventBusConn.IsInterfaceValueNil() || e.eventBusConn.IsClosed() {
```
alexec, line 438: There is another way to do this that is more conventional, have the func perform a nil chek:

```golang
func (e *eventBusConn) IsClosed() {
  return e == nil || ...
}
````

whynowy (approved):
LGTM

Issue #1792, opened 2022-03-31
Expand e2e tests for both Jetstream and STAN


juliev0:
e2e tests for Jetstream (which also apply to STAN) can be expanded to include:
- capability of Sensor to go down and come back up and retain its state and subscription
- capability of multiple Sensors to exist and trigger correctly

Bonus:
- changes to Sensor spec cause state to be maintained for Triggers that have changed but not for those that haven't
- conditional reset test

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #1859, opened 2022-04-19
feat: Expand e2e tests

juliev0:
Fixes: #1792

Checklist:


juliev0:
Not necessarily ready to be merged. Want to see how this runs on the CI server.

Emadmojiri (approved):

test/util/util.go
```diff
@@ -245,35 +245,41 @@ deployWatch:
...
 	if err != nil {
 		return false, fmt.Errorf("error getting sensor pod name: %w", err)
 	}
 
-	return PodsLogContains(ctx, kubeClient, namespace, regex, podList, timeout), nil
+	return PodsLogContains(ctx, kubeClient, namespace, regex, podList, timeout, countOpt), nil
 }
 
-func PodsLogContains(ctx context.Context, kubeClient kubernetes.Interface, namespace, regex string, podList *corev1.PodList, timeout time.Duration) bool {
+func PodsLogContains(ctx context.Context, kubeClient kubernetes.Interface, namespace, regex string, podList *corev1.PodList, timeout time.Duration, countOpt *int) bool {
```
whynowy, line 268: A better approach would be using options like below, and the function signature accepts `...PodLogCheckOption` so that all the invokers will not need to add `nil` as a parameter.

type podLogCheckOptions struct {
  timeout time.Duration
  count int
}

func defaultPodLogCheckOptions() *podLogCheckOptions {
  return &podLogCheckOptions {
    timeout: xxxx,
    count: -1,
  }
}

type PodLogCheckOption func(*podLogCheckOptions)

func WithTimeout(t time.Duration) PodLogCheckOption {
  return func(o *podLogCheckOptions) {
    o.timeout = t
  }
}

func WithCount(c int) PodLogCheckOption {
  return func(o *podLogCheckOptions) {
    o.count = c
  }
}

juliev0: encapsulating timeout and count into a struct sounds good. So, you're recommending a variadic parameter so that the invoker doesn't need to pass anything in? One thing I'm wondering about is that a variadic parameter also implies that you can have more than 1, which would be odd in this case. 

Do we sometimes use variadic parameters but only look at the first one specified?

whynowy: The way it handles it is like below:

unc PodsLogContains(ctx context.Context, kubeClient kubernetes.Interface, namespace, regex string, podList *corev1.PodList, opts ...PodLogCheckOption) bool {
  o := defaultPodLogCheckOptions()
  for _, opt := range opts {
    if opt != nil {
      opt(o)
    }

    // then you can use `o.timeout` and `o.count`.
  }




}

juliev0: I see

whynowy (approved):
Good work!

PR #1860, opened 2022-04-19
feat: User should not be able to change Jetstream StreamConfig

### LINKED ISSUE #1793: Jetstream - User should not be able to change Jetstream Stream settings if Sensor/EventSource are attached
Since there is just one Stream we can't allow the changing of the Stream Settings once EventSources and Sensors are already using it with the original stream settings.

---

juliev0:
Fixes #1793 

User should not be able to change Jetstream StreamConfig, since once it's been used the stream settings have been passed to Jetstream

Checklist:

### DIFF
--- a/controllers/eventbus/validate.go
+++ b/controllers/eventbus/validate.go
@@ -9,7 +9,7 @@ import (
 // ValidateEventBus accepts an EventBus and performs validation against it
 func ValidateEventBus(eb *v1alpha1.EventBus) error {
 	if eb.Spec.NATS == nil && eb.Spec.JetStream == nil {
-		return fmt.Errorf("invalid spec: either \"nats\" or \"jststream\" needs to be specified")
+		return fmt.Errorf("invalid spec: either \"nats\" or \"jetstream\" needs to be specified")
 	}
 	if x := eb.Spec.NATS; x != nil {
 		if x.Native != nil && x.Exotic != nil {
--- a/webhook/validator/eventbus.go
+++ b/webhook/validator/eventbus.go
@@ -33,6 +33,7 @@ func (eb *eventbus) ValidateCreate(ctx context.Context) *admissionv1.AdmissionRe
 	if err := eventbuscontroller.ValidateEventBus(eb.neweb); err != nil {
 		return DeniedResponse(err.Error())
 	}
+
 	return AllowedResponse()
 }
 
--- a/webhook/validator/eventbus.go
+++ b/webhook/validator/eventbus.go
@@ -2,7 +2,6 @@ package validator
 
 import (
 	"context"
-	"fmt"
 
 	admissionv1 "k8s.io/api/admission/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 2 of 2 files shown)


controllers/eventbus/validate.go
```diff
@@ -9,7 +9,7 @@ import (
 // ValidateEventBus accepts an EventBus and performs validation against it
 func ValidateEventBus(eb *v1alpha1.EventBus) error {
 	if eb.Spec.NATS == nil && eb.Spec.JetStream == nil {
-		return fmt.Errorf("invalid spec: either \"nats\" or \"jststream\" needs to be specified")
+		return fmt.Errorf("invalid spec: either \"nats\" or \"jetstream\" needs to be specified")
```
whynowy, line 12: 👍

webhook/validator/eventbus.go
```diff
@@ -64,7 +66,21 @@ func (eb *eventbus) ValidateUpdate(ctx context.Context) *admissionv1.AdmissionRe
 				return DeniedResponse("\"spec.nats.exotic.auth\" is immutable, can not be updated")
 			}
 		}
+	case eb.neweb.Spec.JetStream != nil:
+		if eb.oldeb.Spec.JetStream == nil {
+			return DeniedResponse("Can not change event bus implmementation")
```
whynowy, line 71: typo

whynowy (commented):
Anything else needs to be protected? e.g. pvc changed?

webhook/validator/eventbus.go
juliev0: not seeing the typo - where? This is the case that somebody used to have a "jetstream" section and now doesn't, right? (thus changing the eventbus) This is the same as the "nats" check above on line 49. Admittedly, the error message could be more specific.

whynowy: `implmementation`

juliev0: thanks. will fix in both spots

juliev0:
> Anything else needs to be protected? e.g. pvc changed?

@whynowy I noticed that if somebody does try to change the persistence setting Kubernetes doesn't allow that change to be made for StatefulSet. The Controller log says: “updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden". 

If you do "kubectl describe eventbus <name>", you see:
```
Conditions:
    Last Transition Time:  2022-04-20T19:04:06Z
    Status:                True
    Type:                  Configured
    Last Transition Time:  2022-04-20T19:04:06Z
    Message:               failed to update jetstream statefulset, err: StatefulSet.apps "eventbus-jetstream-ex-js" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden
    Reason:                JetStreamStatefulSetFailed
    Status:                False
    Type:                  Deployed
```

Do you think we should add our own check so it gets caught by the admission controller and the error is more apparent? I guess this is an issue common to all settings that get incorporated into the StatefulSet other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds'.

whynowy:
> 

oh, you are right.

whynowy (approved):
LGTM

PR #1880, opened 2022-04-21
docs: fix some typos in the markdown documentation

jcantosz:
Signed-off-by: Jesse Antoszyk <22500761+jcantosz@users.noreply.github.com>

Checklist:

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,4 +1,4 @@
-# Security policy for Argo Events 
+# Security policy for Argo Events
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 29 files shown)


whynowy (approved):
Thanks for the fix!

I assume you use some kind of IDE plugin, right?

jcantosz:
More hacky than that unfortunately 😆 

My IDE has a [spellchecker](https://github.com/streetsidesoftware/vscode-spell-checker) but as far as I know it wont check all files or fix mistakes, just identify them. I wound up doing this (with [GNU aspell](http://aspell.net/)):
```bash
find ./ -iname "*.md" -exec sh -c "echo {} && cat {} | aspell list | sort | uniq && echo" \; > spellcheck.out
```
Checking the list for things my IDE's spellcheck said were mistakes (squiggly lines) and updating those files.

Will have to see if there is a better way in the future. May look at enhancing that spellcheck plugin

whynowy:
@jcantosz - that's a lot of work! Thanks!

PR #1916, opened 2022-05-01
docs: Clarify in the docs that EventSources in Active-Passive and Sensors should use odd number of replicas

juliev0:
Checklist:

### DIFF
--- a/docs/sensors/ha.md
+++ b/docs/sensors/ha.md
@@ -2,9 +2,10 @@
 
 Sensor controller creates a k8s deployment (replica number defaults to 1) for
 each Sensor object. HA with `Active-Passive` strategy can be achieved by setting
-`spec.replicas` to a number greater than 1, which means only one Pod serves
-traffic and the rest ones stand by. One of standby Pods will be automatically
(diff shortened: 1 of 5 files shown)


docs/sensors/ha.md
```diff
@@ -2,9 +2,10 @@
 
 Sensor controller creates a k8s deployment (replica number defaults to 1) for
 each Sensor object. HA with `Active-Passive` strategy can be achieved by setting
-`spec.replicas` to a number greater than 1, which means only one Pod serves
-traffic and the rest ones stand by. One of standby Pods will be automatically
-elected to be active if the old one is gone.
+`spec.replicas` to an odd number greater than 2, which means only one Pod serves
```
whynowy, line 5: I understand RAFT recommends odd number nodes, but do we see any case that setting `replicas` to 1 or 2 fails the application, and setting to 3 makes it better?

juliev0: not that I've seen

whynowy: Then I think we should not state this.

Issue #1929, opened 2022-05-05
GitHub webhook race condition workaround abuses API when it's not needed


ibabiankou:
**Describe the bug**
Even when `DeleteHookOnFinish` is false (default value AFAICT) [GitHub EventSource starts a daemon](https://github.com/argoproj/argo-events/blob/master/eventsources/sources/github/start.go#L363-L379) that checks every minute the webhook exists on the GitHub side. It puts a hard limit on how many repositories can be handled before exceeding the rate limits.

**To Reproduce**
1. Configure a GitHub EventSource with a webhook and the `DeleteHookOnFinish` disabled.
2. The pod makes API requests to check whether the webhook exists every minute.

**Expected behavior**
The daemon to mitigate race condition should only be started when it's needed (the `DeleteHookOnFinish` flag is enabled).

**Environment (please complete the following information):**
 - Kubernetes: v1.18.20
 - Argo: v3.3.2
 - Argo Events: v1.6.0

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #1930, opened 2022-05-05
fix: Limit github hooks manager daemon lifetime to 10 min

ibabiankou:
Fixes https://github.com/argoproj/argo-events/issues/1929

Checklist:

### DIFF
--- a/eventsources/sources/github/start.go
+++ b/eventsources/sources/github/start.go
@@ -360,23 +360,27 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 		time.Sleep(time.Duration(r1.Intn(2000)) * time.Millisecond)
 		f()
 
-		go func() {
+		if githubEventSource.DeleteHookOnFinish {
 			// Another kind of race conditions might happen when pods do rolling upgrade - new pod starts
 			// and old pod terminates, if DeleteHookOnFinish is true, the hook will be deleted from github.
 			// This is a workaround to mitigate the race conditions.
-			logger.Info("starting github hooks manager daemon")
-			ticker := time.NewTicker(60 * time.Second)
-			defer ticker.Stop()
-			for {
-				select {
-				case <-ctx.Done():
-					logger.Info("exiting github hooks manager daemon")
-					return
-				case <-ticker.C:
-					f()
+			go func() {
+				logger.Info("starting github hooks manager daemon")
+				ticker := time.NewTicker(60 * time.Second)
+				defer ticker.Stop()
+				for {
+					select {
+					case <-ctx.Done():
+						logger.Info("exiting github hooks manager daemon")
+						return
+					case <-ticker.C:
+						f()
+					}
 				}
-			}
-		}()
+			}()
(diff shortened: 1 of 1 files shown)


whynowy (commented):
There's one case:

1. `DeleteHookOnFinish` was `true`;
2. `DeleteHookOnFinish` was updated to `false`
3. The new old will come up before the old pod gets terminated, which means the hook will be deleted but nobody fixes it.

I'm okay with ignoring this corner case if the rate limit is a big concern. What do you think?

ibabiankou:
Ah, very nice! I didn't think about that case 🤔 If that happens, restarting the pod would fix the issue. But that is not obvious to an operator.

Given that the race condition is only relevant first minutes of the pod lifetime, an alternative could be to simply limit the duration the daemon is alive. Let's say run it for 10 minutes.

ibabiankou:
@whynowy I've updated the PR to simply do 10 iterations, which should cover the case you described. Please let me know if there is anything else I should do to get this PR merged 😊

whynowy (commented):
@daniel-codefresh - do you see if there's any issue with this?

daniel-codefresh:
> @daniel-codefresh - do you see if there's any issue with this?

I think we should be fine with a 10 mins daemon lifetime 👍

whynowy (approved):
Thanks!

Piroddi:
@ibabiankou correct me if I'm wrong. But does this the daemon only runs for 10minutes? So if after 10 minutes someone deletes the webhook manually, it will only be recreated once this pod restarts ?

whynowy:
@ibabiankou @daniel-codefresh @Piroddi - per the link below, the rate limit is 5,000 per hour, is that correct? If that's the case, 1 request/min shouldn't be a big deal. Did I miss anything?

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-personal-accounts

whynowy:
@ibabiankou @daniel-codefresh @Piroddi - per the link below, the rate limit is 5,000 per hour, is that correct? If that's the case, 1 request/min shouldn't be a big deal. Did I miss anything?

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-personal-accounts

Stephan202:
(Colleague of @ibabiankou here:) @whynowy IIRC there's one such polling process per Github repository, and we have many of those, with more on the way.

whynowy:
> @whynowy we still using the v1.6 (Nats Streaming), therefore we needed 4 webhooks per repo across 25 repos (4 * 25 * 60 = 6000/hr). So we are getting rate-limited by Github.
> 
> Its looks like this PR just loops once a minute for 10 minutes then it stops. My worry is that if someone accidentally changes the webhook manually after the initial 10 min it will never be reconciled (I could be wrong through :) )
> 
> What about if we made the timer an ENV variable that the user can override (can leave default as 60s). I'm currently running a fork of argo-events which only loops once an hour.

Sorry I have missed the discussion here. I think making it configurable is an option.

PR #1941, opened 2022-05-10
feat: add support of submit from existing resource (#1908)

### LINKED ISSUE #1908: Support for submitting from an existing resource on ArgoWorkflow Trigger

I would like to submit an existing WorkFlow or CronWorkFlow from an Argo Workflow Trigger. Equivalent to this command.

```
argo submit --from cronwf/my-cron-wf
```

Currently the [submit code](https://github.com/argoproj/argo-events/blob/master/sensors/triggers/argo-workflow/argo-workflow.go#L137) does not appear to have this feature. My idea is to add a `from` field in `template.argoWorkflow` or elsewhere, and if there is a `from` field, add the `from` as an argument to the argo cli instead of loading a temporary file.

---

shirou:
This PR fixes #1908 

This PR adds the feature about to submit from existing resources to the argo workflow trigger.

Specify the resource by adding "from" in "metadata" as shown below. In this example, `argo submit --from cronwf/special-trigger` is executed.

```
resource:
  # does not need kind
  metadata:
    name: special-trigger
    from: cronwf
````

I propose this format as a first step. Please give me your comments.

(I tried to add tests, but the existing tests seemed to be hard-coded and adding test makes massive changes. So I don't added any test)

Checklist:

### DIFF
--- a/docs/sensors/triggers/argo-workflow.md
+++ b/docs/sensors/triggers/argo-workflow.md
@@ -60,14 +60,17 @@ Although the sensor defined above lets you trigger an Argo workflow, it doesn't 
 provided by the Argo CLI such as,
 
 1. Submit
-2. Resubmit
-3. Resume
-4. Retry
-5. Suspend
+2. Submit --from
+3. Resubmit
+4. Resume
+5. Retry
+6. Suspend
+7. Terminate
+8. Stop
 
 To make use of Argo CLI operations, The sensor provides the `argoWorkflow` trigger template,
 
         argoWorkflow:
-          operation: submit  # submit, resubmit, resume, retry, suspend, terminate or stop
+          operation: submit  # submit, submit-from, resubmit, resume, retry, suspend, terminate or stop
 
 Complete example is available [here](https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/sensors/special-workflow-trigger.yaml).
--- a/controllers/sensor/validate.go
+++ b/controllers/sensor/validate.go
@@ -179,7 +179,7 @@ func validateArgoWorkflowTrigger(trigger *v1alpha1.ArgoWorkflowTrigger) error {
 	}
 
 	switch trigger.Operation {
-	case v1alpha1.Submit, v1alpha1.Suspend, v1alpha1.Retry, v1alpha1.Resume, v1alpha1.Resubmit, v1alpha1.Terminate, v1alpha1.Stop:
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by whynowy — COMMENTED** [review]
Thanks for submitting the PR!

I think it might be better to simply extend the `operation` with `submit-from-wf` and `submit-from-cronwf`, that way it would make the current `submit` logic clear, also it won't have much code to support the new operations. What do you think?

**Comment by shirou** [comment]
Thank you for the comment!

From my understanding, the `operation` is comes from ArgoWorkflow operation. I was afraid that there would be a difference with ArgoWorkflow. However, as you said, it would be simpler to define the `submit-from` operation in ArgoEvents on its own.

I would like to change your idea a little like following. Since it is natural for k8s to specify "kind" in "resource", add `submit-from` operation and get wf or cronwf from the kind.

```
argoWorkflow:
  operation: submit-from
  source:
    resource:
      kind: CronWorkflow
      metadata:
        name: special-trigger
```

If it is ok, I will update this PR. Thank you!

**Comment by whynowy** [comment]
> 

This is a good idea!

**Comment by shirou** [comment]
I have updated. Now much simpler!

**Review by whynowy — COMMENTED** [review]
It looks good to me! Could you also add it to https://github.com/argoproj/argo-events/blob/master/docs/sensors/triggers/argo-workflow.md#argo-cli?

**Comment by shirou** [comment]
Thanks for pointing that out! updated.

**Comment by shirou** [comment]
Test failed, but it seems not related to the commit. And I can not re-run the test. I'm sorry to bother you, but could you re-run it? Thank you.

File: docs/sensors/triggers/argo-workflow.md [hunk 1]
```diff
@@ -68,6 +68,6 @@ provided by the Argo CLI such as,
 To make use of Argo CLI operations, The sensor provides the `argoWorkflow` trigger template,
 
         argoWorkflow:
-          operation: submit  # submit, resubmit, resume, retry, suspend, terminate or stop
+          operation: submit  # submit, submit-from, resubmit, resume, retry, suspend, terminate or stop
```
whynowy, line 71: Line 62 - 66 as well, also, appreciate if you could add the other missing operations there (terminate, stop...).

shirou: updated!

whynowy (approved):
LGTM, thanks for adding this!

shirou:
Thank you so much!

Issue #1946, opened 2022-05-10
8 Uses of deprecated API can be used to cause DoS in user-facing endpoints


whynowy:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-df5ac057-7fff-2abb-5057-a5a5c658f771"><div dir="ltr" style="margin-left:0pt;" align="left">

Severity | Medium
-- | --
Difficulty | Medium
Target |  

[...]

PR #1966, opened 2022-05-13
chore: discontinue using ioutil

whynowy:

1. Stop using `ioutil`;
2. use limit reader to prevent large content caused crash.

Fixes: #1946

Checklist:

### DIFF
--- a/eventsources/sources/gitlab/start.go
+++ b/eventsources/sources/gitlab/start.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/json"
-	"io/ioutil"
+	"io"
 	"math/big"
 	"net/http"
 	"reflect"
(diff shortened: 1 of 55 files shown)


sarabala1979 (approved):
LGTM

eventsources/sources/gitlab/start.go
```diff
@@ -86,8 +86,8 @@ func (router *Router) HandleRoute(writer http.ResponseWriter, request *http.Requ
 			return
 		}
 	}
-
-	body, err := ioutil.ReadAll(request.Body)
+	request.Body = http.MaxBytesReader(writer, request.Body, 65536)
```
daniel-codefresh, line 89: @whynowy sorry for the delay, I am just now seeing this - Does that mean we are no longer supporting gitlab payloads above 65KB? (same question goes for bitbucket, bitbucket server and webhook eventsources)

daniel-codefresh: @whynowy and btw do we have a different behavior atm in github? I wasn't able to find the same change in github eventsource

daniel-codefresh: From my experience, webhook payloads can grow up to 1MB and even more in some cases, especially with monorepos. I think the best solution would be to allow the users configure the max payload size by themselves (and set some default value in case they don't). I suggest to add an optional `MaxPayloadSize` field under `WebhookContext` which is used by all of the git/webhook related eventsources, wdyt @whynowy ?

whynowy: This is a good idea @daniel-codefresh .

daniel-codefresh: I've created a new issue on the subject and assigned to myself: #2088

whynowy: Much appreciated!

PR #1978, opened 2022-05-17
feat: add support to parse QueryParameter and PostForm on webhook eve…

shirou:
…ntsource

Webhook eventsource can only parse JSON strings in a request body. Therefore, when using the GET method, the body does not exist and an error occurred.

This PR changes the string to parse for each method and content-type.

And also add `application/x-www-form-urlencoded` parse to fix https://github.com/argoproj/argo-events/discussions/1760.

(additional: add HTTP method check)

### DIFF
--- a/eventsources/sources/webhook/start.go
+++ b/eventsources/sources/webhook/start.go
@@ -17,13 +17,13 @@ limitations under the License.
 package webhook
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"time"
-
-	"go.uber.org/zap"
 
 	"github.com/argoproj/argo-events/common"
 	"github.com/argoproj/argo-events/common/logging"
@@ -33,6 +33,8 @@ import (
 	apicommon "github.com/argoproj/argo-events/pkg/apis/common"
 	"github.com/argoproj/argo-events/pkg/apis/events"
 	"github.com/argoproj/argo-events/pkg/apis/eventsource/v1alpha1"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


eventsources/sources/webhook/start.go
```diff
@@ -155,3 +162,56 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 		route: route,
 	}, controller, dispatch)
 }
+
+func GetBody(writer *http.ResponseWriter, request *http.Request, route *webhook.Route, logger *zap.SugaredLogger) (*json.RawMessage, error) {
+	request.Body = http.MaxBytesReader(*writer, request.Body, 65536)
```
whynowy, line 167: Shall this move the `default` content-type section below? Does it make any difference for the rest?

shirou: Thank you for the comment. Yes, we can move to inside the `case` section. However, if we want to add more content-type like `text/csv`, this method will be used. So I want to make this function separated.

But I don't have a strong motivation, so I can move to inside.

whynowy: Let's move inside for now, that could be changed if we support more use cases in the future.

shirou: Updated!

whynowy: hmmm... there might be some misunderstanding :) I meant to move the line below closer to the place that `request.Body` gets read.

```
request.Body = http.MaxBytesReader(*writer, request.Body, 65536)
```

shirou: Ah, sorry for my mistake. I revert the function move change and move the `MaxBytesReader` lines to inside default case.

whynowy (approved):
LGTM, thanks for the contribution!

PR #1980, opened 2022-05-18
Add workflow controller RBAC setup in doc section

pivonroll:
Add a section in documentation on how to setup RBAC for argo-workflow controller so that it can start workflows created in argo-events namespace.

### DIFF
--- a/docs/quick_start.md
+++ b/docs/quick_start.md
@@ -2,18 +2,37 @@
 
 We are going to set up a sensor and event-source for webhook. The goal is to trigger an Argo workflow upon an HTTP Post request.
 
+## Install Argo Workflows
 Note: You will need to have [Argo Workflows](https://argoproj.github.io/argo-workflows/) installed to make this work. 
 The Argo Workflow controller will need to be configured to listen for Workflow objects created in `argo-events` namespace. 
   (See [this](https://github.com/argoproj/argo-workflows/blob/master/docs/managed-namespace.md) link.) 
-  The Workflow Controller will need to be installed either in a cluster-scope configuration (i.e. no "--namespaced" argument) so that it has visiblity to all namespaces, or with "--managed-namespace" set to define "argo-events" as a namespace it has visibility to. To deploy Argo Workflows with a cluster-scope configuration you can use this installation yaml file:
+  The Workflow Controller will need to be installed either in a cluster-scope configuration (i.e. no "--namespaced" argument) 
+  so that it has visibility to all namespaces, or with "--managed-namespace" set to define "argo-events" as a namespace it has visibility to.
 
-        kubectl apply -n argo -f https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/install.yaml
+  1. Install Argo workflows with a cluster-scope configuration without database and artifact repository
 
+    kubectl apply -n argo -f https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/install.yaml
+
+  2. Install Argo workflows with a cluster-scope configuration with database and artifact repository (for local experimentation)
+
+    kubectl apply -n argo -f https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/quick-start-postgres.yaml
+
+  You will also need to set up RBAC to allow argo-workflow controller to access argo-events namespace.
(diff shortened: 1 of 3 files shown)


manifests/workflow-controller-cluster-wide-rbac.yaml
```diff
@@ -0,0 +1,129 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
whynowy, line 1: Does workflow repository has these rbac yaml files? If it has, we should only reference them in the doc.

radovanEmbroker: There are cluster-scope rbac settings available at argo-workflows, but in their original form they will not be bound to argo namespace.
What is possible is to create kustomization.yaml like this:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/cluster-install/workflow-controller-rbac/workflow-controller-clusterrole.yaml
  - https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/cluster-install/workflow-controller-rbac/workflow-controller-clusterrolebinding.yaml

patches:
  - patch: |-
      - op: add
        path: /subjects/0/namespace
        value: argo
    target:
      kind: ClusterRoleBinding
```

pivonroll: @whynowy 
There are cluster-scope rbac settings available at argo-workflows, but in their original form they will not be bound to argo namespace.
What is possible is to create kustomization.yaml like this:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/cluster-install/workflow-controller-rbac/workflow-controller-clusterrole.yaml
  - https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/cluster-install/workflow-controller-rbac/workflow-controller-clusterrolebinding.yaml

patches:
  - patch: |-
      - op: add
        path: /subjects/0/namespace
        value: argo
    target:
      kind: ClusterRoleBinding
```

docs/quick_start.md
```diff
@@ -2,46 +2,65 @@
...
-  The Workflow Controller will need to be installed either in a cluster-scope configuration (i.e. no "--namespaced" argument) so that it has visiblity to all namespaces, or with "--managed-namespace" set to define "argo-events" as a namespace it has visibility to. To deploy Argo Workflows with a cluster-scope configuration you can use this installation yaml file:
+  The Workflow Controller will need to be installed either in a cluster-scope configuration (i.e. no "--namespaced" argument) 
+  so that it has visibility to all namespaces, or with "--managed-namespace" set to define "argo-events" as a namespace it has visibility to.
 
-        kubectl apply -n argo -f https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/install.yaml
+  1. Install Argo workflows with a cluster-scope configuration without database and artifact repository
 
+    kubectl apply -n argo -f https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/install.yaml
+
+  2. Install Argo workflows with a cluster-scope configuration with database and artifact repository (for local experimentation)
```
whynowy, line 16: The quick start is intend to demonstrate how it works, so I think we don't need to give too many workflow installation options here, the 1st one is good enough.

docs/quick_start.md
```diff
@@ -2,46 +2,65 @@
...
+
+    kubectl apply -n argo -f https://raw.githubusercontent.com/argoproj/argo-workflows/master/manifests/quick-start-postgres.yaml
+
+  You will also need to set up RBAC to allow argo-workflow controller to access argo-events namespace.
```
whynowy, line 20: Why is following required? The workflow cluster scope installation should be able to cover all of these already.

pivonroll: Because even in cluster scope installation a workflow is not started, it is just created and sits dormant.
I believe that people would like to have a workflow which is executed as it gets created.

sarabala1979: yes, Cluster scope should pick up all namespace workflows.

PR #1989, opened 2022-05-23
fix: jetstream statefulset resource setting

obanby:
- fixed incorrect resource setting on buildStatefulSetSpec

Checklist:


whynowy (approved):
LGTM, thanks for reporting and fixing it!

Issue #2000, opened 2022-05-27
Support Redis auth ACLs


tomelliff:
**Is your feature request related to a problem? Please describe.**
I'd like Argo Events to be able to connect to a Redis cluster using [ACLs](https://redis.io/docs/manual/security/acl/). This needs a username to be provided as well as a password.

Importantly, [MemoryDB](https://aws.amazon.com/memorydb/) only supports authentication via Redis ACLs and doesn't support older password based authentication.

**Describe the solution you'd like**
The option to pass in a username via config and/or environment variables (`REDIS_USERNAME`)

**Describe alternatives you've considered**
Not using MemoryDB or Redis with ACLs. We want something that handles faults without losing data though.

**Additional context**
N/A

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2007, opened 2022-05-30
Draft: Add support for Redis ACL auth

tomelliff:
Adds support for Redis ACL auth, fixing https://github.com/argoproj/argo-events/issues/2000.

Checklist:


tomelliff:
I've also not changed `USERS.md` as per the checklist on the PR template because unfortunately my organisation requires explicit sign off for anything that uses the company name (eg I can't commit with my company email address or use company resources) so I'd have to run it through legal first. I don't know if that's a blocker on this being accepted though?

PR #2091, opened 2022-07-21
fix: add SCRAM support in Kafka triggers

avasiliev:
Checked this change in our deployment against properly configured Kafka cluster, and it actually fixes argoproj#2087.
Have no idea how to check this configuration issue in unit tests, so didn't touch them in the PR.

### DIFF
--- a/sensors/triggers/kafka/kafka.go
+++ b/sensors/triggers/kafka/kafka.go
@@ -67,6 +67,11 @@ func NewKafkaTrigger(sensor *v1alpha1.Sensor, trigger *v1alpha1.Trigger, kafkaPr
 		if kafkatrigger.SASL != nil {
 			config.Net.SASL.Enable = true
 			config.Net.SASL.Mechanism = sarama.SASLMechanism(kafkatrigger.SASL.GetMechanism())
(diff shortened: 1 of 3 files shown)


sensors/triggers/kafka/kafka.go
```diff
@@ -27,6 +27,7 @@ import (
 
 	"github.com/argoproj/argo-events/common"
 	"github.com/argoproj/argo-events/common/logging"
+	eskafka "github.com/argoproj/argo-events/eventsources/sources/kafka"
```
whynowy, line 30: It seems weird to import a pkg from sources, could you refactor a bit to move the `scram_client.go` file to `common`?

kmeaw: I have moved `scram_client.go` to common. I have also renamed symbols SHA{256,512} to SHA{256,512}New to avoid confusion - one might expect `common.SHA512` to be a structure containing some hash metainformation, `common.SHA512New` represents a function which constructs a new `hash.Hash`.
Please tell me if you have thought of better names for them.

whynowy (approved):
Thanks!

PR #2092, opened 2022-07-21
feat: AWS Temporary credential support for SQS eventsource

### LINKED ISSUE #1901: Add support for temporary AWS IAM credentials via AWS STS in SQS Eventsource
We're trying to use Argo-Events for executing sensors based on SQS messages. Our infra provides temporary (AWS STS generated) IAM credentials stored in a k8s secret and are updated when the credentials expire. Currently Argo-Events only supports static credentials via k8s secret which is not permitted in our environment.

I propose an enhancement to the existing static credentials mechanism and add support for specifying session token (needed for temporary credentials) to make it work with temporary credentials. The event source definition could be like following (adds an additional field for sessionToken).

```apiVersion: argoproj.io/v1alpha1
kind: EventSource
metadata:
  name: sqs-es-argo
  namespace: argo-events
spec:
  sqs:
    test-msg:
      jsonBody: true
      region: us-east-1
      queue: sqs-es-argo
      waitTimeSeconds: 20
      accessKey:
        key: accesskey
        name: aws-secret
      secretKey:
        key: secretkey
        name: aws-secret
      sessionToken:
        key: sessiontoken
        name: aws-secret
```

 When the credentials expire, check the volume mount again for possibly refreshed credentials and re-try getting the messages from SQS. If a session token is not specified in the yaml, assume the credentials to be static.

If this sounds acceptable, I can start working on creating a PR to address this.

Using IRSA (IAM role for service account) is not permitted by our infra and we do not want Argo-Events to assume-role.

**Additional context**

---

harshdeep-23:
Checklist:

fixes #1901


eventsources/sources/awssqs/start.go
```diff
@@ -112,6 +111,7 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 		messages, err := fetchMessages(ctx, sqsClient, *queueURL.QueueUrl, 10, sqsEventSource.WaitTimeSeconds)
 		if err != nil {
 			log.Errorw("failed to get messages from SQS", zap.Error(err))
+			_, sqsClient, err = refreshSessionAndSqsClientIfCredentialsExpired(err, el, log)
```
daniel-codefresh, line 114: What would happen in case `refreshSessionAndSqsClientIfCredentialsExpired` fails? It would return `nil` as the second return value which means `sqsClient` would become nil. This would crash the process because of a nil pointer dereference the next time it tries to call `fetchMessages`..

eventsources/sources/awssqs/start.go
```diff
@@ -123,6 +123,7 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 				})
 				if err != nil {
 					log.Errorw("Failed to delete message", zap.Error(err))
+					_, sqsClient, err = refreshSessionAndSqsClientIfCredentialsExpired(err, el, log)
```
daniel-codefresh, line 126: same issue here. We need to verify that the call to `refreshSessionAndSqsClientIfCredentialsExpired` was successful before reassigning the value of `sqsClient`

daniel-codefresh (changes requested):

whynowy:
I'm trying to understand this PR:

1. It added a new filed for `sessionToken`, referencing a secret, which might be expired very soon;
2. If the sessionToken expired, it automatically send a request to generate a new token, and use it for connection, while the new token is not persisted anywhere;
3. Every time the eventsource pod restarts, it repeats the same process.

Am I correct?

harshdeep-23:
@whynowy Generating new credentials is not handled by the eventsource. Rather it is assumed that if the temporary credentials have expired, there is some external process which keeps updating the k8s secret with a valid credential. When the eventsource encounters credentials expired error from AWS, it tries to read the mounted credentials again (which might have changed by the external process). The issue has more details - https://github.com/argoproj/argo-events/issues/1901

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -653,6 +653,8 @@ type SQSEventSource struct {
 	// Endpoint configures connection to a specific SQS endpoint instead of Amazons servers
 	// +optional
 	Endpoint string `json:"endpoint" protobuf:"bytes,12,opt,name=endpoint"`
+	// SessionToken refers to K8s secret containing AWS temporary credentials(STS) session token
+	SessionToken *corev1.SecretKeySelector `json:"sessionToken,omitempty" protobuf:"bytes,13,opt,name=sessionToken"`
```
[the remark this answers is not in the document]
harshdeep-23: Added

eventsources/common/aws/aws.go
```diff
@@ -53,9 +54,17 @@ func GetAWSCredFromVolume(access *corev1.SecretKeySelector, secret *corev1.Secre
 	if err != nil {
 		return nil, errors.Wrap(err, "can not find secret key")
 	}
+
+	token, err := common.GetSecretFromVolume(sessionToken)
+	if err != nil {
+		log := logging.NewArgoEventsLogger()
+		log.Warnf("Failed to fetch AWS session token from volume: %s", err)
```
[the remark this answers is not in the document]
harshdeep-23: I've made logging this warning only when sessionToken is specified in the eventbus spec.

eventsources/sources/awssqs/start.go
```diff
@@ -185,3 +186,43 @@ func fetchMessages(ctx context.Context, q *sqslib.SQS, url string, maxSize, wait
 	}
 	return result.Messages, nil
 }
+
+func createAWSSession(el *EventListener, log *zap.SugaredLogger) (*session.Session, error) {
```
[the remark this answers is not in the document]
harshdeep-23: Done

eventsources/sources/awssqs/start.go
harshdeep-23: Agreed. Added a nil check before reassigning sqsClient

eventsources/sources/awssqs/start.go
harshdeep-23: Agreed. Added a nil check before reassigning sqsClient

eventsources/sources/awssqs/start.go
```diff
@@ -185,3 +192,44 @@ func fetchMessages(ctx context.Context, q *sqslib.SQS, url string, maxSize, wait
 	}
 	return result.Messages, nil
 }
+
+func (el *EventListener) createAWSSession(log *zap.SugaredLogger) (*session.Session, error) {
+	sqsEventSource := &el.SQSEventSource
+	awsSession, err := awscommon.CreateAWSSessionWithCredsInVolume(sqsEventSource.Region, sqsEventSource.RoleARN, sqsEventSource.AccessKey, sqsEventSource.SecretKey, sqsEventSource.SessionToken)
+	if err != nil {
+		log.Errorw("Error creating AWS credentials", zap.Error(err))
```
whynowy, line 200: Since error is being returned, no need to print it out here - there's already particular log message "failed to create aws session for" wrapped in the returned error to help debug.

eventsources/sources/awssqs/start.go
```diff
@@ -112,6 +111,10 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 		messages, err := fetchMessages(ctx, sqsClient, *queueURL.QueueUrl, 10, sqsEventSource.WaitTimeSeconds)
 		if err != nil {
 			log.Errorw("failed to get messages from SQS", zap.Error(err))
+			_, newSqsClient, err := el.refreshSessionAndSqsClientIfCredentialsExpired(err, log)
```
whynowy, line 114: Can we check the error here, if the error is `expired`, reuse the logic above to get a new client? You can wrap the code above to a function.

eventsources/common/aws/aws.go
```diff
@@ -53,9 +54,20 @@ func GetAWSCredFromVolume(access *corev1.SecretKeySelector, secret *corev1.Secre
 	if err != nil {
 		return nil, errors.Wrap(err, "can not find secret key")
 	}
+
+	var token string
+	if sessionToken != nil {
+		token, err = common.GetSecretFromVolume(sessionToken)
+		if err != nil {
+			log := logging.NewArgoEventsLogger()
+			log.Warnf("Failed to fetch AWS session token from volume: %s", err)
```
whynowy, line 63: Do not print the error but let the returned error to be handled by the invoker.

I don't think this should be a warning, instead, since it's configured (not nil), then it should be honored.

whynowy (commented):
What do you think if we use another approach, that provide an option to configure how ofter it reloads the secrets?

eventsources/common/aws/aws.go
harshdeep-23: Done

eventsources/sources/awssqs/start.go
harshdeep-23: I've made the change to check the error here

eventsources/sources/awssqs/start.go
harshdeep-23: Done

whynowy:
> What do you think if we use another approach, that we provide an option to configure how ofter it reloads the secrets?

@harshdeep-23 - what do you think?

harshdeep-23:
> > What do you think if we use another approach, that we provide an option to configure how ofter it reloads the secrets?
> 
> @harshdeep-23 - what do you think?

@whynowy I'm assuming you mean like a configurable way to periodic reload the secrets. If we go with that, then it might happen that the credentials expire and the requests to pull from SQS keep failing until the next secret reload cycle. We would have to keep the reload frequency very small to quickly recover if the credentials expire. I think it would be an overhead to keep a thread/coroutine continuously reloading secrets (like every X seconds) vs reloading every few mins/hours when the credentials actually expire.

eventsources/sources/awssqs/start.go
```diff
@@ -112,6 +103,15 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 		messages, err := fetchMessages(ctx, sqsClient, *queueURL.QueueUrl, 10, sqsEventSource.WaitTimeSeconds)
 		if err != nil {
 			log.Errorw("failed to get messages from SQS", zap.Error(err))
+			awsError, ok := err.(awserr.Error)
+			if ok && awsError.Code() == "ExpiredToken" && el.SQSEventSource.SessionToken != nil {
+				log.Info("credentials expired, reading credentials again")
+				newSqsClient, err := el.createSqsClient()
+				if newSqsClient != nil && err != nil {
```
whynowy, line 110: I assume it should be `err == nil`, right? Please also print the log if err is not nil.

All the reset looks good to me.

harshdeep-23: Oops, that was a stupid mistake. Fixed now.

daniel-codefresh (approved):
LGTM

whynowy (approved):

Issue #2086, opened 2022-07-18
Argo Event Contoller 1.7.1 Crash with Nill pointer on Jetstream Install


zamedic:
Applying the Jetstream Example eventbus causes a crash on Argo events 1.7.1 , kubernetes: v1.24.1

Argo events was installed using Helm Chart 

The root cause of the problem is that during the helm install, no jetstream versions were specified, resulting in an empty config map. Ideally,  the event controller should not crash but rather catch the null value and show an error, else the event controller can be brought down with 1 bad config. 

argo-events-controller-manager Config Map:
```
apiVersion: v1
data:
  controller-config.yaml: |
    eventBus:
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: argo-events
    meta.helm.sh/release-namespace: argo-events
  creationTimestamp: "2022-07-15T14:16:54Z"
  labels:
    app.kubernetes.io/instance: argo-events
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: argo-events-controller-manager
    app.kubernetes.io/part-of: argo-events
    helm.sh/chart: argo-events-2.0.3
  name: argo-events-controller-manager
  namespace: argo-events
  resourceVersion: "3866"
  uid: 66fe0cd7-96d9-4459-9c42-2422d5c9de82
```

Crash: 
```I0718 07:37:26.381000       1 leaderelection.go:248] attempting to acquire leader lease argo-events/argo-events-controller...
I0718 07:37:41.485998       1 leaderelection.go:258] successfully acquired lease argo-events/argo-events-controller
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x17235ce]

goroutine 548 [running]:
github.com/argoproj/argo-events/controllers/eventbus/installer.(*jetStreamInstaller).Install(0xc000634de0, {0x35f1e88, 0xc000634c90})
	/home/runner/work/argo-events/argo-events/controllers/eventbus/installer/jetstream.go:81 +0xae
github.com/argoproj/argo-events/controllers/eventbus/installer.Install({0x35f1e88, 0xc000634c90}, 0xc000c7ec60, {0x3645ff8, 0xc000d6f720}, 0xc000100800, 0xc000ccf870)
	/home/runner/work/argo-events/argo-events/controllers/eventbus/installer/installer.go:32 +0x1c3
[...]

PR #2103, opened 2022-07-24
fix: added controller config validation

daniel-codefresh:

### DIFF
--- a/controllers/config.go
+++ b/controllers/config.go
@@ -116,3 +116,15 @@ func LoadConfig(onErrorReloading func(error)) (*GlobalConfig, error) {
 	})
 	return r, nil
 }
+
+func ValidateConfig(config *GlobalConfig) error {
+	if len(config.supportedJetStreamVersions()) == 0 {
+		return fmt.Errorf("no jetstream versions were provided in the controller config")
+	}
+
+	if len(config.supportedSTANVersions()) == 0 {
+		return fmt.Errorf("no stan versions were provided in the controller config")
+	}
+
+	return nil
+}
--- a/controllers/cmd/start.go
+++ b/controllers/cmd/start.go
@@ -34,11 +34,16 @@ const (
 func Start(namespaced bool, managedNamespace string) {
 	logger := logging.NewArgoEventsLogger().Named(eventbus.ControllerName)
 	config, err := controllers.LoadConfig(func(err error) {
-		logger.Errorf("Failed to reload global configuration file", zap.Error(err))
+		logger.Errorw("Failed to reload global configuration file", zap.Error(err))
 	})
 	if err != nil {
 		logger.Fatalw("Failed to load global configuration file", zap.Error(err))
 	}
+
+	if err = controllers.ValidateConfig(config); err != nil {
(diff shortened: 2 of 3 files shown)


daniel-codefresh:
Fixes: #2086

controllers/config.go
```diff
@@ -116,3 +116,15 @@ func LoadConfig(onErrorReloading func(error)) (*GlobalConfig, error) {
 	})
 	return r, nil
 }
+
+func ValidateConfig(config *GlobalConfig) error {
+	if len(config.supportedJetStreamVersions()) == 0 {
+		return fmt.Errorf("no jetstream versions were provided in the controller config")
```
tooptoop4, line 122: won't this error always throw in case u are using just STAN?

whynowy: The controller crashes because there's no `EventBusConfig.NATS` or `EventBusConfig.JetStream` is `nil`, I feel like we only need to add some check along with error returning - the error should be displayed in the EventBus status field. WDYT?

daniel-codefresh: @whynowy As I see it, there are 2 approaches we can go with: either fail the controller on startup or reflect the error somehow during reconcile. I'm ok with both approaches but I think it would probably make more sense to fail on startup when the controller is provided with an invalid config (as any service would fail when provided with an invalid config) because it would simply prevent the controller from functioning properly. Also, today we face an invalid `eventbus` config which prevents the controller from reconciling eventbuses but in the future it might be some other config that would make the controller pod crash or prevent it from reconciling other components as well. For that reason I think it might be better to have a general place where we validate the global configuration instead of spreading those checks across different places, and to run the validation as part of the "readiness" requirements.

daniel-codefresh: @whynowy btw, checking if either `EventBusConfig.NATS` or `EventBusConfig.JetStream` are `nil` would also do the trick and prevent the nil deference crash that we have atm but `versions` is still a mandatory field for both nats streaming and jestream, and without `versions` the controller would still fail on every reconcile attempt with a `failed to get jetstream/nats streaming version, err: jetstream/nats streaming version configuration not found` error. From a user's perspective I think this error might be a bit confusing, especially when it occurs during the reconcile process, as the user might think there was something wrong with the version he provided in the `EventBus` spec and not in the controller config (which I'm not sure that most users aware about). WDYT?

daniel-codefresh: @tooptoop4 This change would only validate the global configuration that is provided to the controller. You would still be able to only use stan eventbus. We could support providing only one of jestream or stan versions in the global configuration, and fail during reconcile if a user tries to install an eventbus with no versions provided in the config.
For simplicity, I chose making both mandatory for now but we can change this as well @whynowy

daniel-codefresh: @whynowy WDYT?

whynowy: Sorry for the late response.

I think it's okay to enforce eventbus configuration in the configmap for now, since in our installation files both of them are there. We can always revisit this if anything needs to be changed.

whynowy (approved):

Issue #2076, opened 2022-07-15
The whole event payload is passed as a parameter value when Data or Context selection doesn't match


daniel-codefresh:
**Describe the bug**
When dataKey/dataTemplate/contextKey/contextTemplate do not match the event payload during the parameter value resolving process of the Sensor, the whole event payload is being passed as the parameter value instead. While this might have been an intentional behavior at the time of creation, it is seen by many as a bug ([most recent example from CNCF slack channel](https://cloud-native.slack.com/archives/C01TNKD6KL6/p1657110983700919)). In most cases, this behavior would happen because of a user error (e.g. a typo) and would trigger unwanted actions instead of showing the user an informative error message that would help him identify and fix the issue with the defined parameters.

**To Reproduce**
1. Create a Sensor object.
2. Define a parameter in one of the triggers in the Sensor spec with a dataKey/dataTemplate/contextKey/contextTemplate that won't match the event payload (e.g. event data `{"body": {"test": "payload"} }` and dataKey `body.notExists`). Note: do not define a default value for the parameter.
3. Check the result value of the parameter after triggering the action.

**Expected behavior**
The trigger should fail with an informative error message that would help the user identify and fix the issue with the defined parameters.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

daniel-codefresh:
@whynowy @VaibhavPage do you see any complications or risks in changing that behavior?

PR #2112, opened 2022-07-27
fix: return error in case param value resolving fails

daniel-codefresh:


daniel-codefresh:
Fixes: #2076

whynowy (approved):
This looks good to me, I think returning error is right.

Issue #2110, opened 2022-07-26
Bitbucket event source querying bitbucket api too often


dave08:
After adding more than 20 repos using the same credentials, I start getting `429 Too Many Requests` which prevents webhooks from being created.

**Environment (please complete the following information):**
 - Kubernetes: 1.21
 - Argo: 3.3.6
 - Argo Events: v1.6.3

@daniel-codefresh 

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

daniel-codefresh:
@whynowy I think I have a solution for our race condition issue with `deleteHookOnFinish` which would help us get rid of the webhooks daemon mitigation once and for all, but by looking at this [comment on the PR related to github eventsource webhooks daemon](https://github.com/argoproj/argo-events/pull/1930#issuecomment-1139344304) it seems to me that for some users the mitigation has become a feature. I'm wondering if we should keep the daemon anyway and provide an option to enable/configure it in case it is needed, wdyt?

whynowy:
> @whynowy I think I have a solution for our race condition issue with `deleteHookOnFinish` which would help us get rid of the webhooks daemon mitigation once and for all, but by looking at this [comment on the PR related to github eventsource webhooks daemon](https://github.com/argoproj/argo-events/pull/1930#issuecomment-1139344304) it seems to me that for some users the mitigation has become a feature. I'm wondering if we should keep the daemon anyway and provide an option to enable/configure it in case it is needed, wdyt?

What is your solution?

Make it configurable is also an option I think.

daniel-codefresh:
> > @whynowy I think I have a solution for our race condition issue with `deleteHookOnFinish` which would help us get rid of the webhooks daemon mitigation once and for all, but by looking at this [comment on the PR related to github eventsource webhooks daemon](https://github.com/argoproj/argo-events/pull/1930#issuecomment-1139344304) it seems to me that for some users the mitigation has become a feature. I'm wondering if we should keep the daemon anyway and provide an option to enable/configure it in case it is needed, wdyt?
> 
> What is your solution?
> 
> Make it configurable is also an option I think.

The solution is quite simple: 
[...]

PR #2145, opened 2022-08-20
fix: Stop Bitbucket API rate limit exhaustion by solving deleteHookOnFinish race condition

daniel-codefresh:
This PR solves #2110 and is a POC for all Git related eventsources. It solves the `deleteHookOnFinish` race condition and removes the need in "hooks manager daemon" mitigation that exists in all Git related eventsources and leads to API rate limit exhaustion.

The solution is quite simple:
The main reason that we are facing a race condition when `deleteHookOnFinish` is enabled is that the new eventsource pod is trying to use the same hook that could possibly be deleted by the old eventsource pod. My suggestion is that every new eventsource pod would "recreate" the hook incase it already exists (delete the old hook and create a new one) instead of using the previous one, so that each eventsource pod would work against a different hook subscribtion with a different ID. This way no "collision" can occur.

Checklist:

### DIFF
--- a/eventsources/sources/bitbucket/start.go
+++ b/eventsources/sources/bitbucket/start.go
@@ -187,25 +187,6 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 	time.Sleep(time.Duration(randomNum.Int64()) * time.Millisecond)
 	applyWebhooks()
 
-	// Bitbucket hooks manager daemon
-	go func() {
-		// Another kind of race conditions might happen when pods do rolling upgrade - new pod starts
-		// and old pod terminates, if DeleteHookOnFinish is true, the hook will be deleted from Bitbucket.
-		// This is a workaround to mitigate the race conditions.
-		logger.Info("starting bitbucket hooks manager daemon")
-		ticker := time.NewTicker(60 * time.Second)
-		defer ticker.Stop()
-		for {
-			select {
-			case <-ctx.Done():
-				logger.Info("exiting bitbucket hooks manager daemon")
-				return
-			case <-ticker.C:
-				applyWebhooks()
-			}
-		}
-	}()
-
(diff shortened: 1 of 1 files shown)


whynowy:
Is it going to be a problem when there are multiple replicas? It is possible that multiple webhooks with same config are created on Github.

daniel-codefresh:
> Is it going to be a problem when there are multiple replicas? It is possible that multiple webhooks with same config are created on Github.

So basically what would happen with multiple replicas is something like the following:
If we have 3 replicas R1, R2, R3 -

R1 --> random sleep --> creates new webhook
R2 --> random sleep ---------------> deletes webhook created by R1 --> creates new webhook
R3 --> random sleep -------------------------------------------------------------> deletes webhook created by R2 --> creates new webhook

If we increase the [random sleep range](https://github.com/argoproj/argo-events/blob/2a8fdbd17204d5813f5101ce70d3ad07a231ec2f/eventsources/sources/bitbucket/start.go#L186) from 2 secs to at least 5 secs then I think we should be fine 99% of the time. 
BTW, as far as I know Github doesn't except multiple webhooks with the same config and returns an error which we can ignore. Other git providers actually do except multiple webhooks with the same config (e.g. Bitbucket Cloud) which might be a bit problematic in case we accidentally create multiple webhooks and then receive duplicated events in our eventsources, but I think we face the same problem even today with the current solution when multiple replicas try to create a webhook if it doesn't exist.

TBH I think that probably the best solution would be to somehow leverage NATS to store a distributed lock which all the eventsource replicas could use during the creation and deletion of webhooks, and this way avoid all these race conditions.
But for now I think that this solution might be good enough to replace the current mitigation of the webhooks manager daemon. Also, as we discussed previously, we can still leave the daemon as a backup and let the users enable it if needed and of course configure the interval of that daemon.

whynowy (approved):

Issue #2163, opened 2022-09-03
Configure NATS max_payload for JetStream eventbus


daniel-codefresh:
**Is your feature request related to a problem? Please describe.**
ATM the max_payload size limit of NATS (1MB by default) is only configurable for STAN eventbus. I suggest to add the same support for JetStream eventbus.

**Describe the solution you'd like**
Add a `eventbus.jetstream.maxPayload` field under the JetStream eventbus spec, same as we have today for STAN.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2164, opened 2022-09-03
feat: Add option to configure NATS max_payload in JetStream eventbus

daniel-codefresh:
Solves #2163


whynowy (approved):

whynowy:
@daniel-codefresh - Sorry, I might have clicked it too fast, `max_payload` should not be exposed with this approach, instead, it should be done at https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/base/jetstream.go#L131.

Could you make the change?

daniel-codefresh:
> @daniel-codefresh - Sorry, I might have clicked it too fast, `max_payload` should not be exposed with this approach, instead, it should be done at https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/base/jetstream.go#L131.
> 
> Could you make the change?

@whynowy not sure I follow - do you mean that `max_payload` should be passed as part of [streamConfig] (https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/base/jetstream.go#L131) ? if so then I'm not sure it can be done because `max_payload` should be part of the nats server config and not part of the stream config..
Of course we can create a new `serverConfig` and allow users to pass a yaml similarly to how its done with the stream config or with the jetstream settings if thats what you mean

whynowy:
> 

```
A final control is the Maximum Size any single message may have. NATS have it's own limit for maximum size (1 MiB by default), but you can say a Stream will only accept messages up to 1024 bytes using MaxMsgSize.
```
https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive

Max message size can be configured when creating a stream, see the source code below.

https://github.com/nats-io/nats.go/blob/main/jsm.go#L104

PR #2188, opened 2022-09-13
[examples] fix filter for github sensor example

tieum:
This PR fixes the github sensor example, there was a typo on X-Github-Event (capital H), plus it's in the header part instead of the body one

Cheers


whynowy (approved):
thanks!

PR #2207, opened 2022-09-24
ci: Pin k3d to use rancher/k3s:v1.21.7-k3s1

dfarr:
Pinning the image to older version appears to stabilize the e2e tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -140,7 +140,7 @@ jobs:
       - name: Create k3d registry and cluster
         run: |
           k3d registry create e2e-registry --port 5111
-          k3d cluster create e2e --registry-use k3d-e2e-registry:5111
+          k3d cluster create e2e -i rancher/k3s:v1.21.7-k3s1 --registry-use k3d-e2e-registry:5111
           echo '127.0.0.1 k3d-e2e-registry' | sudo tee -a /etc/hosts
       - name: Run tests
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -137,9 +137,12 @@ jobs:
             ${{ runner.os }}-go-
       - name: Install k3d
         run: curl -sfL https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash &
-      - name: Start k3d cluster
-        run: k3d cluster create e2e -i rancher/k3s:v1.21.7-k3s1
+      - name: Create k3d registry and cluster
+        run: |
+          k3d registry create e2e-registry --port 5111
+          k3d cluster create e2e --registry-use k3d-e2e-registry:5111
+          echo '127.0.0.1 k3d-e2e-registry' | sudo tee -a /etc/hosts
       - name: Run tests
         run: |
-          make start
+          IMAGE_NAMESPACE=k3d-e2e-registry:5111 VERSION=${{ github.sha }} DOCKER_PUSH=true make start
           EventBusDriver=${{ matrix.driver }} make test-functional
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
   codegen:
     name: Codegen
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 5
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
@@ -135,15 +135,10 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
-      - name: Install and start K3S
-        run: |
-          curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.21.2+k3s1 INSTALL_K3S_CHANNEL=stable INSTALL_K3S_EXEC=--docker K3S_KUBECONFIG_MODE=644 sh -
-          until kubectl --kubeconfig=/etc/rancher/k3s/k3s.yaml cluster-info ; do sleep 10s ; done
-          cp /etc/rancher/k3s/k3s.yaml /home/runner/.kubeconfig
-          echo "- name: fake_token_user" >> $KUBECONFIG
-          echo "  user:" >> $KUBECONFIG
-          echo "    token: xxxxxx" >> $KUBECONFIG
-          until kubectl cluster-info ; do sleep 10s ; done
+      - name: Install k3d
+        run: curl -sfL https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash &
+      - name: Start k3d cluster
+        run: k3d cluster create e2e -i rancher/k3s:v1.21.7-k3s1
       - name: Run tests
         run: |
           make start
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -106,29 +106,10 @@ jobs:
       - run: make lint
       - run: git diff --exit-code
 
-  build-image:
-    name: Build image
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
-      - name: Build image
-        run: |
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -138,7 +133,7 @@ docs/assets/diagram.png: go-diagrams/diagram.dot
 	cd go-diagrams && dot -Tpng diagram.dot -o ../docs/assets/diagram.png
 
 .PHONY: start
-start: image
```
dfarr, line 141: I decoupled building the image from starting the infra so that `kind load docker-image` can be called in between the two steps [here](https://github.com/argoproj/argo-events/pull/2207/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR138-R139).

webhook/webhook_test.go
```diff
@@ -30,7 +30,7 @@ func fakeOptions() Options {
 		DeploymentName:  "events-webhook",
 		ClusterRoleName: "argo-events-webhook",
 		ServiceName:     "webhook",
-		Port:            443,
+		Port:            8443,
```
dfarr, line 33: The unit tests were trying to open port 443, which requires the root user to run `make test`. I changed the port to fix this.

dfarr:
I think running the tests in parallel is causing a race condition where the last process to write the kube context to the `~/.kube/config` file "wins". This is causing one of the e2e tests to succeed, and the other to fail. I will figure out a way to resolve this race condition and I suspect we should be good to go after that.

whynowy:
@dfarr - thanks for submitting the PR and appreciate your help! `k3d/k3s` is the the only tool we used across argo projects for local testing and e2e tests, could you fix the issue still based on it?

.github/workflows/ci.yaml
```diff
@@ -10,23 +10,23 @@ jobs:
...
-          key: ${{ runner.os }}-go-build-v1-${{ hashFiles('**/go.mod') }}
       - name: Setup Golang
         uses: actions/setup-go@v3.2.0
         with:
           go-version: "1.19"
-      - name: Add bins to PATH
-        run: |
-          echo /home/runner/go/bin >> $GITHUB_PATH
-          echo /usr/local/bin >> $GITHUB_PATH
+      - name: Restore go build cache
```
dfarr, line 21: From [github action/cache docs](https://github.com/actions/cache/blob/main/examples.md#go---modules. I also noticed that argo workflows is using the cache action in this manner.

.github/workflows/ci.yaml
```diff
@@ -114,34 +141,36 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Install and start K3S
```
dfarr, line 157: I stole this verbatim [from argo workflows](https://github.com/argoproj/argo-workflows/blob/master/.github/workflows/ci-build.yaml#L132-L140).

.github/workflows/ci.yaml
```diff
@@ -87,23 +88,49 @@ jobs:
...
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
       - run: make lint
       - run: git diff --exit-code
 
+  build-image:
```
dfarr, line 109: I pulled this out into an independent step so it can be executed once rather than for each item in the e2e-tests matrix.

whynowy: The idea is we also want to make it one step for local e2e testing, so I'm okay with each matrix testing to build the image for testing, even though it adds several minutes.

.github/workflows/ci.yaml
whynowy: Would prefer to use `k3d` instead of `k3s` directly, as we leverage its image registry feature when doing local development.

whynowy:
If I'm not wrong, the reason it becomes stable is just because of using a low version k3s image. If that's the case, I would prefer to just add `-i rancher/k3s:v1.21.7-k3s1` to `k3d create` cli.

The reason why `port-forward` and log stream become unstable with k3s higher versions remains unknown.

.github/workflows/ci.yaml
```diff
@@ -114,34 +141,36 @@ jobs:
...
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Install and start K3S
         run: |
-          KUBECONFIG=~/.kube/argo-events-e2e-config IMAGE_NAMESPACE=k3d-e2e-registry:5111 VERSION=${{ github.sha }} DOCKER_PUSH=true make start
+          curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.21.2+k3s1 INSTALL_K3S_CHANNEL=stable INSTALL_K3S_EXEC=--docker K3S_KUBECONFIG_MODE=644 sh -
```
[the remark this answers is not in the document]
dfarr: I think so too, let me play around with this to determine the latest version of k3s that still works. With the exception of the latest ci run which failed, the tests have been much more consistent on our fork where the [last 12 runs have all succeed](https://github.com/intuit-data-os/argo-events/actions/workflows/ci.yaml).

.github/workflows/ci.yaml
```diff
@@ -10,23 +10,23 @@ jobs:
   codegen:
     name: Codegen
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 10
```
whynowy, line 13: Codegen usually takes 3 mins, keeping it 5 should be ok.

.github/workflows/ci.yaml
```diff
@@ -114,34 +122,29 @@ jobs:
...
+          echo "    token: xxxxxx" >> $KUBECONFIG
+          until kubectl cluster-info ; do sleep 10s ; done
       - name: Run tests
-        env:
-          GOPATH: /home/runner/go
-        run: KUBECONFIG=~/.kube/argo-events-e2e-config EventBusDriver=${{matrix.driver}} make test-functional
-
+        run: |
+          make start
+          EventBusDriver=${{ matrix.driver }} make test-functional
```
whynowy, line 150: This is not right - it's testing against the `latest` tag, which is the latest image from master branch. You could simply revert this part of changes, still use k3d and use the approach like https://github.com/argoproj/argo-events/pull/2218/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR135.

dfarr: Does it matter if we tag the image as latest though? The ci pipeline isn't used for pushing images to quay.

Makefile
```diff
@@ -87,7 +87,7 @@ test:
 	go test $(shell go list ./... | grep -v /vendor/ | grep -v /test/e2e/) -race -short -v
 
 test-functional:
-	go test -v -timeout 10m -count 1 --tags functional -p 1 ./test/e2e
+	go test -v -timeout 20m -count 1 --tags functional -p 1 ./test/e2e
```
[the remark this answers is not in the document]
dfarr: I've been seeing runs that last longer than 10m and succeed, see screenshot. Maybe we can split the difference and set it to 15m?

image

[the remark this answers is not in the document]
whynowy: Sounds good.

.github/workflows/ci.yaml
whynowy: `latest` image is pushed to quay.io on master branch commit.

whynowy:
@dfarr - pin the version.

dfarr:
> @dfarr - pin the k3s version.

Whoops, to much trial and error I must have reverted that too by mistake

whynowy (approved):
LGTM, thanks!

PR #2245, opened 2022-10-13
feat: update third_party dependencies

jomach:
We use argo events with security scans. Our security scans shows that golang version is to old on the current version from nats container.


Wheedman (approved):

third_party/nats-streaming-docker/README.md
```diff
@@ -5,4 +5,4 @@ This is a partial copy of
 current
 [nats-streaming-server](https://github.com/nats-io/nats-streaming-server)
 version is
-[v0.22.1](https://github.com/nats-io/nats-streaming-server/tree/v0.22.1).
+[v0.25.2](https://github.com/nats-io/nats-streaming-server/releases/tag/v0.25.2).
```
whynowy, line 8: Thanks for updating this, also please be aware these 3rd party related docs and Dockerfile are just references, in the real installation, the nats stream image is controlled by the controller configmap, where you can specify the image version.

Dockerfile
```diff
@@ -2,13 +2,13 @@ ARG ARCH=$TARGETARCH
...
 ####################################################################################################
-FROM alpine:3.12.3 as base
+FROM alpine:3.16.2 as base
 ARG ARCH
 RUN apk update && apk upgrade && \
     apk add ca-certificates && \
     apk --no-cache add tzdata
 
-ENV ARGO_VERSION=v3.4.0
+ENV ARGO_VERSION=v3.4.1
```
whynowy, line 11: 👍

whynowy (approved):

whynowy:
@jomach - I'm also curious, what eventbus are you using? The new `JetStream` or legacy `stan` (`Nats Streaming`)?

third_party/nats-streaming-docker/README.md
jomach: Yes, you are right. I saw that after creating the pr. Actually we could drop this third-party part right ? I saw in the helm chart we can pass in a nats container version. We are actually doing that and using now the latest version of nats-streaming-server

jomach:
> JetStream

No idea to be honest. We have the newest chart 0.20. and we are using: 
   natsStreamingImage: /docker-registry/nats-streaming:0.25.2

third_party/nats-streaming-docker/README.md
whynowy: That's right.

whynowy:
> > JetStream
> 
> No idea to be honest. We have the newest chart 0.20. and we are using: natsStreamingImage: /docker-registry/nats-streaming:0.25.2

Be aware Nat Streaming will be EoL by 2023, so we have a new EventBus implemented with `JetStream`, check https://github.com/argoproj/argo-events/blob/master/docs/eventbus/jetstream.md.

jomach:
@whynowy The Build failed.

whynowy:
> @whynowy The Build failed.

Transient network issue.

PR #2273, opened 2022-10-25
fix: payload serialization in sensor. Fixes #2272

### LINKED ISSUE #2272: Parameterization a block of k8s resource (set of key values)
**Describe the bug**
I am trying to parametrize an entire block of k8s trigger using payload. the payload converted to string with backslashes and can't be injected to overwrite the block. Workflows CRD getting `cannot unmarshall spec: cannot restore struct from: string.`

Here is an example of the sensor:
```apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: example
  namespace: argocd
spec:
  template:
    serviceAccountName: argo-wf
  dependencies:
    - name: example-dep
      eventSourceName: example-event-source
      eventName: event-test
  triggers:
    - template:
        name: example-template
        k8s:
          operation: create
          source:
            resource:
              kind: Workflow
              metadata:
                generateName: injected-block-
              spec:
                entrypoint: main
                arguments:
                  parameters:
                    - name: message
                      value: hello world
                serviceAccountName: argo-wf  
                templates:
                  - name: main
                    container:
                      image: docker/whalesay:latest
                      command: [cowsay]
                      args:
                        - "hello world"
          parameters:
            - src:
                dependencyName: example-dep
                dataKey: body.injected
              dest: spec.templates.0.container```

After debugging the sensor I have found that in line 124 sensors/triggers/params.go get serialized json with backslashes:
```
func ApplyParams(jsonObj []byte, params []v1alpha1.TriggerParameter, events map[string]*v1alpha1.Event) ([]byte, error) {
...
tmp, err := sjson.SetBytes(jsonObj, param.Dest, *value)
...

*values values is:
```
Actual state: "{\"image\":\"docker/whalesay:latest\",\"command\":[\"cowsay\"],\"args\":[\"injected\"]}"
 Desired state: "{"image":"docker/whalesay:latest","command":["cowsay"],"args":["injected"]}"
```

The source of that serialization in github.com/cloudevents/sdk-go/v2 that used in eventbus/stan/sensor.(*STANTriggerConn).Subscribe.func1. 

Better option would be to use this differently and use sjson.SetRawBytes function instead of sjson.SetBytes in sensors/triggers/params.go

**To Reproduce**
1. Deploy the sensor from above.
2. send the following json body to the event source:
```
{
  "injected": {
    "image": "docker/whalesay:latest",
    "command": ,
    "args": [
      "injected"
    ]
  }
}
```
**Expected behavior**
Create a workflow that marshal spec right.

**Screenshots**
Screen Shot 2022-10-25 at 10 37 58

 - Kubernetes: v1.24.0
 - ArgoWorkflows: v3.4.2
 - Argo Events: v1.7.3

**Additional context**

---

GoshaDo:
Fixes #2272
Signed-off-by: GoshaDo <goshatoo@gmail.com>


sensors/triggers/params.go
```diff
@@ -274,10 +293,15 @@ func getValueWithTemplate(value []byte, templString string) (string, error) {
...
 // or an error if it does not exist.
-func getValueByKey(value []byte, key string) (string, error) {
+func getValueByKey(value []byte, key string) (string, string, error) {
 	res := gjson.GetBytes(value, key)
 	if res.Exists() {
-		return res.String(), nil
+		if res.Type.String() == "String" {
+			return res.String(), res.Type.String(), nil
+		} else if res.Type.String() == jsonType {
+			return res.Raw, res.Type.String(), nil
```
whynowy, line 302: What will happen if we always use `Raw`?

sensors/triggers/params_test.go
```diff
@@ -274,7 +274,7 @@ func TestResolveParamValue(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			result, err := ResolveParamValue(test.source, events)
+			result, _, err := ResolveParamValue(test.source, events)
```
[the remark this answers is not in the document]
GoshaDo: I think that GoLand have changed it when I executed the unit tests, and I didn't noticed.
running `git config --local core.fileMode false` on the forked repo. 
maybe its a good idea to tests it, or to write in down in contribution manual.

sensors/triggers/params.go
GoshaDo: It will brake what other function expect to get ( uni-test are failing). I wanted to save the functionality of the function and not barking anything else.

USERS.md
```diff
@@ -40,3 +40,5 @@ Organizations below are **officially** using Argo Events. Please send a PR with
 1. [Yubo](https://www.yubo.live/)
 1. [WooliesX](https://wooliesx.com.au/)
 1. [Woolworths Group](https://www.woolworthsgroup.com.au/)
+1. [Rookout](https://www.rookout.com/)
```
whynowy, line 43: Alphabetical order.

sensors/triggers/params.go
```diff
@@ -161,12 +180,13 @@ func renderEventDataAsJSON(event *v1alpha1.Event) ([]byte, error) {
 
 // helper method to resolve the parameter's value from the src
 // returns an error if the Path is invalid/not found and the default value is nil OR if the eventDependency event doesn't exist and default value is nil
-func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*v1alpha1.Event) (*string, error) {
+func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*v1alpha1.Event) (*string, string, error) {
```
whynowy, line 183: Can you add more docs about the added returning variable?

sensors/triggers/params.go
```diff
@@ -183,7 +203,7 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
 			}
 
 			if err == nil {
-				return &resultValue, nil
+				return &resultValue, "", nil
```
whynowy, line 206: Return something instead of `""`?

sensors/triggers/params.go
```diff
@@ -203,52 +223,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
 			}
 			fmt.Printf("failed to execute the src event template, falling back to key or value. err: %+v\n", err)
 		}
 		if key != "" {
-			resultValue, err = getValueByKey(eventPayload, key)
+			tmp, typ, err := getValueByKey(eventPayload, key)
+			// For block injection support
+			fmt.Printf("failed to get value by key: %+v\n", err)
```
whynowy, line 255: What is this?

How come we have so many `fmt.Printf` in the code.....

sensors/triggers/params.go
```diff
@@ -203,52 +223,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
+		return nil, "", err
 	}
-
 	// Get the value corresponding to specified key or template within event payload
 	if eventPayload != nil {
 		if tmplt != "" {
 			resultValue, err = getValueWithTemplate(eventPayload, tmplt)
 			if err == nil {
-				return &resultValue, nil
+				return &resultValue, "", nil
```
whynowy, line 248: Returning stringType by default?

sensors/triggers/params.go
gosharo: already been there. 
[https://github.com/argoproj/argo-events/blob/25f5a50df5963b0b5c6c7e0508474ade24b715ba/sensors/triggers/params.go#L240](url)
changing it's place  to after the error check

GoshaDo: already been there.
[https://github.com/argoproj/argo-events/blob/25f5a50df5963b0b5c6c7e0508474ade24b715ba/sensors/triggers/params.go#L240](https://github.com/argoproj/argo-events/pull/url)
changing it's place to after the error check

sensors/triggers/params.go
```diff
@@ -183,7 +205,7 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
 			}
 
 			if err == nil {
-				return &resultValue, nil
+				return &resultValue, errorType, nil
```
whynowy, line 208: Is this right? If there's no error, return `errorType`?

The original logic looks very tricky. Can we make it simple like this:

```
if src.Value != nil {
  return *src.Value
} else {
  eventPayload, err = json.Marshal.....
  if err != nil {
    return err
  }
  return string(eventPayload)
}
```

sensors/triggers/params.go
```diff
@@ -203,52 +225,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
-				return &resultValue, nil
+				return &resultValue, typ, nil
 			}
 			fmt.Printf("failed to get value by key: %+v\n", err)
 		}
 		// In case neither key nor template resolving was successful, fall back to the default value if exists
 		if src.Value != nil {
 			resultValue = *src.Value
-			return &resultValue, nil
+			return &resultValue, typ, nil
```
whynowy, line 266: Does the `typ` returned here have a value?

sensors/triggers/params.go
```diff
@@ -203,52 +225,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
 	case src.Value != nil:
 		// Use the default value set by the user in case the event is missing
 		resultValue = *src.Value
-		return &resultValue, nil
+		return &resultValue, stringType, nil
 	default:
 		// The parameter doesn't have a default value and is referencing a dependency that is
 		// missing in the received events. This is not an error and may happen with || conditions.
-		return nil, nil
+		return nil, errorType, nil
```
whynowy, line 232: The original return seems to be incorrect, it should return an error.

sensors/triggers/params.go
GoshaDo: have changed the error check just after json.Marshal
about the return, we should return a pointer to the casted eventPayload variable. I think best to write it in memory.

GoshaDo: and the function returns (*string, string, error)
we can't return error only...

sensors/triggers/params.go
GoshaDo: Right... I will return errorType, because we fail the function

sensors/triggers/params.go
GoshaDo: checkout the comment in lines 230-231

sensors/triggers/params.go
```diff
@@ -179,12 +200,11 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
 			} else {
 				// Default value doesn't exist so return the whole event payload
 				eventPayload, err = json.Marshal(&event)
+				if err == nil {
```
whynowy, line 203: Why moved here? It changed the behavior. when `src.Value != nil`, it originally returns directly, but now it does not.

sensors/triggers/params.go
whynowy: If that's the case, can you return `stringType` as that's the original expectation?

sensors/triggers/params.go
```diff
@@ -203,52 +223,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
+		return nil, errorType, err
 	}
-
 	// Get the value corresponding to specified key or template within event payload
 	if eventPayload != nil {
 		if tmplt != "" {
 			resultValue, err = getValueWithTemplate(eventPayload, tmplt)
 			if err == nil {
-				return &resultValue, nil
+				return &resultValue, errorType, nil
```
whynowy, line 248: Why return `errorType` when `err == nil`?

sensors/triggers/params.go
```diff
@@ -203,52 +223,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
-				return &resultValue, nil
+				return &resultValue, typ, nil
 			}
 			fmt.Printf("failed to get value by key: %+v\n", err)
 		}
 		// In case neither key nor template resolving was successful, fall back to the default value if exists
 		if src.Value != nil {
 			resultValue = *src.Value
-			return &resultValue, nil
+			return &resultValue, errorType, nil
```
whynowy, line 264: Why return `errorType`?

sensors/triggers/params.go
```diff
@@ -179,12 +200,11 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
 			} else {
 				// Default value doesn't exist so return the whole event payload
 				eventPayload, err = json.Marshal(&event)
+				if err == nil {
+					return &resultValue, errorType, nil
```
whynowy, line 204: Why return errorType?

sensors/triggers/params.go
```diff
@@ -160,8 +180,9 @@ func renderEventDataAsJSON(event *v1alpha1.Event) ([]byte, error) {
 }
 
 // helper method to resolve the parameter's value from the src
+// returns value and value type (jsonType or stringType or errorType). jsonType represent a block while stringType represent a single value.
 // returns an error if the Path is invalid/not found and the default value is nil OR if the eventDependency event doesn't exist and default value is nil
-func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*v1alpha1.Event) (*string, error) {
+func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*v1alpha1.Event) (*string, string, error) {
```
whynowy, line 185: Even you added an `errorType` in the return, but you never checked it. To make it simple, the guideline should be:

1. If there's no error, return `(xxx, xx, nil)`
2. Or return `(nil, "", err)`

sensors/triggers/params.go
whynowy: It falls back to the default value, does not fail the function.

GoshaDo: What do you suggest to return here?

sensors/triggers/params.go
GoshaDo: reverted

sensors/triggers/params.go
GoshaDo: because the we didn't resolver the param. what are you suggestions ?

sensors/triggers/params.go
GoshaDo: have changed it

sensors/triggers/params.go
GoshaDo: that what originally been there...  reverting those changes..

GoshaDo: Done

sensors/triggers/params.go
```diff
@@ -203,52 +224,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
 		if src.Value != nil {
 			fmt.Printf("failed to parse the event payload, using default value. err: %+v\n", err)
 			resultValue = *src.Value
-			return &resultValue, nil
+			return &resultValue, stringType, nil
 		}
-
 		// Otherwise, return the error
-		return nil, err
+		return nil, errorType, err
```
whynowy, line 242: There's no need to have an `errorType`, since you never checked it.

Simply `return nil, "", err`,

sensors/triggers/params.go
```diff
@@ -203,52 +224,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
 	}
-
 	// Get the value corresponding to specified key or template within event payload
 	if eventPayload != nil {
 		if tmplt != "" {
 			resultValue, err = getValueWithTemplate(eventPayload, tmplt)
 			if err == nil {
-				return &resultValue, nil
+				return &resultValue, "", nil
```
whynowy, line 249: What is the reason it returns `""` as the type instead of `stringType` when `err == nil`?

sensors/triggers/params.go
```diff
@@ -203,52 +224,51 @@ func ResolveParamValue(src *v1alpha1.TriggerParameterSource, events map[string]*
...
-				return &resultValue, nil
+				return &resultValue, typ, nil
 			}
 			fmt.Printf("failed to get value by key: %+v\n", err)
 		}
 		// In case neither key nor template resolving was successful, fall back to the default value if exists
 		if src.Value != nil {
 			resultValue = *src.Value
-			return &resultValue, nil
+			return &resultValue, "", nil
```
whynowy, line 265: Why it returns `""` as the type instead of `stringType`?

sensors/triggers/params.go
```diff
@@ -272,12 +292,18 @@ func getValueWithTemplate(value []byte, templString string) (string, error) {
...
-		return res.String(), nil
+		if res.Type.String() == stringType {
+			return res.String(), res.Type.String(), nil
+		} else if res.Type.String() == jsonType {
+			return res.Raw, res.Type.String(), nil
+		}
+		return res.Raw, res.Type.String(), nil
 	}
-	return "", fmt.Errorf("key %s does not exist to in the event payload", key)
+	return "", errorType, fmt.Errorf("key %s does not exist to in the event payload", key)
```
whynowy, line 308: no need to have an `errorType`, same reason as above.

sensors/triggers/params.go
GoshaDo: removed

sensors/triggers/params.go
GoshaDo: done

sensors/triggers/params.go
GoshaDo: done

sensors/triggers/params.go
GoshaDo: removed

sensors/triggers/params.go
```diff
@@ -272,12 +291,18 @@ func getValueWithTemplate(value []byte, templString string) (string, error) {
...
+func getValueByKey(value []byte, key string) (string, string, error) {
 	res := gjson.GetBytes(value, key)
 	if res.Exists() {
-		return res.String(), nil
+		if res.Type.String() == stringType {
+			return res.String(), res.Type.String(), nil
+		} else if res.Type.String() == jsonType {
+			return res.Raw, res.Type.String(), nil
+		}
+		return res.Raw, res.Type.String(), nil
```
whynowy, line 305: This is the default return, make it consistent as before:

`return res.String(), stringType, nil`

sensors/triggers/params.go
whynowy, line 206: `return &resultValue, stringType, nil`

sensors/triggers/params.go
gosharo: Done

PR #2292, opened 2022-11-03
fix: Consistent metricsExporterImage version in install manifests

terrytangyuan:
Checklist:


whynowy:
@terrytangyuan - nats streaming is going to be EOF, so we don't want to touch it any more, including the metrics exporter image, which might be different from before.

terrytangyuan:
Got it. Closing this then.

PR #2324, opened 2022-11-24
[issue-1863] username and password auth support for mqtt eventsource

premadk:
Signed-off-by: Prema devi Kuppuswamy <premadk@gmail.com>

Checklist:


daniel-codefresh (approved):
LGTM! thanks

whynowy (approved):
Thanks!

BebbiBraeu:
This doesn't work for me.
We just updated to argo-events 1.7.5.
But whenever I try to create a MQTT EventSource with credentials, "auth" just gets ignored and doesn't appear in the EventSource in the argo UI. Also when adding it in the UI, it gets deleted after the update.
I would guess that the schema validation fails.
Can this be checked somehow and how could this be fixed?

whynowy:
> 

@BebbiBraeu - Can you create issue with what you found (please include the spec)?

Issue #2323, opened 2022-11-22
Would be nice if argo-events support customize logging


emmayylu:
**Is your feature request related to a problem? Please describe.**
I am using argo-events in an application that runs tasks. Each task has a unique id used as a "correlation id" created by the front end. The flow looks like this: Frontend -> service A (can use correlation id to find logs)-> argo source -> argo event bus -> argo event sensor -> trigger (can use correlation id to find logs)

The problem is, lots of tasks are triggered by the front end every minute. Sometimes, some of them will run into errors in argo events source, bus, and sensor. It is hard for us to find the logs for a certain task triggered by the front end because argo-events logs don't contains the correlation id we use.

Should argo-events allow customize logging? or there is already a walkaround for this issue? 

**Describe the solution you'd like**
instead of using the eventId, I would prefer to use a customized "event id". A user should be able to define how the event source can get the correlation id.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

emmayylu:
duplicate

talebzeghmi:
related https://github.com/argoproj/argo-events/issues/1111

PR #2325, opened 2022-11-24
feat: Enable adding customized logging fields in sensor

emmayylu:
Checklist:


emmayylu:
Partially address https://github.com/argoproj/argo-events/issues/2323 will create other PRs for eventsource and eventbus after this one is merged

emmayylu:
@whynowy Hello Derek! I need this feature for a ticket I am working on at work. This has become my blocker. Could you please review it when you have time? I am very new to Golang, so it would be nice if you could provide some feedback, and I will work on making the PR better! Thank you!

pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -110,6 +110,9 @@ type SensorSpec struct {
 	// RevisionHistoryLimit specifies how many old deployment revisions to retain
 	// +optional
 	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty" protobuf:"varint,7,opt,name=revisionHistoryLimit"`
+	// LoggingFields add additional key-value pairs when logging happens
+	// +optional
+	LoggingFields []LoggingField `json:"loggingFields" protobuf:"bytes,8,opt,name=loggingFields"`
```
whynowy, line 115: Is a `map[string]string` good enough?

use `rep` instead of `opt` in the `json` part

yolu-kxs: Thanks for the feedback. I am new to go. I believe if I use `map[string]string`, I won't be able to define the fields (name and value)?

ok! I will change it! I am curious what is the difference between `opt` and `rep`?

emmayylu: Thanks for the feedback. I am new to go. I believe if I use `map[string]string`, I won't be able to define the fields (name and value)?

ok! I will change it! I am curious what is the difference between `opt` and `rep`?

whynowy: `rep` means repeated fields in protobuf.


With `map[string]string`, you do
```yaml
  loggingFields:
    a: b
    e: f
```

yolu-kxs: > `rep` means repeated fields in protobuf.
> 
> With `map[string]string`, you do
> 
> ```yaml
>   loggingFields:
>     a: b
>     e: f
> ```

oh I get what you mean now. I was initially thinking to make it look like
```
loggingFields:
- name: job_id
   value: 123
```
and you are suggesting to make it
```
loggingFields:
    job_id: abc
```
Your approach is better

emmayylu: oh I get what you mean now. I was initially thinking to make it look like
```
loggingFields:
- name: job_id
  value: 123
```
and you are suggesting to make it
```
loggingFields:
    job_id: abc
```
Your approach is better

whynowy (approved):
LGTM

PR #2353, opened 2022-12-12
fix: typo in gitlab example eventsource

jwillker:
This example refers to GitLab usage, so I think this is a typo


whynowy (approved):
Thanks!

Issue #2357, opened 2022-12-15
[Partition] Kafka Trigger


Thomasr-02:
**Describe the bug**

I cannot use round-robin on Kafka Trigger

**Additional context**

The partition parameter on Kafka trigger is mandatory following the doc, but, if we see the code, the parameter is not validated by  `validateKafkaTrigger`. it is a point. 

if we do not set the partition parameter, It is set to 0 as a default value ( Meaning that: all messages are delivered to topic on 0 partition)
if we set X on partition parameter, all message is delivery to topic on X partition, 
but if we want use, default round-robin on Kafka ( it is not supported by the argo-events)

**Expected behavior**

I hoped that by setting it as `-1` As the partition parameter, I would be removing the partition property. Thus using the default kafka’s round-robin

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

jwillker:
This will be very useful.

PR #2360, opened 2022-12-16
Implement multiple partions usage in Kafka trigger 

Thomasr-02:
Resolving this issue (https://github.com/argoproj/argo-events/issues/2357)

Option to use multiple partitions to produce messages in kafka
Changes in ( Sensor ) -> KafkaTrigger -> Partition (parameter)

Description what changed : 
>  I add to doc more information about `partition` parameter
- `0` is default value, this was not being mentioned in the documentation
- `-1` is a new parameter option to use randon partition. This includes (use a new **key**(UUID) generated by [google go lib](https://pkg.go.dev/github.com/google/UUID), because the default behavior is to hash the key of a record to get the a random partition, the second part of the solution is to not define the partition parameter for the sarama client, because we don't want to define any partitions this is chosen automatically.

Tests screenshots:
Using 3 partitions of topic: 
image

Checklist:


pkg/apis/sensor/v1alpha1/types.go
```diff
@@ -568,7 +568,10 @@ type KafkaTrigger struct {
 	// Name of the topic.
 	// More info at https://kafka.apache.org/documentation/#intro_topics
 	Topic string `json:"topic" protobuf:"bytes,2,opt,name=topic"`
+	// +optional
 	// Partition to write data to.
+	// By default the value is “0” to use random partition you need set “-1”  more info at "https://www.confluent.io/blog/apache-kafka-producer-improvements-sticky-partitioner/
```
whynowy, line 573: What about making it `*int32`?

sensors/triggers/kafka/kafka.go
```diff
@@ -194,14 +195,22 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
 		pk = trigger.URL
 	}
 
-	t.Producer.Input() <- &sarama.ProducerMessage{
+	if trigger.Partition == -1 {
```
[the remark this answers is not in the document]
Thomasr-02: No, because Kafka chooses an option based on the key hash (if it's the same for all messages, all messages are delivered to the same partition). We need generate random key. You can see more information about how Kafka choosen an partition [here](https://www.confluent.io/blog/apache-kafka-producer-improvements-sticky-partitioner/)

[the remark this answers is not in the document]
whynowy: I believe, in line 56 (`sarama.NewConfig()`), it use a `HashPartitioner` by default (see https://github.com/Shopify/sarama/blob/eaf1def9262590b03747d463eb4e5cb287945c4d/config.go#L519) , which means if the key is `nil`, it randomly pick a partition (see https://github.com/Shopify/sarama/blob/main/partitioner.go#L158-L168), so we don't need to generate a random key.

pkg/apis/sensor/v1alpha1/types.go
OneCricketeer: This blog is unrelated to the general concept of Kafka partitioning.

The sticky partitioner is specific to the Java implementation

OneCricketeer:
Which option should someone set if want ordered events? As in, events grouped by partition, and events are sent to a multi partitioned topic?

whynowy:
> Does ordering matter? That's why you'd typically be having one partition (partition 0) to begin with.
> 
> Which option should someone set if want ordered events across partitions? As in, events grouped by key? Hashing random ids won't work for that.
> 
> It should be documented that if absolute ordering is necessary, the topic should have one partition.

Ordering can be achieved by either specifying a partition, or a key (which also write to same partition).

whynowy:
> > Does ordering matter? That's why you'd typically be having one partition (partition 0) to begin with.
> > Which option should someone set if want ordered events across partitions? As in, events grouped by key? Hashing random ids won't work for that.
> > It should be documented that if absolute ordering is necessary, the topic should have one partition.
> 
> Ordering can be achieved by either specifying a partition, or a key (which also writes to same partition).

We should also make the `PartitioningKey` optional with pointer, if it's nil, set it nil in the producer message.

Thomasr-02:
@OneCricketeer @whynowy , sorry, maybe the blog mencioned caused a mess. 

This thread reffer to producer messages to all partition randomic. I tested as you mencioned if we set PartitioningKey null, works, the message chossen a randomic partition. This problem is solved.

But talk about the key, we can have a option to use random key, this is very important to many cases you can see [here ](https://stackoverflow.com/questions/29511521/is-key-required-as-part-of-sending-messages-to-kafka) more about benefits to use random key


New purpose: 
* When set `partition: -1` and empty in partitioningKey param  -> This generate a random key and messages is delivered to random partition. ( I add new validation to verify if partitioningKey is empty)

* To others scenarios behaviour still the same

whynowy:
> @OneCricketeer @whynowy , sorry, maybe the blog mencioned caused a mess.
> 
> This thread reffer to producer messages to all partition randomic. I tested as you mencioned if we set PartitioningKey null, works, the message chossen a randomic partition. This problem is solved.
> 
> But talk about the key, we can have a option to use random key, this is very important to many cases you can see [here ](https://stackoverflow.com/questions/29511521/is-key-required-as-part-of-sending-messages-to-kafka) more about benefits to use random key
> 
> New purpose:
> 
> * When set `partition: -1` and empty in partitioningKey param  -> This generate a random key and messages is delivered to random partition. ( I add new validation to verify if partitioningKey is empty)
> * To others scenarios behaviour still the same

To be honest, I don't understand what is the benefit of generating a random key, even after reading the link you posted, could you explain more about it?

Thomasr-02:
@whynowy 

Improved distribution: By using a random key, you can ensure that messages are distributed more evenly across the partition, rather than all ending up in the same partition. This can help to improve the overall performance of the Kafka cluster.

Better load balancing: When a message is produced with a specific key, it will always end up in the same partition. This can lead to hot spots in the cluster where certain partitions receive a much higher volume of messages than others. By using a random key, you can distribute the load more evenly across the cluster.

Better parallelism: When consuming messages from a Kafka topic, the consumer can specify the number of consumer threads to use. If the messages are evenly distributed across the partitions, each consumer thread will have an equal number of messages to process, leading to better parallelism and potentially improved performance.

Enhanced security: If you are using Kafka for secure communication, using a random key can help to improve the security of the system. By using a random key, it becomes more difficult for an attacker to predict which partition a message will end up in, making it more difficult to intercept messages.


This is a pattern here, I need to generate a random key😁😬

OneCricketeer:
> ensure that messages are distributed more evenly

This isn't true. Random UUID can very easily have hash collisions.

> By using a random key, you can distribute the load more evenly across the cluster.

You can achieve the same with a null key.

> Enhanced security

Kafka record keys aren't PII data, and have nothing to do with security. If you want to secure Kafka, you use TLS

OneCricketeer:
The main benefit I can see of having a UUID is traceability. E.g. Argo logs it generates some event with that ID, and you want to ensure your kafka producer does not drop, or consumer does not lose any events. This can also be used in systems like Jaeger. 

But you can stick the UUID in the Kafka record header as well. It doesn't need to be the partitioning key.

sensors/triggers/kafka/kafka.go
```diff
@@ -194,14 +199,20 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
...
-	t.Producer.Input() <- &sarama.ProducerMessage{
+	msg := &sarama.ProducerMessage{
 		Topic:     trigger.Topic,
-		Key:       sarama.StringEncoder(pk),
 		Value:     sarama.ByteEncoder(payload),
-		Partition: trigger.Partition,
 		Timestamp: time.Now().UTC(),
 	}
 
+	if trigger.Partition == -1 && trigger.PartitioningKey == "" {
```
OneCricketeer, line 208: In terms of readability, this condition could be moved closer to definition of `pk` since `trigger.PartitioningKey == ""` is already being checked.

Thomasr-02:
> > ensure that messages are distributed more evenly
> 
> This isn't true. Random UUID can very easily have hash collisions.

Using UUIDs as keys for messages in Kafka can have several benefits. Because UUIDs are unique across all devices and all time, there is no risk of collision, even if the same key is used for multiple messages. This can be especially useful in distributed systems, where messages may be produced by multiple devices and it is not possible to guarantee that the same key will not be used multiple times.

> 
> > By using a random key, you can distribute the load more evenly across the cluster.
> 
> You can achieve the same with a null key.
> 
What the benefits to use null as key ?

> > Enhanced security
> 
> Kafka record keys aren't PII data, and have nothing to do with security. If you want to secure Kafka, you use TLS

I'm not talking about the type of connection, but rather a scenario where it becomes more difficult for an attacker to predict which partition a message will end up in, making it difficult to intercept messages.

OneCricketeer:
> Because UUIDs are unique across all devices and all time, there is no risk of collision

Device doesn't matter. The collison I was referring to was the hashing algorithm used by Sarama. There's no guarantee ID1 and ID2 wouldn't end up in the same partition, despite `ID1 != ID2`

> benefits to use null as key

That's explained in the Stackoverflow post you linked to. It's truly round robin distribution, not key hash based.

> attacker... intercept

What does this even mean? Consumer groups are unique per Kafka client, so wouldn't be "intercepted". As stated, Kafka records aren't "secret". You would use certificates and ACLs to protect against "attackers" or rogue Kafka clients, and that's not within scope of this PR, at all.  

What scenario are you thinking of where being able to "predict" a partition is actually a problem for a real Kafka producer or consumer?

If I know any UUID, compute its hash then modulo the number of partitions in the topic, then *I do know exactly what partition that record will be in*, so that logic is flawed, anyway. 

If you send null keys, then you _definitely_ don't know what partition the value will exist in, *and it will be sent in round robin ordering*.

whynowy:
@Thomasr-02 - thanks for the detail, but I agree on @OneCricketeer's opinions.

My preference is to make `partitioningKey` a pointer, so that it can accept `nil`.

1.) If `partitioningKey == nil`, set it `nil` in the `ProducerMessage`, so that the message will be randomly distributed to a partition;
2.) if `partitioningKey != nil`, messages with same key will always be sent to same partition. 

For `partition`, I took a look at the sarama source code, it looks like it's not used to determine the real partition at all, which means only the `key` is used to make a decision. If that's the case, we can deprecate it from Argo Events. Correct me if I'm wrong.

Thomasr-02:
Ok @whynowy ,

I refactor adding your points.

whynowy (approved):
LGTM, thanks!

PR #2370, opened 2022-12-18
chore: sign container images and assests

34fathombelow:
Signed-off-by: Justin Marquis <34fathombelow@protonmail.com>

This PR implements container images and SBOM to be signed using ```sigstore/cosign```  It also consolidates the checksums of the cli-binaries  into a single file that will also be signed.

Also fixed SBOM and Manifest not being added to release assets.

### Three GitHub secrets will need to be created before this PR is merged. The process to do this is listed below.

TLDR: https://docs.sigstore.dev/cosign/git_support

1. [Install Cosign](https://docs.sigstore.dev/cosign/installation) on your workstation. Linux, Homebrew, and container options are available.  Execute the command ```cosign version``` to verified it has been installed correctly.
2. Create or have a valid GitHub PAT token available.
3. Export your PAT as the environment variable "GITHUB_TOKEN"
4. Execute ```cosign generate-key-pair github://argoproj/argo-events``` This will start the process of creating the GitHub Secrets automatically for you, and prompt you to enter a password.  This Password will be stored as the Github secret ```COSIGN_PASSWORD``` I would recommend to use well respected password generator such as KeePassXC or Bitwarden and use a paranoid level of characters of at least 32. 
5. Three GitHub secretes should have been created, please verify.  ```COSIGN_PASSWORD``` ```COSIGN_PRIVATE_KEY``` & ```COSIGN_PUBLIC_KEY```
6.  Delete these keys from your local workstation.

Checklist:


.github/workflows/release.yml
```diff
@@ -116,23 +116,69 @@ jobs:
...
+          username: ${{ secrets.QUAYIO_USERNAME }}
+          password: ${{ secrets.QUAYIO_PASSWORD }}
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@9becc617647dfa20ae7b1151972e9b3a2c338a2b # v2.8.1
+        with:
+          cosign-release: 'v1.13.1'
+
+      - name: Install crane to get digest of image
+        uses: imjasonh/setup-crane@e82f1b9a8007d399333baba4d75915558e9fb6a4
```
whynowy, line 154: What is the reason of not to use `imjasonh/setup-crane@v0.1` as mentioned in https://github.com/imjasonh/setup-crane?

whynowy:
@34fathombelow - thanks for doing all of these, I really learned something! For this PR, I will make the configuration change as you provided later - on vacation.

.github/workflows/release.yml
34fathombelow: I used the latest version of the action.  It appears the `docs` were not updated.  Time between versions was only three days, with only some minor changes.   Let me know if you prefer to use `v0.1`

Looks like I should comment the version.

34fathombelow:
@whynowy  It's always great to learn something new.  Hopefully in the near future I will be implementing keyless signing across all four Argo projects.  Docs on how to verify the signatures will soon follow in a separate PR.

Enjoy your vacation

.github/workflows/release.yml
whynowy: Actually there's a `v0.2` available, which is the latest.

https://github.com/imjasonh/setup-crane/tree/v0.2

34fathombelow: I just double checked.  I have pinned v0.2 with the hash.  Just sent a commit that comments the version being used.

whynowy (approved):
Thank you so much @34fathombelow !

Issue #2372, opened 2022-12-19
Github Sensor example has incorrect path header.X-Github-Event


rakhbari:
**Describe the bug**
The `X-Github-Event` path under `dependencies.test-dep.filters.data` in the Github Sensor example YAML file is incorrect and causes a `data filter error` when the Sensor consumes the Github event. 

**To Reproduce**
1. Follow the Github EventSource example setup
2. Start a log of your `Sensor` pod
3. Go to your configured repo and create a PR
5. See the `data filter error` stating: `(path 'header.X-Github-Event' does not exist)`

**Expected behavior**
The configured `Sensor` consumes the Github `pull_request` event and triggers the configured `Workflow` spec

**Screenshots**
N/A

**Environment (please complete the following information):**
 - Kubernetes: v1.25.4+k3s1
 - Argo: v3.4.4
 - Argo Events: v1.7.3

**Additional context**
The fix is quite simple. Change the word `header` to plural form `headers` (`path: headers.X-Github-Event`)

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2373, opened 2022-12-19
fix: Fixed Github Sensor example and minor doc correction

rakhbari:
Resolves issue #2372 

Also fixes the missing (404) link in the root `README.md` to the `CONTRIBUTING.md` doc that's under `docs/` dir.

Checklist:


whynowy (approved):
Thanks!

PR #2374, opened 2022-12-21
fix: fix bug in evaluation of filters with filtersLogicalOperator=or

jsvk:
Checklist:

This PR fixes a bug where when specifying only a subset of filters, in our case an expr filter and data filter, the filter would always evaluate to a 'pass'ing condition. 

This happens because the code doesn't check that we specified a filter before attempting to evaluate it, at which point it will default to `true`. 

The PR changes that behavior to check if the filter was `nil` before looking at what it evaluates to.


whynowy (approved):
LGTM, thanks for fixing it!

Issue #1503, opened 2022-01-08
Kafka event source - schemaRegistry/Avro support


tooptoop4:
This feature is about the kafka event source having the ability to consume avro messages and use schemaRegistry to deserialize it

Some inspirations:
https://github.com/nats-io/nats-kafka/blob/9e878e16ee0a378eb9d23cd41ee2df8e406ece40/server/kafka/consumer.go#L117,L126
https://github.com/stn1slv/kafka-console-avro-tools/blob/7e41775e44628b86dd12b750c134770a678d1908/main.go#L165-L167
https://github.com/confluentinc/confluent-kafka-go/issues/69#issuecomment-677822042
https://github.com/san-services/kafkaclient/commit/0d28abd2fdf654c1b2efd29fd3d05849bf99962e
https://github.com/vishalmodak/avro_poc/commit/d28498dbe09b788ed4069c6c5c5d6a287f9f83e5
https://github.com/bavix/go-kafka-comsume/blob/b07b423c6ca620d4d62777991bd84cd9cfe7caab/avro/marshaler.go#L11-L27
https://github.com/Dieken/kafkabat/blob/d60a254ebfe10977e0dcf686e0ae469c83c85463/main.go#L592
https://github.com/dangkaka/go-kafka-avro/blob/master/schemaRegistry.go#L91

Thomasr-02:
This issue already resolved! @whynowy  
https://github.com/argoproj/argo-events/releases/tag/v1.7.5

tooptoop4:
not completely, the id should not have to be defined in config

OneCricketeer:
> id should not have to be defined in config

The ID helps form a concrete event payload where fields can be filtered. 

If the suggestion is to always use the latest schema, then that can change in non-compatible ways. Handling that is more of a policy around the compatibility mode set on your usage of the registry.

tooptoop4:
don't want to hardcode ID, kafka connect does not require consumers to do that

OneCricketeer:
Kafka Connect supports schema evolution, it rarely maps fields into external, strict schemas (JDBC tables can evolve with alter commands, for example).

echzhai:
example how to use this schemaRegistry in eventsource yaml file?

OneCricketeer:
@echzhai See #2385

PR #2385, opened 2022-12-27
feat(sensor): Kafka Trigger - support avro/schemaRegistry

Thomasr-02:
Producer avro message Kafka Trigger

If is defined schemaRegistryConfig, the message are published on avro format 

```yml
    schemaRegistry:
      url: schema-registry-url
      username:
        key: user
        name: registry-user
      password:
         key: password
         name: password
      schemaId: 100
```

Maybe, at other moment it can be used to [EventSource Kafka support avro](https://github.com/argoproj/argo-events/issues/1503).

Checklist:


sensors/triggers/kafka/kafka.go
```diff
@@ -211,3 +230,51 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
 func (t *KafkaTrigger) ApplyPolicy(ctx context.Context, resource interface{}) error {
 	return nil
 }
+
+// getSchemaFromRegistry returns a schema from registry
+func getSchemaFromRegistry(sr *apicommon.SchemaRegistryConfig) (*srclient.Schema, error) {
+	user, err := common.GetSecretFromVolume(sr.UserSecret)
```
tooptoop4, line 236: user should be optional

sensors/triggers/kafka/kafka.go
```diff
@@ -211,3 +230,51 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting user value from secret, %w", err)
+	}
+	password, err := common.GetSecretFromVolume(sr.PasswordSecret)
```
tooptoop4, line 240: password should be optional

sensors/triggers/kafka/kafka.go
```diff
@@ -211,3 +230,51 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting password value from secret, %w", err)
+	}
+
+	schemaRegistryClient := srclient.CreateSchemaRegistryClient(sr.URL)
+	schemaRegistryClient.SetCredentials(user, password)
```
tooptoop4, line 246: don't set credentials if no user/pass supplied

tooptoop4 (changes requested):

pkg/apis/common/common.go
```diff
@@ -156,6 +156,18 @@ type SASLConfig struct {
 	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty" protobuf:"bytes,3,opt,name=password"`
 }
 
+// SchemaRegistryConfig refers to configuration for a client
+type SchemaRegistryConfig struct {
+	// Schema Registry URL.
+	URL string `json:"url" protobuf:"bytes,1,opt,name=url"`
+	// Api key is the authentication for registry
+	UserSecret *corev1.SecretKeySelector `json:"userSecret,omitempty" protobuf:"bytes,2,opt,name=user"`
```
tooptoop4, line 164: // +optional

pkg/apis/common/common.go
```diff
@@ -156,6 +156,18 @@ type SASLConfig struct {
...
+	// Api secret is the authentication for registry
+	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty" protobuf:"bytes,3,opt,name=password"`
```
tooptoop4, line 166: // +optional

sensors/triggers/kafka/kafka.go
```diff
@@ -211,3 +230,51 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
...
+
+	schema, err := schemaRegistryClient.GetSchema(int(sr.SchemaId))
```
tooptoop4, line 248: rather than have id as a value in config, which is likely to be different per environment, instead could the id be found on the message itself? like https://github.com/nats-io/nats-kafka/blob/9e878e16ee0a378eb9d23cd41ee2df8e406ece40/server/kafka/consumer.go#L320-L330

Thomasr-02: I do not agree with this point.

This means: All payloads/messages must have schemaID.
If we have a fixed wehbook source like Gitlab where we can't edit the payload/messages then all messages need to be edited to add schemaId.

tooptoop4: @Thomasr-02 if schemaId not supplied in the config then read schemaid off the message

Thomasr-02: I understood. 

I believe this can be an improvement, it doesn't fit the scope of the solution I proposed

tooptoop4 (approved):

tooptoop4:
@whynowy pls merge

pkg/apis/common/common.go
```diff
@@ -156,6 +156,20 @@ type SASLConfig struct {
...
+	// Schema Registry URL.
+	URL string `json:"url" protobuf:"bytes,1,opt,name=url"`
+	// +optional
+	// Api key is the authentication for registry
+	UserSecret *corev1.SecretKeySelector `json:"userSecret,omitempty" protobuf:"bytes,2,opt,name=user"`
+	// +optional
+	// Api secret is the authentication for registry
+	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty" protobuf:"bytes,3,opt,name=password"`
+	// Schema ID
+	SchemaId int32 `json:"schemaId,omitempty" protobuf:"varint,4,opt,name=schemaId"`
```
whynowy, line 170: Nit: Use `SchemaID` in golang (keep `schemaId` in the `json`).

Is this field required when the registry config is in-use?

pkg/apis/common/common.go
```diff
@@ -156,6 +156,20 @@ type SASLConfig struct {
 	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty" protobuf:"bytes,3,opt,name=password"`
 }
 
+// SchemaRegistryConfig refers to configuration for a client
+type SchemaRegistryConfig struct {
+	// Schema Registry URL.
+	URL string `json:"url" protobuf:"bytes,1,opt,name=url"`
+	// +optional
+	// Api key is the authentication for registry
+	UserSecret *corev1.SecretKeySelector `json:"userSecret,omitempty" protobuf:"bytes,2,opt,name=user"`
```
whynowy, line 165: Can we keep the property names same in `json` and `protobuf`?

BTW, these can be replaced by using something like:

```golang
	// xxxxxx
	// +optional
	apicommon.BasicAuth `json:",inline" protobuf:"bytes,2,opt,name=basicAuth"`
```

pkg/apis/common/common.go
```diff
@@ -156,6 +156,20 @@ type SASLConfig struct {
...
+	// +optional
+	// Api secret is the authentication for registry
+	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty" protobuf:"bytes,3,opt,name=password"`
```
whynowy, line 168: Same here.

pkg/apis/common/common.go
Thomasr-02: Ok, sorry.
Yes schemaID is mandatory when config is defined.

pkg/apis/common/common.go
Thomasr-02: Okay! I'm using BasicAuth now.

sensors/triggers/kafka/kafka.go
```diff
@@ -210,3 +227,45 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
 func (t *KafkaTrigger) ApplyPolicy(ctx context.Context, resource interface{}) error {
 	return nil
 }
+
+// getSchemaFromRegistry returns a schema from registry
+func getSchemaFromRegistry(sr *apicommon.SchemaRegistryConfig) (*srclient.Schema, error) {
+	schemaRegistryClient := srclient.CreateSchemaRegistryClient(sr.URL)
+	user, _ := common.GetSecretFromVolume(sr.Auth.Username)
```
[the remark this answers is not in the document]
Thomasr-02: okay

sensors/triggers/kafka/kafka.go
```diff
@@ -189,6 +193,19 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
 		return nil, err
 	}
 
+	// Producer with avro schema
+	if trigger.SchemaRegistry != nil {
+		schema, err := getSchemaFromRegistry(trigger.SchemaRegistry)
```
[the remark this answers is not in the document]
Thomasr-02: First: It is not frequently to change the scheme
Yes I agree. I think of 2 possible solutions:
- Cache (We need think about the key and ttl)
- Use schema raw (.avsc) file as file, this means we no longer need schema registry setting to get schema.

What do you think @whynowy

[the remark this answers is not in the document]
tooptoop4: cache is better,  if a schema changes it would generate new id anyway.  since schemaid is hardcoded in the config, I think can read the schema into a variable at start of the process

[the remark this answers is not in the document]
whynowy: Maybe we can put the `schema` into `KafkaTrigger` as a property, and init it in the New function. 

Today the trigger instance is newed per message, we can change it later to be cached.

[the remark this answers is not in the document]
Thomasr-02: @whynowy @tooptoop4 , 

Using schema as a property maybe make sensor file much bigger,  usually the schema is bigger.  
Ex:  .avsc file has (64rows)
My sensor has (35rows)

I thinked to use a base64 of schema but I don't know if is a good ideia to reduce file size

My purpose is to use the raw file because it is more easy to control version and commits, and more easy to define and handling it.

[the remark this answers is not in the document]
tooptoop4: @whynowy would moving this code to eventsource mean it can cache response from schema registry?

as defining the avsc content means schema registry is not supported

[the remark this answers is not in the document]
whynowy: @Thomasr-02 - What I meant was not moving it to the sensor spec, but getting the schema with remote call in `NewKafkaTrigger()` function, and put the result as a private property in `KafkaTrigger` struct, in this way, for each instance of `KafkaTrigger`, there's only 1 time remote call to get the schema.

Today, every time when there's a need to trigger an action, we new a trigger instance (here, it's KafkaTrigger instance), see https://github.com/argoproj/argo-events/blob/master/sensors/trigger.go#L55, which means, with the approach above, we still do lots of remote call to get schemas, but this can be changed in a separate PR to cache the trigger instance.

@tooptoop4 - sensor and enventsource are totally decoupled, we should not move it to eventsource. Not quite clear about what you said about `as defining the avsc content means schema registry is not supported`.

Thomasr-02:
@whynowy i refactor now, review again, pls.

How works release, you can generate a version after this PR, I need use it.

sensors/trigger.go
```diff
@@ -99,7 +102,16 @@ func (sensorCtx *SensorContext) GetTrigger(ctx context.Context, trigger *v1alpha
 	}
 
 	if trigger.Template.Kafka != nil {
-		result, err := kafka.NewKafkaTrigger(sensorCtx.sensor, trigger, sensorCtx.kafkaProducers, log)
+		var schema *srclient.Schema
+		if trigger.Template.Kafka.SchemaRegistry != nil {
```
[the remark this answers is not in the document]
Thomasr-02: Moved!!

sensors/triggers/kafka/kafka.go
```diff
@@ -42,6 +46,8 @@ type KafkaTrigger struct {
 	Producer sarama.AsyncProducer
 	// Logger to log stuff
 	Logger *zap.SugaredLogger
+	// Avro schema of message
+	Schema *srclient.Schema
```
whynowy, line 50: Make it private (lowercase)

sensors/triggers/registry.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"fmt"
+
+	"github.com/argoproj/argo-events/common"
+	apicommon "github.com/argoproj/argo-events/pkg/apis/common"
+	"github.com/riferrei/srclient"
+)
+
+// GetSchemaFromRegistry returns a schema from registry.
+func GetSchemaFromRegistry(sr *apicommon.SchemaRegistryConfig) (*srclient.Schema, error) {
```
whynowy, line 12: We don't need this file, the function is particularly for kafka trigger, putting it in the kafka trigger is better.

whynowy (commented):
It looks good to me other than those 2 comments.

Thomasr-02:
@whynowy how works release in argo-events?! You can generate a release after merged it! I need use this feature.

whynowy (approved):
LGTM, thanks

Thomasr-02:
@whynowy , you can generate a relase, pls!?

Thomasr-02:
@whynowy you know when this feat will be published ?

whynowy:
> @whynowy you know when this feat will be published ?

Will be release in the coming couple of days.

Issue #1977, opened 2022-05-16
Asynchronous triggering of actions makes it possible to ack a message prior to performing the action


juliev0:
We try to be "exactly once" in most cases and "at least once" in the case of SIGFAULT. 

To be "exactly once" in the case of normal shutdown (SIGTERM), the logic prevents shutting down in the middle of processing a message. But the fact that actions are triggered [asynchronously](https://github.com/argoproj/argo-events/blob/master/sensors/listener.go#L335) means that we don't wait until the triggers have occurred to shut down.

Also, to be "at least once" in the case of the Sensor getting a SIGFAULT, we specifically ack the EventBus (NATS streaming as well as Jetstream) at the end of processing the message. But because the action gets triggered asynchronously, the ack can come before the action, so if the Sensor got a SIGFAULT after performing the ack and prior to triggering the action then the message would be lost.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
keeping alive in the hopes I'll work on it at some point

juliev0:
@whynowy Could we add this to our project? I talked to Qianbo and he says this would be a good thing to tackle when we get a chance.

dfarr:
@juliev0 we can take a look into supporting this

bilalba:
In order to make it at least once, we looked into implementing a callback that would ack at the end of asynchronous trigger call completing. 

First we looked into just implementing [the ack ](https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/sensor/trigger_conn.go#L260-L268) as a callback to an asynchronous trigger completing. We would pass this callback function as an additional argument [here](https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/sensor/trigger_conn.go#L330) and forward the call back to [triggerActions()](https://github.com/argoproj/argo-events/blob/master/sensors/listener.go#L325), and [triggerwithRateLimit()](https://github.com/argoproj/argo-events/blob/master/sensors/listener.go#L325), and finally call this callback in triggerWithRateLimit. 

However, two things may happen from this. 

1. [clearAllDependencies()](https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/sensor/trigger_conn.go#L450) will get invoked even though the message has not been acked. If clearAllDependencies gets invoked, and the system shuts down, we lose out on previous messages.

[...]

PR #2402, opened 2023-01-09
Implement optional at least once semantics

bilalba:
Checklist:

When a trigger is invoked, optionally wait until the action is complete before acknowledging the message in order to execute the action at least once. Currently, Argo Events immediately acks(as noted in #1977), resulting in at most once semantics. The behavior can be specified on the trigger, defaults to at most once.

Instead of immediately acknowledging the message, mark the message [InProgress](https://docs.nats.io/using-nats/developer/develop_jetstream#acknowledging-messages) to prevent an AckTimeout and ack it once it has completed.


eventbus/jetstream/sensor/trigger_conn.go
```diff
@@ -257,15 +257,35 @@ func (conn *JetstreamTriggerConn) processMsg(
...
 	}
 
-	defer func() {
-		err = m.AckSync() // this call does the acknowledgment and waits for an acknowledgment from the server (that it received the ack) too
-		if err != nil {
-			errStr := fmt.Sprintf("Error performing AckSync() on message: %v", err)
-			conn.Logger.Error(errStr)
+	done := make(chan bool)
+	go func() {
+		ticker := time.NewTicker(500 * time.Millisecond)
```
tooptoop4, line 262: 500ms is too long

dfarr: It looks like we don't set an `AckWait` when we create the [pull subscription](https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/sensor/trigger_conn.go#L136). Do you know what the default is? I can't find it in the docs [here](https://pkg.go.dev/github.com/nats-io/nats.go#AckWait)

PR #2404, opened 2023-01-11
Implement optional at least once semantics

bilalba:
Checklist:

When a trigger is invoked, optionally wait until the action is complete before acknowledging the message in order to execute the action at least once. Currently, Argo Events immediately acks(as noted in #1977), resulting in at most once semantics. The behavior can be specified on the trigger, defaults to at most once.

Instead of immediately acknowledging the message, mark the message [InProgress](https://docs.nats.io/using-nats/developer/develop_jetstream#acknowledging-messages) to prevent an AckTimeout and ack it once it has completed.


eventbus/jetstream/sensor/trigger_conn.go
```diff
@@ -257,15 +257,35 @@ func (conn *JetstreamTriggerConn) processMsg(
...
+		for {
+			select {
+			case <-done:
+				err = m.AckSync()
+				if err != nil {
+					errStr := fmt.Sprintf("Error performing AckSync() on message: %v", err)
+					conn.Logger.Error(errStr)
+				}
+				conn.Logger.Debugf("acked message of Stream seq: %s:%d, Consumer seq: %s:%d", meta.Stream, meta.Sequence.Stream, meta.Consumer, meta.Sequence.Consumer)
+				ticker.Stop()
```
whynowy, line 272: Use `defer ticker.Stop()` after line 262.

eventbus/jetstream/sensor/trigger_conn.go
```diff
@@ -257,15 +257,35 @@ func (conn *JetstreamTriggerConn) processMsg(
...
 	}
 
-	defer func() {
-		err = m.AckSync() // this call does the acknowledgment and waits for an acknowledgment from the server (that it received the ack) too
-		if err != nil {
-			errStr := fmt.Sprintf("Error performing AckSync() on message: %v", err)
-			conn.Logger.Error(errStr)
+	done := make(chan bool)
+	go func() {
+		ticker := time.NewTicker(500 * time.Millisecond)
```
whynowy, line 262: @juliev0 - how is the jetstream consumer created?

eventbus/jetstream/sensor/trigger_conn.go
```diff
@@ -257,15 +257,35 @@ func (conn *JetstreamTriggerConn) processMsg(
...
+			select {
+			case <-done:
+				err = m.AckSync()
+				if err != nil {
+					errStr := fmt.Sprintf("Error performing AckSync() on message: %v", err)
+					conn.Logger.Error(errStr)
+				}
+				conn.Logger.Debugf("acked message of Stream seq: %s:%d, Consumer seq: %s:%d", meta.Stream, meta.Sequence.Stream, meta.Consumer, meta.Sequence.Consumer)
+				ticker.Stop()
+				break
```
whynowy, line 273: return

eventbus/jetstream/sensor/trigger_conn.go
```diff
@@ -257,15 +257,35 @@ func (conn *JetstreamTriggerConn) processMsg(
...
+					errStr := fmt.Sprintf("Error performing InProgess() on message: %v", err)
+					conn.Logger.Error(errStr)
+				}
+				conn.Logger.Debugf("InProgess message of Stream seq: %s:%d, Consumer seq: %s:%d", meta.Stream, meta.Sequence.Stream, meta.Consumer, meta.Sequence.Consumer)
+			}
 		}
+	}()
 
-		conn.Logger.Debugf("acked message of Stream seq: %s:%d, Consumer seq: %s:%d", meta.Stream, meta.Sequence.Stream, meta.Consumer, meta.Sequence.Consumer)
+	defer func() {
```
whynowy, line 285: This means, no matter the triggering is successful not not, the message will be acked anyways, this is not `atLeastOnce`.

sensors/listener.go
```diff
@@ -331,7 +331,11 @@ func (sensorCtx *SensorContext) triggerActions(ctx context.Context, sensor *v1al
 		depNames = append(depNames, k)
 		eventIDs = append(eventIDs, v.ID())
 	}
-	go sensorCtx.triggerWithRateLimit(ctx, sensor, trigger, eventsMapping, depNames, eventIDs)
+	if trigger.AtLeastOnce {
+		sensorCtx.triggerWithRateLimit(ctx, sensor, trigger, eventsMapping, depNames, eventIDs)
```
whynowy, line 335: My understanding this is block processing, how is it related to `atLeastOnce`?

eventbus/jetstream/sensor/trigger_conn.go
dfarr: When the retry loop is exhausted at some point we have no other choice than to ack the message and move on. This has no bearing on at least once vs at most once semantics.

At most once: attempt invocation only once, if a response is not received (eg: service unavailable) do not attempt again, as we cannot say if the downstream system will eventually succeed.

At least once: attempt invocation multiple times, this may result in the action being invoked multiple times. However, we still need to give up and move on at some point.

sensors/listener.go
dfarr: A function higher up in the call chain acks the message in a defer statement. If `sensorCtx.triggerWithRateLimit` is called in go routine, then we ack the message without waiting to see if the trigger invocation was successful or unsuccessful, this is equivalent of at most once. If `sensorCtx.triggerWithRateLimit` is called synchronously (aka not in a go routine) then we wait to ack the message until know the trigger has been invoked (or we have exhausted the retry loop).

dfarr: @bilalba maybe we should add an explanatory comment here?

eventbus/jetstream/sensor/trigger_conn.go
juliev0: See line 132 of this file where getDurableName() is called based on sensor name, trigger name, and dependency name and then passed on line 136 to the PullSubscribe call.

juliev0: (Not sure if that's what you were asking...I guess my answer is really related to the 'consumer group'

eventbus/jetstream/sensor/trigger_conn.go
bilalba: Like discussed in #1977, we've looked to change the behavior in the case of a SIGFAULT, that's still something we can control to make it from AtMostOnce to AtLeastOnce.

juliev0 (approved):

eventbus/jetstream/sensor/trigger_conn.go
whynowy: Trying to understand what you proposed.

Let's assume there are 1M stacked messages, with async triggering, it's possible that lots of triggering won't be able to get a chance to execute because the goroutine resources might have exhausted, but the messages have been acked, so there's no chance to retrigger. But with sync triggering, this will not happen. Am I right?

bilalba: Yes. The messages will start processing but won’t get acked until the trigger processing completes. 
And we show in the functional test that if a sensor goes down while the message is being processed, it will get processed again.

dfarr: Just to expand a bit on this. When a trigger is satisfied and needs to be invoked we can either:
1. Call the action function _before_ we acknowledge the message (current implementation, albeit in a roundabout way because of the go routine)
2. Call the action function _after_ we acknowledge the message

In both scenarios, our application may crash in between message acknowledgement and calling the action function. But in the second scenario our application will re-process the message on restart. In total there are four possibilities:

| ack | crash | invocations |
| --- | --- | --- |
| before | no | 1 |
| before | yes | 0 |
| after | no | 1 |
| after | yes | 1+ |

The first two constitute at most once and the second two constitute at least once. With that being said, once the retry policy has been exhausted we still need to ack the message and move on even in the case of at least once.

whynowy (approved):
LGTM

dfarr:
> @bilalba question. Suppose I have argo event sensing a kafka topic and that the trigger action consists for example in calling an argo workflow and I set the at_least_once flag to true. Suppose the message is delivered and the argo workflow goes down before completing the work, is that message going to be retried? No, right? since argo events doesn't control what happens within a trigger, right?

No, it will not be retried in this case. In the context of an argo workflow at least once is with respect to the _creation_ of the argo workflow. Your argo workflow is guaranteed to be created at least once, whether or not that workflow succeeds has no affect on your sensor/trigger.

Issue #2403, opened 2023-01-10
Remove hard requirement of cluster-wide secret list/watch privileges


jessesuen:
Argo Events, when installed in a cluster-scoped mode, currently require cluster-wide `list`/`watch`/`get` on Secrets. The controller will simply not start if the RBAC is removed and the controller can't start the watch. This issue is to remove this hard requirement for cluster-wide secret `list`/`watch`/`get`.

**Describe the solution you'd like**

The reason argo-events watches secrets, is because it re-queues the parent EventBus object for whenever its child Secret changes. But since the child secret is controlled by the controller (not by admins), this is largely unnecessary. Also, it will eventually be recreated/updated during normal resync periods of EventBus objects.

I think Argo Events can improve its security by removing this watch requirement, and allowing administrators to allow `list/get` Secret calls at a namespace granularity. 

This is how both Argo Workflows and Argo Rollouts are implemented. See:
* https://github.com/argoproj/argo-workflows/issues/8534
* https://github.com/argoproj/argo-workflows/issues/9572
* https://github.com/argoproj/argo-workflows/pull/8555
* https://github.com/argoproj/argo-rollouts/pull/829

**Describe alternatives you've considered**

Tried patching the RBAC manually but the way the code is implemented is impossible. Currently, the only way to run argo-events in a locked-down cluster is to run multiple instances of it. But this is an expensive and heavy-handed workaround which also does not allow config reuse.

**Additional context**

Chatted with @whynowy about this and he agrees this is something that can be accomplished.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2410, opened 2023-01-14
feat: allow granular secret privileges. additional controller logging/leader-election options

jessesuen:
Resolves https://github.com/argoproj/argo-events/issues/2403

## 1. Removes requirement of having cluster-scoped secret privileges.

Currently, argo-events requires secrets list/watch/get privileges at a cluster scope. However, the only reason it wants this is to requeue the eventbus to the eventbus controller in case the NATS auth credentials secrets becomes modified or deleted. The watch is largely unnecessary because:
  * the secret will not naturally become modified or deleted. It is owned by the eventbus resource and not something a user manages.
  * even if it does, it will eventually be reconciled by virtue of the eventbus sync interval

This change improves the security of argo-events by allowing Argo Events to run with only list/get secrets granted at a per-namespace basis (by creating a namespaced RoleBinding to map a ClusterRole to the argo-events:argo-events-sa ServiceAccount)

In order to achieve this, the following changes were made:
  * do not start the secret watch during startup. As I mentioned earlier, I feel this is unnecessary.
  * switch to a normal kubernetes client instead of the controller-runtime client. The reason this is necessary, is because the controller-runtime client still attempts to perform cluster-wide API list calls (even when a namspace is supplied to the `List()` call). 

## 2. Improved controller and developer convenience options

During development, I found it quite inconvenient to develop so I added the following:
  * option to start the controller without leader election (so we can set breakpoints and pause indefinitely)
  * option to increase the *kubernetes* log level (to see what API calls were being made)

Signed-off-by: Jesse Suen <jesse@akuity.io>

### DIFF
--- a/cmd/commands/controller.go
+++ b/cmd/commands/controller.go
@@ -3,24 +3,42 @@ package commands
 import (
 	"github.com/spf13/cobra"
 
+	"github.com/argoproj/argo-events/common"
+	"github.com/argoproj/argo-events/common/logging"
 	controllercmd "github.com/argoproj/argo-events/controllers/cmd"
 	envpkg "github.com/argoproj/pkg/env"
 )
 
 func NewControllerCommand() *cobra.Command {
 	var (
+		leaderElection   bool
 		namespaced       bool
 		managedNamespace string
+		metricsPort      int32
+		healthPort       int32
+		klogLevel        int
 	)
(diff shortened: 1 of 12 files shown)


jessesuen:
Oops. Meant to submit this from my fork.

PR #2438, opened 2023-01-23
feat: Optional kubernetes-based leader election

dfarr:
Implements optional kubernetes-based leader election. When an `EventSource` or `Sensor` opts in to kubernetes-based leader election, the Kubernetes lease resource will be used to coordinate pods. This change decouples the leader election mechanism from the `EventBus`. The default leaves an EventBus-based leader election mechanism in place.

The following annotation can be specified on either `EventSource` or `Sensor` to opt in to kubernetes-based leader election.
```yaml
annotations:
  events.argoproj.io/leader-election: k8s
```

Checklist:

### DIFF
--- a/common/leaderelection/leaderelection.go
+++ b/common/leaderelection/leaderelection.go
@@ -4,17 +4,22 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"time"
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/nats-io/graft"
 	nats "github.com/nats-io/nats.go"
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 
 	"github.com/argoproj/argo-events/common"
 	"github.com/argoproj/argo-events/common/logging"
 	eventbuscommon "github.com/argoproj/argo-events/eventbus/common"
-	apicommon "github.com/argoproj/argo-events/pkg/apis/common"
 	eventbusv1alpha1 "github.com/argoproj/argo-events/pkg/apis/eventbus/v1alpha1"
 )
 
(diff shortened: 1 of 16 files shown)


dfarr:
I am unsure if an annotation is the best approach for user opt-in. An alternative would be to add a section to both the `EventSource` and `Sensor` schema. This section would be optional, it is omitted, the current EventBus-based leader election would be used.

```yaml
leaderElection:
  leaseDuration: 15
  renewDeadline: 10
  retryPeriod: 2
```

The settings correspond to options available on the [LeaderElectionConfig config](https://pkg.go.dev/k8s.io/client-go@v0.26.1/tools/leaderelection#LeaderElectionConfig)

whynowy:
> I am unsure if an annotation is the best approach for user opt-in. An alternative would be to add a section to both the `EventSource` and `Sensor` schema. This section would be optional, if omitted the current EventBus-based leader election would be used.
> 
> ```yaml
> leaderElection:
>   leaseDuration: 15
>   renewDeadline: 10
>   retryPeriod: 2
> ```
> 
> The settings correspond to options available on the [LeaderElectionConfig config](https://pkg.go.dev/k8s.io/client-go@v0.26.1/tools/leaderelection#LeaderElectionConfig)

How about using annotations, but automatically convert to ENV variables?

sensors/listener.go
```diff
@@ -56,11 +56,15 @@ func subscribeOnce(subLock *uint32, subscribe func()) {
 func (sensorCtx *SensorContext) Start(ctx context.Context) error {
 	log := logging.FromContext(ctx)
 	custerName := fmt.Sprintf("%s-sensor-%s", sensorCtx.sensor.Namespace, sensorCtx.sensor.Name)
-	elector, err := leaderelection.NewEventBusElector(ctx, *sensorCtx.eventBusConfig, custerName, int(sensorCtx.sensor.Spec.GetReplicas()))
+	replicas := int(sensorCtx.sensor.Spec.GetReplicas())
+	leasename := fmt.Sprintf("sensor-%s", sensorCtx.sensor.Name)
+
+	elector, err := leaderelection.NewElector(ctx, *sensorCtx.eventBusConfig, custerName, replicas, sensorCtx.sensor.Namespace, leasename, sensorCtx.hostname)
```
juliev0, line 62: I know this was already in there but maybe fix spelling of "custerName"?

juliev0: (like you did in the other spot)

dfarr: Sounds good, will do 👍

eventsources/cmd/start.go
```diff
@@ -62,6 +64,15 @@ func Start() {
 
 	logger.Infow("starting eventsource server", "version", argoevents.GetVersion())
 	adaptor := eventsources.NewEventSourceAdaptor(eventSource, busConfig, ebSubject, hostname, m)
+
+	// add annotations to context
+	for key, value := range eventSource.Annotations {
```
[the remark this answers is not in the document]
dfarr: Sounds good! Let me give this a try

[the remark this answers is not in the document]
dfarr: I was able to pass the information by reading the annotation in the controller, no need to use the downwards API. And it worked! https://github.com/argoproj/argo-events/pull/2438/files#diff-1a0be55c7a0c0283ffb810e0a1b4c38b2c1d0b7513793aea5c1f265ee5dd0291R196

common/leaderelection/leaderelection.go
```diff
@@ -179,3 +188,67 @@ func (e *natsEventBusElector) RunOrDie(ctx context.Context, callbacks LeaderCall
...
+			Name:      e.leasename,
+			Namespace: e.namespace,
+		},
+		Client: client.CoordinationV1(),
+		LockConfig: resourcelock.ResourceLockConfig{
+			Identity: e.hostname,
+		},
+	}
+
+	for {
```
whynowy, line 230: There's no need to to have a `for/select` here, it's not working as expected - when it executing `default`, where `RnOrDie` is blocking, even though there's SIGTERM, the `case <-ctx.Done()` won't be executed.

Simply removing the `for/select` should work.

common/leaderelection/leaderelection.go
```diff
@@ -179,3 +188,67 @@ func (e *natsEventBusElector) RunOrDie(ctx context.Context, callbacks LeaderCall
...
+				Callbacks: leaderelection.LeaderCallbacks{
+					OnStartedLeading: callbacks.OnStartedLeading,
+					OnStoppedLeading: callbacks.OnStoppedLeading,
+				},
+			})
+
+			// When the leader is lost, leaderelection.RunOrDie will
+			// cease blocking and we will cancel the context. This
+			// will halt all eventsource/sensor go routines.
+			cancel()
```
whynowy, line 251: Use `defer`?

common/leaderelection/leaderelection.go
```diff
@@ -179,3 +188,67 @@ func (e *natsEventBusElector) RunOrDie(ctx context.Context, callbacks LeaderCall
...
+	}
+
+	lock := &resourcelock.LeaseLock{
+		LeaseMeta: metav1.ObjectMeta{
+			Name:      e.leasename,
+			Namespace: e.namespace,
+		},
+		Client: client.CoordinationV1(),
+		LockConfig: resourcelock.ResourceLockConfig{
+			Identity: e.hostname,
```
whynowy, line 226: How to opt in following configuration?

```yaml
  leaseDuration: 15
  renewDeadline: 10
  retryPeriod: 2
```

test/e2e/testdata/es-calendar-ha-k8s.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: argoproj.io/v1alpha1
```
whynowy, line 1: Add some docs?

common/leaderelection/leaderelection.go
dfarr: That was my initial inclination too, but it turns out (oddly imo) that `leaderelection.RunOrDie` will stop blocking if a pod is the leader and then stops being the leader. The for loop ensures the pod puts itself back in the leader election pool and the `<-ctx.Done()` case ensures that the function will exit gracefully if there's a `SIGTERM`.

common/leaderelection/leaderelection.go
dfarr: What are your thoughts on adding this to Sensor/EventSource, if omitted we can default to the EventBus-based leader election
```
leaderElection:
  leaseDuration: int
  renewDeadline: int
  retryPeriod: int
```

common/leaderelection/leaderelection.go
dfarr: When a sigterm occurs the parent context is cancelled, which causes all child child contexts to be cancelled, including the one passed to `leaderelection.RunOrDie`. This causes the leader election to halt and for the first branch of the select statement to be invoked on the next iteration of the for loop. I can confirm this behaviour when running locally.

common/leaderelection/leaderelection.go
whynowy: How about keeping it as is, and revisit it later if it's highly demanded for customization?

dfarr: Sounds like a good plan to me!

docs/eventsources/ha.md
```diff
@@ -53,6 +53,16 @@ old one is gone.
...
 
+## Kubernetes Leader Election
+
+By default, Argo Events will use NATS for the HA leader election. Alternatively,
+you can opt-in to a Kubernetes native leader election (that uses a Lease) by
+specifying the following annotation.
+```yaml
+annotations:
+  events.argoproj.io/leader-election: k8s
+```
```
whynowy, line 64: Thanks! Also need to mention the service account used for the EventSource need to have extra RBAC config, and give the example `Role` and `RoleBinding`.

dfarr: Yes! Thank you, added :+1:

whynowy (approved):
LGTM, thanks!

PR #2466, opened 2023-02-10
fix(controller): use a copied object to update because of mutation

whynowy:

Controller-runtime client.Update() will mutate the object, which makes the following up Status.Update() do nothing - it didn't cause any issue because there are couple of reconciliations, eventually the Status gets updated correctly. 

Thanks @dfarr @bilalba for reporting the issue. 

Checklist:


dfarr (approved):

dfarr:
Looks good! Very clean solution

sarabala1979 (approved):

PR #2474, opened 2023-02-14
feat(eventsource): Gitlab eventsource - support groups (#2386)

### LINKED ISSUE #2386: gitlab eventsource support for groups
My company has a GitLab with a lot of groups and subgroups with thousands of projects, so if I want to create an event source to listen to all my company webhooks or a particular group is a huge pain to write all in an EventsSource object, and maintain this updated is another pain, because everyday new projects die and born.

E.g:

```yaml
kind: EventSource
metadata:
  annotations:
  name: gitlab-webhook
  namespace: argo-events
spec:
  gitlab:
    my-company:
      events:
      - JobEvents
      gitlabBaseURL: https://gitlab.my-company.com/api/v4
      projects:
      - my-company/group1/subgroup1/projectA
      - my-company/group1/subgroup1/projectB
      - my-company/group1/subgroup1/projectC
      - my-company/group2/subgroup1/projectA
      - my-company/group2/subgroup72/projectZ
      - ....
```
I have about 8k projects now :cry: 

It would be nice if the GitlabEventSource had a new field, `group` that accepts a group path and create the webhook at the group level instead of a specific project. This allows us to receive events from a specific group or all the groups of the company.

Other solutions that I can take is create a EventsSource to a unique project and use their listener URL to create manually a webook in some GitLab groups. This works I have tested but its is ugly and depend of manual actions for each Group.

---

neo502721:
Fix #2386. Support groups level hook to gitlab eventsource.  This is available in Premium and Ultimate Gitlab.

Checklist:


pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -863,6 +863,10 @@ type GitlabEventSource struct {
 	// Filter
 	// +optional
 	Filter *EventSourceFilter `json:"filter,omitempty" protobuf:"bytes,12,opt,name=filter"`
+	// List of group IDs or group name like "test".
+	// Group level hook available in Premium and Ultimate Gitlab.
+	// +optional
+	Groups []string `json:"groups,omitempty" protobuf:"bytes,13,rep,name=groups"`
```
whynowy, line 869: Not familiar with Gitlab, is the field `Projects` still required when `Groups` field is present? What is the relationship between them?

eventsources/sources/gitlab/start.go
```diff
@@ -216,26 +241,37 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 			return fmt.Errorf("failed to initialize client, %w", err)
 		}
 
-		getHook := func(hooks []*gitlab.ProjectHook, url string) *gitlab.ProjectHook {
-			for _, h := range hooks {
-				if h.URL != url {
+		router.hookIDs = make(map[string]int)
```
whynowy, line 244: Is the hook id unique, no matter if it's a project hook, or a group hook?

pkg/apis/eventsource/v1alpha1/types.go
neo502721: Group hooks and project hooks could exist at the same time. Group Hooks are different from project Hooks that are limited to one project.  [https://docs.gitlab.com/ee/api/groups.html#hooks](url)
From my perspective, both of them (Group & Project) should be optional. However, at least one of them should be set.
Only  `premium` and `ultimate ` version Gitlab supports Group level hook. Since all projects under the group could trigger group hook. I avoid to make it too many api calls (which is truly ugly) by using the premium feature of Gitlab.

eventsources/sources/gitlab/validate.go
```diff
@@ -31,7 +31,7 @@ func validate(eventSource *v1alpha1.GitlabEventSource) error {
 	if eventSource == nil {
 		return common.ErrNilEventSource
 	}
-	if len(eventSource.GetProjects()) == 0 {
+	if len(eventSource.GetProjects()) == 0 && len(eventSource.GetGroups()) == 0 {
 		return fmt.Errorf("projects can't be empty")
```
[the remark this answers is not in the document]
neo502721: OK, I will update it later.

eventsources/sources/gitlab/start.go
neo502721: Don't have to consider whether the hook id is unique between project hook and group hook, Because they are different Gitlab API and `router.hookIDs` key.

whynowy: My question is, if it's not unique across groups and projects, will using one `router.hookIDs` be a problem?

neo502721: I see what you mean now. How do you think to add another map in `router` to save group level hook ids？

whynowy: That sounds like a plan.

whynowy (approved):
LGTM, thanks!

Issue #1345, opened 2021-09-10
IAM MSK support


tooptoop4:
Receive events from MSK with IAM auth config https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html

tooptoop4:
https://github.com/Shopify/sarama/issues/1985#issuecomment-981885298 could help

whynowy:
@tooptoop4 - would you like to contribute?

tooptoop4:
would you be ok with maintaining this patch https://github.com/Shopify/sarama/pull/2192/files on argo side?

mailtoraja18:
looking forward for this IAM MSK support.

tooptoop4:
https://github.com/aws/aws-msk-iam-sasl-signer-go might help

kaiyuanlim:
Any update on this?

kaiyuanlim-adh:
@dfarr : I saw your comment here: https://github.com/argoproj/argo-events/pull/2502#issuecomment-1483170346

kaiyuanlim-adh:
Hi, i created a PR: https://github.com/argoproj/argo-events/pull/3963

kaiyuanlim-adh:
PR has been merged. Waiting for new release.

PR #2502, opened 2023-03-03
feat: Kafka EventBus

dfarr:
# Kafka EventBus

An implementation of EventBus leveraging Kafka. Supports simple conditions, complex conditions (ones containing 'and' clauses), trigger condition resets, and both at-most-once and at-least-once trigger semantics. In contrast to existing EventBus implementations, Sensors are horizontally scalable (EventSources are sometimes horizontally scalable).

The implementation uses three topics, an **event** topic that EventSources produce to and Sensors consume from. A **trigger** topic that is used to maintain state and ensure resiliency for triggers with complex conditions. And finally an **action** topic used to decouple event processing from trigger actions. The event topic is shared between all EventSources and Sensors wired up to the same EventBus, whereas the trigger and action topics are specific to a single Sensor.

**Default naming convention**
| topic | name |
| --- | --- |
| event | `{namespace}-{eventbus}` |
| trigger | `{namespace}-{eventbus}-{sensor}-trigger` |
| action | `{namespace}-{eventbus}-{sensor}-action` |

**High level architecture**
architecture

For simple trigger conditions (such as `t2`) we can skip the trigger topic and publish messages directly to the action topic. For complex trigger conditions (such as `t1`) we use the trigger topic to hold on to events until the trigger condition is satisfied.

Please see my [blog post](https://medium.com/p/a259f663bd30#d412-b23e5a699d8f) for in-depth implementation details. The post poses a dilemma with the Kafka implementation that occurs under the following scenario. Imagine a Sensor with the following two triggers.
```
t1: a && b
t2: c && d
```

Assume the events below are received in the following order. Furthermore, assume all topics {event, trigger, action} contain only a single partition.
```
{a, c, d}
```

Trigger `t2` will be invoked once, however, because all messages land on the same partition (there is only one) the dilemma is that we cannot bump our consumer’s offset as we need to hold on to the `a` event in order to maintain resiliency. But, imagine a restart occurs and we re-consume all three events starting with event `a`. How can we ensure that we do not re-invoke `t2`? To solve this problem we opted to maintain metadata alongside the offset. 

The metadata maintains a mapping of triggers to offsets, for example:
```
{
  "t1": 0,
  "t2": 3
}
```

If ever a restart occurs and we start consuming from offset 0, however, any message pertaining to trigger `t2` are skipped until we get to offset 3 and therefore no errant trigger action is invoked.

Checklist:


controllers/eventsource/resource.go
```diff
@@ -240,42 +252,25 @@ func buildDeployment(args *AdaptorArgs, eventBus *eventbusv1alpha1.EventBus) (*a
...
+		})
 	}
-	deploymentSpec.Template.Spec.Volumes = volumes
-	deploymentSpec.Template.Spec.Containers[0].VolumeMounts = volumeMounts
 
-	envs := deploymentSpec.Template.Spec.Containers[0].Env
-	envs = append(envs, envVars...)
-	deploymentSpec.Template.Spec.Containers[0].Env = envs
+	// secrets
+	volSecrets, volSecretMounts := common.VolumesFromSecretsOrConfigMaps(common.SecretKeySelectorType, secretObjs...)
```
dfarr, line 262: When the eventBus type is kafka, we need can use the `common.VolumesFromSecretsOrConfigMaps` function to attach the tls and sasl secrets. They need to be attached at the same time as the secrets in the sensor/eventsource so that any secrets with the same name will be deduplicated.

test/e2e/testdata/es-durable-consumer.yaml
```diff
@@ -3,6 +3,8 @@ kind: EventSource
 metadata:
   name: e2e-durable-consumer
 spec:
+  template:
+    serviceAccountName: argo-events-sa
```
dfarr, line 7: Added the `argo-events-sa` service account to tests as the kafka eventsource requires a kubernetes lease and the default service account does not have permission to create these objects in the ci tests

.github/workflows/ci.yaml
```diff
@@ -123,6 +123,7 @@ jobs:
         include:
           - driver: stan
           - driver: jetstream
+          - driver: kafka
```
dfarr, line 126: This executes all e2e tests against kafka as well

eventbus/common/interface.go
```diff
@@ -42,5 +42,9 @@ type EventSourceDriver interface {
 
 type SensorDriver interface {
 	Initialize() error
-	Connect(triggerName string, dependencyExpression string, deps []Dependency) (TriggerConnection, error)
+	Connect(ctx context.Context,
```
dfarr, line 45: Added context and atLeastOnce boolean to this interface, the stan and jetstream implementation just ignore these values

common/leaderelection/leaderelection.go
```diff
@@ -40,7 +40,7 @@ type LeaderCallbacks struct {
 
 func NewElector(ctx context.Context, eventBusConfig eventbusv1alpha1.BusConfig, clusterName string, clusterSize int, namespace string, leasename string, hostname string) (Elector, error) {
 	switch {
-	case strings.ToLower(os.Getenv(common.EnvVarLeaderElection)) == "k8s":
+	case eventBusConfig.Kafka != nil || strings.ToLower(os.Getenv(common.EnvVarLeaderElection)) == "k8s":
 		return newKubernetesElector(namespace, leasename, hostname)
```
juliev0, line 44: So, if Kafka is defined we still do leader election here? (even though it's "master/master")

dfarr: We still need the leader election for the eventsource. For example if an eventsource uses the calendar we need active/passive so that both pods aren't emitting calendar events at the same time.

juliev0: Got it. And I see it's not being called on the Sensor side.

juliev0:
I assume you'll be adding docs for Kafka after this PR, right?

controllers/eventsource/resource.go
```diff
@@ -240,42 +252,25 @@ func buildDeployment(args *AdaptorArgs, eventBus *eventbusv1alpha1.EventBus) (*a
...
+	// config maps
+	volConfigMaps, volCofigMapMounts := common.VolumesFromSecretsOrConfigMaps(common.ConfigMapKeySelectorType, eventSourceCopy)
+	volumeMounts = append(volumeMounts, volCofigMapMounts...)
+	volumes = append(volumes, volConfigMaps...)
 
-	deploymentSpec.Template.Spec.Volumes = vols
-	deploymentSpec.Template.Spec.Containers[0].VolumeMounts = volMounts
+	deploymentSpec.Template.Spec.Containers[0].Env = append(deploymentSpec.Template.Spec.Containers[0].Env, env...)
+	deploymentSpec.Template.Spec.Containers[0].VolumeMounts = append(deploymentSpec.Template.Spec.Containers[0].VolumeMounts, volumeMounts...)
+	deploymentSpec.Template.Spec.Volumes = append(deploymentSpec.Template.Spec.Volumes, volumes...)
```
juliev0, line 273: Regarding risk to Jetstream/nats, this code is generic to all: do you think there should be any additional testing done for this or do you think it's been covered?

juliev0: (same with controllers/sensor/resource.go below)

pkg/apis/eventbus/v1alpha1/kafka_eventbus.go
```diff
@@ -0,0 +1,43 @@
...
+	// TLS configuration for the kafka client.
+	// +optional
+	TLS *apicommon.TLSConfig `json:"tls,omitempty" protobuf:"bytes,4,opt,name=tls"`
+	// SASL configuration for the kafka client
+	// +optional
+	SASL *apicommon.SASLConfig `json:"sasl,omitempty" protobuf:"bytes,5,opt,name=sasl"`
+	// Consumer group for kafka client
+	// +optional
+	ConsumerGroup *KafkaConsumerGroup `json:"consumerGroup,omitempty" protobuf:"bytes,6,opt,name=consumerGroup"`
+}
```
juliev0, line 31: So, basically if user specifies no ConsumerGroup and they have a Sensor with multiple replicas then they will each receive copies of the message, right? Is the expectation that if they don't specify one then they should just be using one replica?

juliev0: Just for my own education, is there a reason they would want no ConsumerGroup?

eventbus/kafka/sensor/kafka_sensor.go
```diff
@@ -0,0 +1,367 @@
...
+	sensor *sensorv1alpha1.Sensor
+
+	// kafka
+	topics    *Topics
+	client    sarama.Client
+	consumer  sarama.ConsumerGroup
+	hostname  string
+	groupName string
+
+	// triggers
```
juliev0, line 33: can we expand on this comment? looks like it's actual kafka connections?

controllers/eventsource/resource.go
dfarr: I can add some more tests! I think that's a good idea

eventbus/kafka/sensor/kafka_sensor.go
dfarr: The triggers is a map of KafkaTriggerHandler (specifically `map[string]KafkaTriggerHandler`), which is an implementation of the TriggerHandler interface. With Jetstream and Nats, this information is not required in the sensor because events are consumed by individual triggers. However, with kafka we need this information here in the Sensor because events are consumed centrally. Will expand in the comment :+1:

eventbus/kafka/sensor/kafka_sensor.go
```diff
@@ -0,0 +1,367 @@
...
+
+			s.Logger.Errorw("Failed to consume", zap.Error(err))
+			return
+		}
+
+		if err := ctx.Err(); err != nil {
+			s.Logger.Errorw("Kafka error", zap.Error(err))
+			return
+		}
+	}
```
juliev0, line 220: That's interesting that all the logic is in the Sensor and not in the KafkaTriggerConnection - did it fit better to put it here?

pkg/apis/eventbus/v1alpha1/kafka_eventbus.go
juliev0: I see now that I think you set ConsumerGroup if they don't define anything, right?

eventbus/kafka/sensor/kafka_sensor.go
juliev0: (the Trigger-processing part that is...I know the Events are generic and all)

dfarr: This implementation is top down, which is in contrast to how it works for jetstream. When an event is consumed from a kafka topic, the 3 handler functions (Event, Trigger, Action) are invoked depending on which topic the event was consumed from. These functions then use the Triggers map (part of the KafkaSensor struct) to send these messages to each trigger that requires it.

To maintain seperation, I try to keep all state in the TriggerConnection struct (the events) and the Sensor only interacts with the trigger connections by invoking the [interface functions](https://github.com/argoproj/argo-events/pull/2502/files#diff-127d6164e8d0604a26ba3abaa3c694580b746ad45138ad7e601e954a6d3f4772R13-R24) (which is a superset of the TriggerConnection interface).

pkg/apis/eventbus/v1alpha1/kafka_eventbus.go
dfarr: Exactly, we need a consumer group. Kafka does have a way to consume from a single partition (and this is an option with the Kafka EventSource) but for the Kafka EventBus we need to use a ConsumerGroup. You can either set it explicitly or use the [default](https://github.com/argoproj/argo-events/pull/2502/files#diff-b778225ecfb035ee9a44077f3f987fd20bc3381f1786485f86639419b0c9196fR48-R53).

dfarr:
> I assume you'll be adding docs for Kafka after this PR, right?

Yes! Actually let me add at least the minimum docs to this PR and if we need more in depth details I'm happy to follow up with another PR.

eventbus/kafka/sensor/kafka_sensor.go
```diff
@@ -0,0 +1,367 @@
...
+		}()
+	}
+
+	// need to determine smallest possible offset against all
+	// triggers as other triggers may have messages that land on the
+	// same partition
+	for _, trigger := range s.triggers {
+		offset = trigger.Offset(msg.Partition, offset)
+	}
+
```
juliev0, line 350: can this handler be called by two threads at the same time or no? if so, any issues related to both threads executing lines 347-349 at the same time?

eventbus/kafka/sensor/kafka_sensor.go
juliev0: Thanks. I've looked at it further since my comment. It's all very elegant.

juliev0:
> > I assume you'll be adding docs for Kafka after this PR, right?
> 
> Yes! Actually let me add at least the minimum docs to this PR and if we need more in depth details I'm happy to follow up with another PR.

I'm fine if you want to get this in first before the docs - up to you (and Derek)

pkg/apis/eventbus/v1alpha1/kafka_eventbus.go
dfarr: Yes, let me start by adding a blurb in the generate docs 👍

eventbus/kafka/sensor/kafka_sensor.go
dfarr: No, luckily this is not possible by construction. We use the kafka message key to look up the trigger [here](https://github.com/argoproj/argo-events/pull/2502/files#diff-b778225ecfb035ee9a44077f3f987fd20bc3381f1786485f86639419b0c9196fR317), a key always maps to only one partition and all messages pertaining to the same partition are processed sequentially (and by one pod). That means we don't have to worry about concurrency in our stateful code (mostly the KafkaTriggerConnection). This is also how we enable scaling - all events that pertain to a specific trigger will land on the same partition and are guaranteed to be processed by the same pod.

eventbus/kafka/sensor/kafka_sensor.go
dfarr: Thank you :)

eventbus/kafka/sensor/kafka_sensor.go
juliev0: Right, I guess my confusion was that we were looking at other triggers outside of our own on line 347

juliev0: I think I see what you're saying. The Offset() function only goes through events on this same partition, and all messages pertaining to the same partition are processed sequentially.

Makefile
```diff
@@ -141,6 +141,8 @@ docs/assets/diagram.png: go-diagrams/diagram.dot
 start: image
 	kubectl apply -f test/manifests/argo-events-ns.yaml
 	kubectl kustomize test/manifests | sed 's@quay.io/argoproj/@$(IMAGE_NAMESPACE)/@' | sed 's/:$(BASE_VERSION)/:$(VERSION)/' | kubectl -n argo-events apply -l app.kubernetes.io/part-of=argo-events --prune=false --force -f -
+
+	sleep 10
```
whynowy, line 145: What is the reason to sleep 10?

common/leaderelection/leaderelection.go
```diff
@@ -40,7 +40,7 @@ type LeaderCallbacks struct {
 
 func NewElector(ctx context.Context, eventBusConfig eventbusv1alpha1.BusConfig, clusterName string, clusterSize int, namespace string, leasename string, hostname string) (Elector, error) {
 	switch {
-	case strings.ToLower(os.Getenv(common.EnvVarLeaderElection)) == "k8s":
+	case eventBusConfig.Kafka != nil || strings.ToLower(os.Getenv(common.EnvVarLeaderElection)) == "k8s":
```
whynowy, line 43: Needs to be documented as a requirement for kafka eventbus.

eventbus/kafka/base/utils.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+	"time"
+)
+
+func EventKey(source string, subject string) string {
+	return fmt.Sprintf("%s.%s", source, subject)
+}
+
+func Batch[T any](n int, d time.Duration, in <-chan T) <-chan []T {
```
whynowy, line 12: Add some docs for this function?

pkg/apis/eventbus/v1alpha1/kafka_eventbus.go
```diff
@@ -0,0 +1,43 @@
+package v1alpha1
+
+import (
+	apicommon "github.com/argoproj/argo-events/pkg/apis/common"
+)
+
+// KafkaBus holds the KafkaBus EventBus information
+type KafkaBus struct {
+	// Exotic holds an exotic Kafka config
+	Exotic *KafkaConfig `json:"exotic,omitempty" protobuf:"bytes,1,opt,name=exotic"`
```
whynowy, line 10: Remove `exotic`? I don't think we will do in cluster kafka support.

sensors/listener.go
```diff
@@ -59,6 +59,12 @@ func (sensorCtx *SensorContext) Start(ctx context.Context) error {
 	replicas := int(sensorCtx.sensor.Spec.GetReplicas())
 	leasename := fmt.Sprintf("sensor-%s", sensorCtx.sensor.Name)
 
+	// sensor for kafka eventbus can be scaled horizontally,
+	// therefore does not require an elector
+	if sensorCtx.eventBusConfig.Kafka != nil {
```
whynowy, line 64: Documentation.

test/manifests/kafka/kafka.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps/v1
```
whynowy, line 1: Do we clean up these after testing is done?

sensors/triggers/kafka/kafka.go
```diff
@@ -216,6 +216,7 @@ func (t *KafkaTrigger) Execute(ctx context.Context, events map[string]*v1alpha1.
 
 	msg := &sarama.ProducerMessage{
 		Topic:     trigger.Topic,
+		Key:       sarama.StringEncoder(t.Trigger.Template.Name),
```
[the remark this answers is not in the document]
dfarr: No, I can remove this.

One benefit of setting the trigger name as the key is that downstream messages will arrive in order, whereas if no key is specified two trigger invocations may end up on different partitions and therefore may be processed out of order. However, since the key was not set in the first place I'm guessing this hasn't been much of an issue in practice.

Makefile
dfarr: We added kafka to the `test/manifests` for the functional tests. The kafka specs use a statefulset which for some odd reason don't set a ready condition the same way as deployments do. Because of this the `kubectl -n argo-events wait --for=condition=Ready` did not take the kafka statefulset into account, so the 10s accounts for this.

I am going to move the kafka stuff to only spin up (and tear down) when `make test-functional` is run, so I can remove the sleep statement here.

juliev0 (approved):

whynowy (approved):
LGTM, thanks @dfarr @juliev0 !

tooptoop4:
any chance of supporting IAM MSK auth on Kafka for this ? https://github.com/argoproj/argo-events/issues/1345 has background on eventsource side

PR #2510, opened 2023-03-10
add labels to jetstream statefulset generated by eventbus controller

sudeshjethoe:
When migrating from the NATS EventBus to the JetStream EventBus I noticed that the deployment was blocked by our admission controller as it requires certain labels to be present on the StatefulSets and Deployments.

It seemed that the EventBus controller only appends the labels from the EventBus CRD to the EventBus StatefulSet for the NATS EventBus and not for the JetStream EventBus, this PR adds this for JetStream EventBuses as well.

Checklist:

- [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).


whynowy (approved):
LGTM, thanks!

PR #2522, opened 2023-03-15
fix github sensor example to use generateName

jdiglesias:
Just fixing a booboo that tripped me up when implementing the github sensor.


whynowy (approved):
Thanks!

PR #2572, opened 2023-04-12
feat: sensor live reload

AalokAhluwalia:
We've developed an optional field in the sensor spec to "live reload", perform a sensor update without pod restart. The feature relies on the controller creating and mounting a configmap containing the sensor definition, rather than using an env variable. The sensor pod watches for changes to the mounted file and will cancel its context on change to update without pod restart. The feature is optional and by default set to false. 

Checklist:


AalokAhluwalia:
Closing in favor of https://github.com/argoproj/argo-events/pull/2574

Issue #2146, opened 2022-08-24
SENSOR_OBJECT hits environment variable max length on large sensors 


danielpof:
**Describe the bug**
When creating a big Sensor manifest the `SENSOR_OBJECT` will be too big and the sensor container will fail to start with `exec user process caused: argument list too long`.

**To Reproduce**
1. Create a big sensor manifest, our case it was 152Kb
2. Attempt to deploy it
3. The sensor container will crash loop with `exec user process caused: argument list too long` error

**Expected behavior**
Sensor container would be created and run as normal

**Screenshots**

**Environment (please complete the following information):**
 - Kubernetes: AKS 1.23.8
 - Argo: 3.3.9
 - Argo Events: 1.7.1

**Additional context**
Kubernetes cluster is using istio service mesh 1.14.1. 

The Environment Variable memory limit seems to be dependent on linux kernel compilation and thus hard to change.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
@daniel-anova - thanks for reporting this issue! Can you let me know what is the reason your sensor spec gets so large?

tooptoop4:
@daniel-anova can u share what OS version/bash version/AKS instance type your k8s workers are? Also output from:
```
getconf -a | grep ARG_MAX
true | xargs --show-limits
```
and https://stackoverflow.com/a/49119526

daniel-anova:
> @daniel-anova - thanks for reporting this issue! Can you let me know what is the reason your sensor spec gets so large?

We're using argo to schedule data processing for which only the inputs change, we've made a frontend that allows users to add scheduling and it was all being added to a single sensor manifest (since then we've refactored to split the sensor into multiple files).

> @daniel-anova can u share what OS version/bash version/AKS instance type your k8s workers are? Also output from:

For the nodes:
- AKS/Kubernetes version: 1.23.8
- Kernel 5.4.0-1086-azure
- OS Image: Ubuntu 18.04.6 LTS
- Container Runtime: containerd://1.5.11+azure-2
- Bash Version: Your sensor container doesn't have bash


For the other commands I used an ubuntu:latest container:
```
<<K9s-Shell>> Pod: default/ubuntu-r8bfj | Container: ubuntu
root@ubuntu-r8bfj:/# getconf -a | grep ARG_MAX
true | xargs --show-limits
ARG_MAX                            2097152
[...]

PR #2574, opened 2023-04-13
feat: sensor live reload

AalokAhluwalia:
We've developed an optional field in the sensor spec to "live reload", perform a sensor update without pod restart. The feature relies on the controller creating and mounting a configmap containing the sensor definition, rather than using an env variable. The sensor pod watches for changes to the mounted file and will cancel its context on change to update without pod restart. The feature is optional and by default set to false.

Checklist:


whynowy:
@AalokAhluwalia - thanks for submitting the PR!

I did a quick peek on it, I don't think this implementation is something we want to proceed. To move forward, I would recommend you create an issue, and elaborate your proposals with detailed design first (and going forward, please also follow this process before submitting a PR for implementation, especially for big changes), so that anyone who is interested in could have the context at least.

AalokAhluwalia:
Sure we'll can create a separate issue explaining the reasoning for this feature. This change would additionally solve for: https://github.com/argoproj/argo-events/issues/2146

dfarr:
# Sensor Live Reload Design
## Proposal
Sensor deployments should be able to automatically update their internal definition of the Sensor object. This would have the advantage of eliminating downtime when certain changes to a Sensor definition occurs.

## Current Approach
Sensor information is currently passed from the controller to the deployment by setting the encoded Sensor specification as an environment variable. When changes to the Sensor specification occur, the controller mutates the deployment and the mutated environment variable will result in a new ReplicaSet.

## Option 1: Read Sensor Definition
The Sensor application could read the Sensor specification through the kubernetes API, removing the need for the Sensor specification environment variable. This approach would require no additional kubernetes resources, however, it would have the following downsides:
- The sensor deployment service account would require read access to Sensor objects
- The sensor application would need to poll (watch) the kubernetes API

## Option 2: Volume Mount ConfigMap
We ended up going with this approach.

In addition to creating (and updating) a deployment, the controller can also create/update a ConfigMap whenever changes to the Sensor specification occurs. The controller will also volume mount this file to a known file path so that the Sensor application can retrieve this information. Because the kubelet automatically propagates ConfigMap changes to the volume mount, the Sensor application can simply watch the file for changes.

architecture

When the Sensor definition is mutated (resulting in a change to the ConfigMap), the Sensor application needs to restart it’s internal processes (note: this is different from a Pod recreation and requires negligible downtime) by canceling the current context. Luckily, the main sensor process is [already bound to a context](https://github.com/argoproj/argo-events/blob/master/sensors/cmd/start.go#L86-L88). Before we invoke the start function, we can pass this context to [context.WithCancel](https://pkg.go.dev/context#WithCancel) to get a new cancellable context. Whenever a file change is detected we first call `cancel()` to stop the current process, and then [create and start a new SensorContext](https://github.com/argoproj/argo-events/pull/2574/files#diff-7c6e3f9dab2b101aae1efe37888503488a33f5705363b48d47a6533f56e1ec85R253-R256) with the updated Sensor specification; rinse and repeat. Please note that this approach is completely transparent to the SensorContext, and therefore no code outside the of the `cmd` module has been changed in the Sensor. When implemented we get the following state machine.

state machine

This is the approach that has been implemented in this PR as opt-in only by specifying `liveReload: true` in the Sensor specification. If this field is omitted or set to false, the current environment variable approach will be maintained to mitigate disruption.

whynowy:
> # Sensor Live Reload Design
> ## Proposal
> Sensor deployments should be able to automatically update their internal definition of the Sensor object. This would have the advantage of eliminating downtime when certain changes to a Sensor definition occurs.
> 
> ## Current Approach
> Sensor information is currently passed from the controller to the deployment by setting the encoded Sensor specification as an environment variable. When changes to the Sensor specification occur, the controller mutates the deployment and the mutated environment variable will result in a new ReplicaSet.
> 
> ## Option 1: Read Sensor Definition
> The Sensor application could read the Sensor specification through the kubernetes API, removing the need for the Sensor specification environment variable. This approach would require no additional kubernetes resources, however, it would have the following downsides:
> 
> * The sensor deployment service account would require read access to Sensor objects
> * The sensor application would need to poll (watch) the kubernetes API
> 
> ## Option 2: Volume Mount ConfigMap
> We ended up going with this approach.
> 
> In addition to creating (and updating) a deployment, the controller can also create/update a ConfigMap whenever changes to the Sensor specification occurs. The controller will also volume mount this file to a known file path so that the Sensor application can retrieve this information. Because the kubelet automatically propagates ConfigMap changes to the volume mount, the Sensor application can simply watch the file for changes.
> 
> architecture
> 
> When the Sensor definition is mutated (resulting in a change to the ConfigMap), the Sensor application needs to restart it’s internal processes (note: this is different from a Pod recreation and requires negligible downtime) by canceling the current context. Luckily, the main sensor process is [already bound to a context](https://github.com/argoproj/argo-events/blob/master/sensors/cmd/start.go#L86-L88). Before we invoke the start function, we can pass this context to [context.WithCancel](https://pkg.go.dev/context#WithCancel) to get a new cancellable context. Whenever a file change is detected we first call `cancel()` to stop the current process, and then [create and start a new SensorContext](https://github.com/argoproj/argo-events/pull/2574/files#diff-7c6e3f9dab2b101aae1efe37888503488a33f5705363b48d47a6533f56e1ec85R253-R256) with the updated Sensor specification; rinse and repeat. Please note that this approach is completely transparent to the SensorContext, and therefore no code outside the of the `cmd` module has been changed in the Sensor. When implemented we get the following state machine.
> 
> state machine
> 
> This is the approach that has been implemented in this PR as opt-in only by specifying `liveReload: true` in the Sensor specification. If this field is omitted or set to false, the current environment variable approach will be maintained to mitigate disruption.

@dfarr - thanks for the update! 

I ever thought about supporting hot reload for eventsource/sensor changes, but I realized that would be wrong.

As a design principle, CRD objects should be deterministic, the generated resources should be immutable, which means no matter when you apply them, they should always lead to the same results. Immutability means the containers should not be modified during its life - no updates, no patches, no configuration changes. If you really want to do so, rebuild an image or redeploy.

Today, it's not 100% immutable in Argo Events, for example, a triggered workflow definition could come from a github repo, or a configmap, but they are a little bit different because they are consider as "data" - as a sensor application, I don't care what is really triggered, I just applied some "data". However, the change in this PR is a different story - the sensor definition could be changed by a externalized configmap, which leads to runtime application changes in the pod.

Based on these, I don't think this would be an appropriate appoach.

AalokAhluwalia:
Closing, we're no longer pursuing live reload. WIll open another fix for https://github.com/argoproj/argo-events/issues/2146

PR #2589, opened 2023-04-23
feat: add azure queue storage event source

eduardodbr:
This PR adds [Azure Queue Storage](https://learn.microsoft.com/en-us/azure/storage/queues/storage-queues-introduction) event source. The implementation mimics the AWS SQS one. 

It allows connecting via connection String or Azure AD via managed identity.


whynowy (approved):
LGTM, thanks!

PR #2590, opened 2023-04-24
fix: make start to also work with arm64

### LINKED ISSUE #2588: "Make start" doesnt work with ARM
**Describe the bug**

I'm trying to test my local changes, however, the command `make start` fails if executed on ARM architecture. It seems easy to fix the issue, it is only necessary to check if the arch is ARM and build with `dist/argo-linux-arm64` instead of `dist/argo-linux-amd64`.  Should I open a PR? 

**To Reproduce**
Execute `make start` in a machine with ARM architecture. 

**Expected behavior**
It should build the image in order to do local testing.

---

eduardodbr:
fix #2588


whynowy (approved):
Thanks!

PR #2600, opened 2023-05-02
 chore: adding tests for the removeConflicts function & the Convert2WaitBackoff function

zhaque44:
1. The `Convert2WaitBackoff` func converts `apicommon.Backoff` struct to a wait.Backoff struct used by the k8s.io/apimachinery/pkg/util/wait package. It extracts the fields from the input struct and converts them to the appropriate types for the output struct. Having a unit test for the `Convert2WaitBackoff` function is important to ensure that it converts the input `apicommon.Backoff` object to a `wait.Backoff` object with the correct values. The function has to parse and convert various fields, and the unit test verifies that the conversions are done correctly. 
2. The `removeConflicts` func removes any conflicting keys from a map. If there are two or more keys that share the same prefix, where one is a parent of the other, then the less-specific key is removed. This func is used in the `Expand` function to remove any conflicts before expanding environment variables in the given map. This test is important because it plays a critical role in ensuring that the map used to evaluate expressions is valid. If conflicts are not properly handled, the expressions being evaluated may produce incorrect results for Argo Events users.

Checklist:

### DIFF
--- a/common/expr/eval_test.go
+++ b/common/expr/eval_test.go
@@ -2,6 +2,7 @@ package expr
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


whynowy (approved):
Thanks!

PR #2627, opened 2023-05-24
feat: Allow to base64 decode messages  from Azure Queue Storage

eduardodbr:
The messages in azure queue storage may be base64 encoded. This new configuration field allows to decode them before converting to a json, if necessary.


whynowy (approved):
LGTM, thanks!

PR #2647, opened 2023-06-04
docs: add missing service account role for operator installation

bellec1u:
The webhook pod currently need to run with an admin user on openshift, we need to add the scc anyuid to the used user.


whynowy (approved):
Thanks!

Issue #2581, opened 2023-04-19
Argo Events docker image uses an old version of Argo CLI


AdrianSkierniewski:
**Describe the bug**
Argo events docker image uses an old version of Argo CLI, which causes the issue when submitting a template that contains `when` statement with an expression.  This behavior was added in Argo v3.4.2

```
Failed to submit workflow: templates.start when expression doesn't support 'expr' format '{{='. 'Wh
en' expression is only support govaluate format {{
```

**To Reproduce**
Create a sensor with workflow-trigger that uses `workflowTemplateRef` to reference the template that includes when statement with expression `{{=`

**Expected behavior**
I'd expect that I can submit all workflows that are supported by argo workflows server through argo events workflow trigger.

**Environment:**
 - Kubernetes: v1.25.4
 - Argo: v3.4.4
 - Argo Events: latest

**Additional context**
It looks like this version hasn't been updated for some time https://github.com/argoproj/argo-events/blob/master/Dockerfile#L11

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

AdrianSkierniewski:
Any plans to do this upgrade before the next release?

daniel-codefresh:
@AdrianSkierniewski The upgrade to v3.4.8 should now be present in `latest` tag. Hopefully it will be included in the next stable release.

PR #2671, opened 2023-06-20
chore: upgrade Argo CLI version to v3.4.8

daniel-codefresh:
Upgrade Argo CLI to the latest version which is v3.4.8. Solves: #2581 

Basic functionality of the argo-workflows trigger was tested with various WorkflowTemplates, including new features that have been added since the old version we used, such as the 'expr' format in 'when' conditions as requested in the issue itself.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN apk update && apk upgrade && \
     apk add ca-certificates && \
     apk --no-cache add tzdata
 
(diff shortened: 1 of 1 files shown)


daniel-codefresh:
@whynowy Are there any potential risks that you see in this that I might have missed?

whynowy (approved):

PR #2673, opened 2023-06-22
#1896 For OpenShift security policy, argo webhook port from 443 to 9443

antoinetran:
Checklist:

In the frame of https://github.com/argoproj/argo-helm/issues/1896
Deploying to OpenShift is not possible because of no root policy, which prevents opening port <1024. This allows it without impact because the service port source is still 443.

### DIFF
--- a/manifests/extensions/validating-webhook/events-webhook-deployment.yaml
+++ b/manifests/extensions/validating-webhook/events-webhook-deployment.yaml
@@ -24,5 +24,5 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         - name: PORT
-          value: "443"              
+          value: "9443"              
       serviceAccountName: argo-events-webhook-sa
--- a/manifests/extensions/validating-webhook/events-webhook-service.yaml
+++ b/manifests/extensions/validating-webhook/events-webhook-service.yaml
@@ -5,6 +5,7 @@ metadata:
 spec:
   ports:
     - port: 443
-      targetPort: 443
+      # For Openshift compatibility (no root means no port < 1024).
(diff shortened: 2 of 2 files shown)


whynowy:
The port is customizable, no need to make the default manifest change.

antoinetran:
> The port is customizable, no need to make the default manifest change.

Last time I used argo-helm, the port was not customizable and hard-coded to 443. If I look at source: https://github.com/argoproj/argo-events/blob/17f01801ff34101ab27c714c8297958e0f3e2fe6/manifests/extensions/validating-webhook/events-webhook-deployment.yaml#L27 , it still looks hard-coded to me.

whynowy:
> > The port is customizable, no need to make the default manifest change.
> 
> Last time I used argo-helm, the port was not customizable and hard-coded to 443. If I look at source:
> 
> https://github.com/argoproj/argo-events/blob/17f01801ff34101ab27c714c8297958e0f3e2fe6/manifests/extensions/validating-webhook/events-webhook-deployment.yaml#L27
> 
> , it still looks hard-coded to me.

Then it should be changed in the helm charts.

jmeridth:
It can be overridden in your value.yaml in your helm chart as stated [here](https://github.com/argoproj/argo-helm/issues/1896#issuecomment-1508014617)

antoinetran:
@jmeridth ok, you are right. After relooking at my code and argo-helm code, the Helm chart is supposed to let us override the webhook port. However it does the job for webhook service but not for webhook deployment. I will close this ticket in favor of fixing in argo-helm code. Thanks!

PR #2693, opened 2023-07-07
SFTP event source

### LINKED ISSUE #1652: Support SFTP as input source
I'm looking for the same feature as described in https://stackoverflow.com/questions/68912793/how-to-configure-argo-eventsource-for-aws-sftp-file-creation-event - however not against AWS specifically, but towards any sftp source.

Trigger on new files.

none, any workarounds are appreciated

**Additional context**
N/A

---

dillonstreator:
Fixes #1652 

This PR introduces an SFTP event source with support for `fsnotify.Remove` and `fsnotify.Create` operations.

A couple of caveats/questions on the implementation:
- Changes in the watched SFTP Server path which are negated within the `pollIntervalDuration` window will not be captured and result in events. This should be understood by consumers given the nature of polling.
- The SFTP event source is not stateful. Any changes that occur during a period of event source outage or restart are not captured. This could be implemented via the persistence interface that exists in the `calendar` event source, if deemed necessary.
- This PR uses `fsnotify.Op` and `fsnotify.Event` but doesn't directly match the fsnotify interface. A file rename, different modification time, or size difference are all captured as `fsnotify.Create`.

Are we okay with these?
If yes, Is there a good place for me to document these caveats?
If no, I'm open to feedback on how we can adjust to support needs.


whynowy:
@dillonstreator - thanks! could you run `make codegen`?

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -288,6 +290,31 @@ type FileEventSource struct {
...
+	// Metadata holds the user defined metadata which will passed along the event payload.
+	// +optional
+	Metadata map[string]string `json:"metadata,omitempty" protobuf:"bytes,6,rep,name=metadata"`
+	// Filter
+	// +optional
+	Filter *EventSourceFilter `json:"filter,omitempty" protobuf:"bytes,7,opt,name=filter"`
+	// PollIntervalDuration the interval at which to poll the SFTP server
+	// defaults to 10 seconds
+	// +optional
+	PollIntervalDuration string `json:"pollIntervalDuration" protobuf:"varint,8,opt,name=pollIntervalDuration"`
```
whynowy, line 315: Use `*string` to avoid parsing empty string?

whynowy:
Could you also add a doc in https://github.com/argoproj/argo-events/tree/master/docs/eventsources/setup?

pkg/apis/eventsource/v1alpha1/types.go
dillonstreator: Is that preferred? I'm following the pattern defined by `HDFSEventSource` with `CheckInterval`

https://github.com/dillonstreator/argo-events/blob/980c63b8576969e67f27e91c65e049b5f9d6e072/pkg/apis/eventsource/v1alpha1/types.go#L1091-L1093
https://github.com/dillonstreator/argo-events/blob/980c63b8576969e67f27e91c65e049b5f9d6e072/eventsources/sources/hdfs/start.go#L96-L104

whynowy (approved):
LGTM, thanks!

eventsources/sources/sftp/start.go
```diff
@@ -0,0 +1,282 @@
...
+		return fmt.Errorf("password not found, %w", err)
+	}
+	address, err := common.GetSecretFromVolume(el.SFTPEventSource.Address)
+	if err != nil {
+		return fmt.Errorf("address not found, %w", err)
+	}
+
+	sftpConfig := &ssh.ClientConfig{
+		User:            username,
+		Auth:            []ssh.AuthMethod{ssh.Password(password)},
```
tooptoop4, line 87: would be good to have option for sshkey based auth instead of password based too: https://pkg.go.dev/golang.org/x/crypto/ssh#PublicKeys

dillonstreator: https://github.com/argoproj/argo-events/pull/2918

PR #2703, opened 2023-07-11
chore: adding tests for source_jetstream

zhaque44:
Adding tests for `source_jetstream:`

`TestNewSourceJetstream:`

- It checks that a new SourceJetstream instance is created without no errors.
- It ensures that the EventSourceName field of the SourceJetstream instance is set correctly.

`TestSourceJetstream_Connect:` 

- It asserts sourceJetstream is not nil, letting us know that the creation of the SourceJetstream instance was successful.
- It asserts that err is nil, indicating that no error occurred during the creation of the SourceJetstream instance.
- It calls the Connect method on the sourceJetstream instance, passing the "test-client-id" as the clientID argument, and assigns the returned conn and err values.

Checklist:


zhaque44:
@whynowy is this PR good to go?

PR #2718, opened 2023-07-19
chore: add kafka-ttl sensor example

xuliang2019:
Why this PR: No example in `sensors` folder with the usage of `ttlStrategy`
Change I made: added `kafka-ttl` yaml file as an example on how to use `ttlStrategy`


whynowy (approved):
thanks

PR #2750, opened 2023-08-13
chore: adding eventbus/base/kafka tests

zhaque44:
Adding tests for `eventbus/base/kafka.go:`

`func TestBrokers`

The Brokers method basically splits the comma-separated broker URLs and return them as a slice of strings.

In the test:

A sample `eventbusv1alpha1.KafkaBus` configuration is created with two broker URLs: "broker1:9092" and "broker2:9092".
Then the Brokers method of the created kafka instance is called to get the actual list of broker addresses. This test will ensure that the Brokers method correctly parses the comma-separated broker URLs and returns them as a list of strings.

`func TestConfig`

This test is verifying the behavior of the `Config` func in the context of the Kafka struct. The Config func is responsible for generating a `sarama.Config` instance with specific configuration settings based on the KafkaBus configuration. It's validating that the logic for translating the higher-level configuration to the lower-level Kafka client library's configuration is working.

`func TestConfig_NoSASL`

This test is checking the behavior when SASL configuration is not provided, by setting SASL to nil. 

`func TestNewKafka`

This test validates the `NewKafka` func, which initializes a `kafka` object with a given configuration and logger. It ensures that the `kafka` object and its logger are properly initialized.

### DIFF
--- a/eventbus/kafka/base/kafka_test.go
+++ b/eventbus/kafka/base/kafka_test.go
@@ -1,0 +1,105 @@
+package base
+
+import (
+	"testing"
+
+	"github.com/Shopify/sarama"
+	eventbusv1alpha1 "github.com/argoproj/argo-events/pkg/apis/eventbus/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+)
+
+func TestBrokers(t *testing.T) {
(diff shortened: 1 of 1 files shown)


zhaque44:
@whynowy can I have a review please, this PR is ready to go. I merged master into the branch as well

Issue #2450, opened 2023-02-01
how to ratelimit reading from eventsource?


tooptoop4:
i see https://argoproj.github.io/argo-events/sensors/more-about-sensors-and-triggers/#trigger-rate-limit has an option to rate limit from sensor, but is there option to rate limit from eventsource itself? ie so i don't read off SQS too often

PR #2757, opened 2023-08-19
Add sleep in sqs eventsource

tooptoop4:
fixes https://github.com/argoproj/argo-events/issues/2450
should also aid https://github.com/argoproj/argo-events/issues/1547 as can workaround eventbus's comparison on nanosecond

### DIFF
--- a/eventsources/sources/awssqs/start.go
+++ b/eventsources/sources/awssqs/start.go
@@ -137,6 +137,7 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 					}
 				}
 			}, log)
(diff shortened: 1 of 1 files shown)


eventsources/sources/awssqs/start.go
```diff
@@ -137,6 +137,7 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
 					}
 				}
 			}, log)
+			time.Sleep(50 * time.Millisecond)
```
whynowy, line 140: What is the problem without sleeping?

tooptoop4:
crunchy

whatnick:
crunchy

PR #2758, opened 2023-08-19
Make sqs eventsource fetch messages independently

tooptoop4:
should aid https://github.com/argoproj/argo-events/issues/1547 as can workaround eventbus's comparison on nanosecond


tooptoop4:
it is deleted before wf off a sensor is run

tooptoop4:
crunchy

Issue #2775, opened 2023-08-26
EventSource "Resource" label selector doesn't support "exists" operator (and possibly others)


azazel75:
**Describe the bug**
I've added the following EventSource to get events when the Workflow instances are completed:
``` YAML
apiVersion: argoproj.io/v1alpha1
kind: EventSource
metadata:
  name: notification
spec:
  template:
    serviceAccountName: viewer
    container:
      env:
        - name: DEBUG_LOG
          value: "true"
  resource:
    completed-workflows:
      namespace: devops
      group: argoproj.io
      version: v1alpha1
      resource: workflows
      eventTypes:
        - UPDATE
      filter:
        afterStart: true
        labels:
          - key: workflows.argoproj.io/completed
            operation: "=="
            value: "true"
          - key: workflows.argoproj.io/phase
            operation: exists
```

but the event source containers crashes during startup, this the relevant log:
```
2023-08-26T14:33:06.147Z	ERROR	argo-events.eventsource	eventsources/eventing.go:564	Failed to start listening eventsource	{"eventSourceName": "notification", "eventSourceType": "resource", "eventName": "completed-workflows", "error": "failed after retries: failed to create the label selector for the event source completed-workflows, values: Invalid value: []string{\"\"}: values set must be empty for exists and does not exist"}
```

I believe the error is where the requirements are created: there it always initializes a string value

**To Reproduce**
Add the already mentioned eventsource with `kubectl apply -f eventsource.yaml`

**Expected behavior**
To not crash and filter updates that only ave that label

**Environment (please complete the following information):**
 - Kubernetes: 1.26.4
 - Argo: 3.4.8
 - Argo Events: 1.7.6
---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

gokulav137:
If we check the documentation for selectors here https://github.com/argoproj/argo-events/blob/master/api/event-source.md#selector. It says  "Supported operations like ==, !=, <=, >= etc"

But if we use a "<=" operator we will get an error.
```operator: Unsupported value: "<=": supported values: "in", "notin", "=", "==", "!=", "gt", "lt", "exists", "!"```

Argo uses NewRequirements under the hood: https://pkg.go.dev/k8s.io/apimachinery@v0.24.3/pkg/labels#NewRequirement 
Which has following constraints that is throwing the error
```go
// NewRequirement is the constructor for a Requirement.
[...]

PR #2795, opened 2023-09-11
fix: Eventsource resourse label selector operators not working

gokulav137:
Fixes : #2775

Fixed Exist and DoesNotExist throwing error
```2023-08-26T14:33:06.147Z	ERROR	argo-events.eventsource	eventsources/eventing.go:564	Failed to start listening eventsource	{"eventSourceName": "notification", "eventSourceType": "resource", "eventName": "completed-workflows", "error": "failed after retries: failed to create the label selector for the event source completed-workflows, values: Invalid value: []string{\"\"}: values set must be empty for exists and does not exist"}```

Fixed documentation mentioning unsupported operators

Add multiple value support for In and NotIn using comma seperator

Checklist:

### CODE REVIEW

File: eventsources/sources/resource/start.go [hunk 1]
```diff
@@ -250,7 +250,14 @@ func LabelReq(sel v1alpha1.Selector) (*labels.Requirement, error) {
 	if sel.Operation != "" {
 		op = selection.Operator(sel.Operation)
 	}
-	req, err := labels.NewRequirement(sel.Key, op, []string{sel.Value})
+	values := []string{sel.Value}
+	if (op == selection.Exists || op == selection.DoesNotExist) && sel.Value == "" {
```
whynowy, line 254: Return an error when `sel.Value != ""`?

eventsources/sources/resource/start.go
```diff
@@ -250,7 +250,14 @@ func LabelReq(sel v1alpha1.Selector) (*labels.Requirement, error) {
 	if sel.Operation != "" {
 		op = selection.Operator(sel.Operation)
 	}
-	req, err := labels.NewRequirement(sel.Key, op, []string{sel.Value})
+	values := []string{sel.Value}
```
whynowy, line 253: Start with `var values = make([]string, 0)`, and use `if` `else`?

gokulav137: Makes sense will make the change

eventsources/sources/resource/start.go
gokulav137: If operator is exists or doesnotexists, an error will be thrown from labels.NewRequirement if values is not empty.

Should I add an error from our end? @whynowy

eventsources/sources/resource/start.go
gokulav137: Used ```switch``` and ```var []string``` instead of if else block due to linting error.

eventsources/sources/resource/start.go
whynowy: This is good, thanks!

whynowy (approved):

Issue #2810, opened 2023-09-18
Controller keeps updating eventsource pod deployment thousands of times instead of just once


popsu:
**Describe the bug**
When changing an `eventsource` CustomResource, the `argo-events` `controller` pod keeps updating the deployment multiple times in a row. Sometimes it takes hours/days for these constant updates to stop. It looks like some kind of a race issue where multiple entities are mutating the eventsource at the same time.

What ends up happening that since the deployment keeps getting changed constantly, kubernetes creates a new replicaset and the replicasets are created faster than the kubernetes garbage collector is deleting them (since as default each deployment keeps only 10 replicasets) and we end up having up to 50k replicasets sometimes, until the `argo-events` controller stops mutating the deployment and eventually things calm down and replicasets number goes back to 10.

This happens (almost?) every time when we change our eventsource.

**To Reproduce**
1. The cluster has an `EventSource` deployed:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: EventSource
metadata:
  name: webhook
  namespace: argo-managed-analytics-staging
spec:
  template:
    serviceAccountName: webhook-event-sources-sa
    metadata:
      annotations:
        sidecar.istio.io/inject: "true"
        sidecar.istio.io/proxyCPU: 100m
        sidecar.istio.io/proxyMemory: 100Mi
  service:
    ports:
      - port: 12000
        targetPort: 12000
  webhook:
    analytics-aircraft-detection-stage-gforms:
      port: "12000"
      endpoint: "/aircraft-detection/stage/gforms"
      method: "POST"
      authSecret:
        name: analytics-aircraft-detection-stage-gforms
        key: AUTH_TOKEN
    analytics-dark-vessel-detection-stage-gforms:
      port: "12000"
      endpoint: "/dark-vessel-detection/stage/gforms"
      method: "POST"
      authSecret:
        name: analytics-dark-vessel-detection-stage-gforms
        key: AUTH_TOKEN
    # 11 more webhooks, that have been removed from this snippet, but they're all similar
    # and are all running in same port (each with unique endpoint tho)
```

2. Make changes to the `EventSource`
3. This triggers `argo-events` controller to keep constantly updating the deployment that the `eventsource` has created.

**Expected behavior**
After changing an eventsource, the `argo-events` controller would only update the deployment once, not multiple times in a row.

**Screenshots**

[...]

PR #2811, opened 2023-09-18
fix: eventsource: deterministic volume/volumeMount order

popsu:
The order was non-deterministic earlier and this caused the controller to change the deployment on every reconcile until it the order happened to be same twice in a row.

Fixes: https://github.com/argoproj/argo-events/issues/2810

I've tested this by making the change to `v1.7.1` tag (since we were running `v1.7.1` when I made the fix), building the image, and deploying it and the issue went away. 

Checklist:


popsu:
Let me know if there's something missing, first time doing PR here :bow:

popsu:
https://github.com/argoproj/argo-events/blob/v1.8.1/common/util.go#L395-L417

I wonder if I should make the change in these functions instead? I don't know the codebase, so not sure at what level should the fix be made.

whynowy:
@popsu - thanks for submitting the PR! Could you also add the same logic in the sensor controller?

popsu:
I tried adding `make codegen`, but it failed:

```shell
2023/09/25 14:07:48 google/protobuf/descriptor.proto: File not found.                                                                                                                                                                                                                     
github.com/gogo/protobuf/gogoproto/gogo.proto:32:1: Import "google/protobuf/descriptor.proto" was not found or had errors.                                                                                                                                                                
github.com/gogo/protobuf/gogoproto/gogo.proto:38:8: "google.protobuf.EnumOptions" is not defined.                                                                                                                                                                                         
github.com/gogo/protobuf/gogoproto/gogo.proto: "google.protobuf.EnumOptions" is not defined.    
[... similar rows sniped...]
github.com/argoproj/argo-events/vendor/k8s.io/apimachinery/pkg/util/intstr/generated.proto:23:1: Import "github.com/gogo/protobuf/gogoproto/gogo.proto" was not found or had errors.
2023/09/25 14:07:48 protoc -I . -I /tmp/tmp.evfHAmKZF6/src -I ./vendor --gogo_out=/tmp/tmp.evfHAmKZF6/src /tmp/tmp.evfHAmKZF6/src/github.com/argoproj/argo-events/vendor/k8s.io/apimachinery/pkg/util/intstr/generated.proto
2023/09/25 14:07:48 Unable to generate protoc on k8s.io.apimachinery.pkg.util.intstr: exit status 1
Makefile:135: recipe for target 'codegen' failed
make: *** [codegen] Error                                 
```

I installed `protoc` from https://github.com/protocolbuffers/protobuf/releases. I'm using Linux so didn't use brew.

popsu:
:heavy_check_mark: added the same sorting for Sensor deployment.

I'm not sure if the Sensor spec has any place (I didn't find with quick look) where the same bug could occur. Since the bug in `EventSource` case was due to storing webhooks (and other eventsources) in a Map and then when creating volume/volumemounts for the secrets the code iterates through the map and in Go the order is arbitrary.

If there is similar place in Sensor then the testcase I added should be modified (like the one I added for `EventSource` has.

whynowy (approved):
LGTM, thank you!

PR #2814, opened 2023-09-23
fix: Concurrent access to trigger connection maps

### LINKED ISSUE #2660: Race conditions in sensor
**Describe the bug**
The argo sensor crashed right after receiving a batch of requests.

**To Reproduce**
Not reproducible, seems to be a race condition.

**Screenshots**

 - Kubernetes: v1.26.3
 - Argo Events: v1.8.0

**Config**
ArgoEvents is calls a webhook, but I think the error is more generic and not really dependent on the config
```
    template:
      http:
        method: POST
        payload:
        - dest: dataref
          src:
            dataKey: body.dataref
            dependencyName: configchange
        - dest: traceparent
          src:
            dataKey: body.traceparent
            dependencyName: configchange
            value: ""
        url: http://config-downloader:9090/
```

**Logs**
```
{"level":"info","ts":1686904703.0876606,"logger":"argo-events.sensor","caller":"http/http.go:194","msg":"Making a http request...","sensorName":"cc-configchange","triggerName":"configchange","triggerType":"HTTP","url":"http://config-downloader:9090/opt-osagavira-etc/policy.surfers.conf"}
{"level":"info","ts":1686904703.087826,"logger":"argo-events.sensor","caller":"sensor/trigger_conn.go:348","msg":"Triggering actions after receiving dependency configchange","sensorName":"cc-configchange","triggerName":"configchange","sensorName":"cc-configchange"}
{"level":"info","ts":1686904703.0881736,"logger":"argo-events.sensor","caller":"http/http.go:194","msg":"Making a http request...","sensorName":"cc-configchange","triggerName":"configchange","triggerType":"HTTP","url":"http://config-downloader:9090/opt-osagavira-etc/policy.ptc.conf"}
{"level":"info","ts":1686904703.1868587,"logger":"argo-events.sensor","caller":"sensors/listener.go:423","msg":"Successfully processed trigger 'configchange'","sensorName":"cc-configchange","triggerName":"configchange","triggerType":"HTTP","triggeredBy":["configchange"],"triggeredByEvents":["62333238313764342d333730382d343332352d396563342d663866313433653965643465"]}
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 188 [running]:
github.com/argoproj/argo-events/sensors/triggers/nats.NewNATSTrigger(0xc000682d00, 0xc00161ed00, 0x2e68632e6e65706f?, 0x2e72657473756c63?)
	/home/runner/work/argo-events/argo-events/sensors/triggers/nats/nats.go:70 +0x1d7
github.com/argoproj/argo-events/sensors.(*SensorContext).GetTrigger(0xc000b22000, {0x3b6c688?, 0xc000e30050?}, 0xc00161ed00)
	/home/runner/work/argo-events/argo-events/sensors/trigger.go:120 +0x70e
github.com/argoproj/argo-events/sensors.(*SensorContext).triggerOne(0xc000b22000, {0x3b6c688?, 0xc000e30050}, 0xc000682d00, {0xc0009e9ea0, {0x0, 0x0, 0x0}, 0x0, 0x0, ...}, ...)
	/home/runner/work/argo-events/argo-events/sensors/listener.go:383 +0x36b
github.com/argoproj/argo-events/sensors.(*SensorContext).triggerWithRateLimit(0xc000b22000, {0x3b6c688, 0xc000e30050}, 0xc000682d00, {0xc0009e9ea0, {0x0, 0x0, 0x0}, 0x0, 0x0, ...}, ...)
	/home/runner/work/argo-events/argo-events/sensors/listener.go:360 +0x165
created by github.com/argoproj/argo-events/sensors.(*SensorContext).triggerActions
	/home/runner/work/argo-events/argo-events/sensors/listener.go:350 +0x6ea
```
Complete logs attached:
[argoventscrash.txt](https://github.com/argoproj/argo-events/files/11770363/argoventscrash.txt)

gokulav137:
Fixes #2660

Refactor connection maps for trigger clients from map to concurrent safe map.

Checklist:


sensors/context.go
```diff
@@ -51,25 +52,25 @@ type SensorContext struct {
 	hostname        string
 
 	// httpClients holds the reference to HTTP clients for HTTP triggers.
-	httpClients map[string]*http.Client
```
whynowy, line 54: Define a generic struct in the utilities like below?

```golang
type StringKeyedMap[T any] struct {
	items map[string]T
	lock     *sync.RWMutex
}
```

gokulav137: Cool. I'll make the changes. Thanks.

common/util.go
```diff
@@ -475,3 +476,35 @@ func StructToMap(obj interface{}, output map[string]interface{}) error {
 
 	return json.Unmarshal(data, &output) // Convert to a map
 }
+
+// Concurrent Safe String keyed map
```
whynowy, line 480: Moving it to a dedicated file would be better, all the rest looks good to me. Thanks again!

gokulav137: Where do you think this would fit. Kinda puzzled.

gokulav137: I'm thinking same package but, in a concurrent.go file?

whynowy: Same package, `string_keyed_map.go`?

gokulav137: I was thinking along the lines any common concurrent safe implementations required moving forward could be placed in concurrent.go. If this doesn't track, I will change the filename. Let me know.

whynowy: I think the word `concurrent` is too big.

whynowy: This is even bigger...

It is good to use straightforward file names.

common/concurrent.go
```diff
@@ -0,0 +1,35 @@
...
+
+import "sync"
+
+// Concurrent Safe String keyed map
+type StringKeyedMap[T any] struct {
+	items map[string]T
+	lock  *sync.RWMutex
+}
+
+func NewStringKeyedMap[T any]() *StringKeyedMap[T] {
```
whynowy, line 11: There's no need to return a pointer, right?

common/util.go
gokulav137: Oh big as in vague. `concur_safe_type.go`? Sorry if I'm dragging this naming out 😅.

common/concurrent.go
gokulav137: Makes sense. I'll make it return by value.

common/util.go
gokulav137: Its just that the other files in `common/` are `json.go` , `retry.go`, `string.go`, `time.go` etc are generic names

whynowy: My personal opinions:

1. The package structure in Argo Events (including file names) should be refactored;
2. Those file names (json.go, string.go ...) are kind of simple, because they provide that particular file name related utility functions (**not struct functions**), and maybe there's no better names to describe it - you can imagine we could name them something like `json_util.go` or `string_util.go`, that wouldn't make too much difference. But it's for sure it does not make much sense to separate them into different files.
3. If it's util structs (like the one you added), there is already a clear boundary (the struct itself), putting them in separated files and naming them as straightforward as possible might be better, so that ppl clearly know what they are about.

gokulav137: This makes a lot of sense. Thanks for clarifying!

whynowy (approved):
LGTM, thanks!

PR #2831, opened 2023-09-29
Added capability to return generated eventid if an event has been dispatched after applying filters in the dispatch method

vinsri-dev:
Added logic in the dispatch process to return the generated eventid by the eventing process, whether if the event has been dispatched or not post applying filters. If the event is not dispatched post applying filters, the event id is empty.

This flag helps to handle certain aspects like to log if the event has been dispatched.

Added log function in the sqs to capture the message attributes only if the event is dispatched. This logging helps to correlate the eventid generated in eventbus with the upstream message attributes for end to end correlation and tracing
Checklist:


vinsri-dev:
@VaibhavPage , @whynowy : Could you please review and let me know your thoughts.

Thanks

vinsri-dev:
@VaibhavPage @whynowy 
Can you please help me comment if i missed anything for this PR. Could you please help progress.

whynowy:
Do we really need this? There's an option in the `dispatch()` function, to `WithID()`, which is used for tracking. See the Kafka example.


https://github.com/argoproj/argo-events/blob/master/eventsources/sources/kafka/start.go#L247

vinsri-dev:
@whynowy : Sorry by an accident i deleted my forked repository.

Anyways, i can resume it.
Still i think this code change might make sense for below reasons.

The idea is not only to get the event id, but also to be considered only if dispatched.

If the filter condition is not met (https://github.com/argoproj/argo-events/blob/master/eventsources/eventing.go#L533), we dont want to consider the event as dispatched since it is not dispatched to eventbus.

Please let me know your thoughts

PR #2850, opened 2023-10-09
fix: Replace antonmedv/expr with expr-lang/expr

terrytangyuan:
The current link is down and the author has created a fork in expr-lang. We use a replace here since the original repo would likely recover once GitHub handles the issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -330,3 +330,5 @@ require (
 	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
+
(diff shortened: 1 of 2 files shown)


gdsoumya (approved):

whynowy:
@terrytangyuan - what happened with the original link?

terrytangyuan:
See https://x.com/antonmedv/status/1711295286562189395?s=20

terrytangyuan:
The author [confirmed](https://x.com/antonmedv/status/1711448876677280051?s=20) so we should be good to proceed

whynowy (approved):

terrytangyuan:
@whynowy would you like to merge this?

PR #2867, opened 2023-10-16
Added capability to return generated eventid if an event has been dispatched after applying filters in the dispatch method

vinsri-dev:
Checklist:


vinsri-dev:
@whynowy 

This a follow up to the PR : https://github.com/argoproj/argo-events/pull/2831
Closed the 2831 PR by an accident.

Based on your comments in 2831 PR 
**"
@whynowy :
Do we really need this? There's an option in the dispatch() function, to WithID(), which is used for tracking. See the Kafka example.

https://github.com/argoproj/argo-events/blob/master/eventsources/sources/kafka/start.go#L247
"**


Still i think this code change might make sense for below reasons.

The idea is not only to get the event id, but also to be considered only if dispatched.

If the filter condition is not met (https://github.com/argoproj/argo-events/blob/master/eventsources/eventing.go#L533), we dont want to consider the event as dispatched since it is not dispatched to eventbus.

Please let me know your thoughts

PR #2921, opened 2023-11-26
feat(bitbucketserver): multiple improvements to bitbucketserver event source

ryancurrah:
There are no breaking changes in the pull request, the event source will behave the same by default. One bigger internal change is that the Bitbucket Server client is now created once, set on the Route struct.

---

This Pull Request contains 3 improvements to the Bitbucket Server Event source.

1. Customizable Apply Webhook Check Interval

Currently the apply webhook check runs every 60 seconds which could be to often. This is now configurable.

2. Manage Webhooks for Entire Projects

When a list of Bitbucket Server Projects are provided the event source will ensure a webhook exists in all repositories in the provided project.

3. Ability to skip `Refs Changed` event type when a Pull Request is Open for the same commit.

This feature aims to reduce redundant workflow triggers caused by Bitbucket's event handling. Specifically, when a Pull Request is opened and a new commit is pushed, Bitbucket sends two events: `repo:refs_changed` and either `pr:from_ref_created` (for a new PR) or `pr:from_ref_updated` (for an existing PR). This results in multiple workflows being launched for both Branch and Pull Request events.

To address this, the Pull Request adds a boolean option named `skipBranchRefsChangedOnOpenPR`. This option, when enabled, performs a check to determine if a Pull Request is open for a specific branch commit associated with a `repo:refs_changed` event. If a PR is open for that commit, the new feature will skip sending the `repo:refs_changed` event. This effectively prevents the triggering of multiple workflows for the same commit in the scenario where a Pull Request is opened.


pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -1059,46 +1059,54 @@ type BitbucketBasicAuth struct {
...
 	// +optional
 	DeprecatedRepositorySlug string `json:"repositorySlug,omitempty" protobuf:"bytes,3,opt,name=repositorySlug"`
-	// Repositories holds a list of repositories for which integration needs to set up
+	// Projects holds a list of projects for which integration needs to set up, this will add the webhook to all repositories in the project.
+	// +optional
+	Projects []string `json:"projects,omitempty" protobuf:"bytes,4,opt,name=projects"`
+	// Repositories holds a list of repositories for which integration needs to set up.
 	// +optional
-	Repositories []BitbucketServerRepository `json:"repositories,omitempty" protobuf:"bytes,4,rep,name=repositories"`
+	Repositories []BitbucketServerRepository `json:"repositories,omitempty" protobuf:"bytes,5,opt,name=repositories"`
```
whynowy, line 1077: It should be `rep`.

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -1059,46 +1059,54 @@ type BitbucketBasicAuth struct {
...
 	// Refer https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html
-	Events []string `json:"events" protobuf:"bytes,5,opt,name=events"`
-	// AccessToken is reference to K8s secret which holds the bitbucket api access information
-	AccessToken *corev1.SecretKeySelector `json:"accessToken,omitempty" protobuf:"bytes,6,opt,name=accessToken"`
-	// WebhookSecret is reference to K8s secret which holds the bitbucket webhook secret (for HMAC validation)
-	WebhookSecret *corev1.SecretKeySelector `json:"webhookSecret,omitempty" protobuf:"bytes,7,opt,name=webhookSecret"`
-	// BitbucketServerBaseURL is the base URL for API requests to a custom endpoint
-	BitbucketServerBaseURL string `json:"bitbucketserverBaseURL" protobuf:"bytes,8,opt,name=bitbucketserverBaseURL"`
+	// +optional
+	Events []string `json:"events" protobuf:"bytes,6,opt,name=events"`
```
whynowy, line 1081: This should be `rep` as well.

whynowy (commented):
@daniel-codefresh - could you take a look?

pkg/apis/eventsource/v1alpha1/types.go
ryancurrah: Should `Projects []string `json:"projects,omitempty" protobuf:"bytes,4,opt,name=projects"` also be `rep` as well since it is a list of values as well?

ryancurrah: Done ✅

pkg/apis/eventsource/v1alpha1/types.go
ryancurrah: Done ✅

ryancurrah:
Fixed incorrect protobuf struct tag options https://github.com/argoproj/argo-events/pull/2921/commits/37958b87f5bee8e2aa6f817ba62882d8f07610b5. 

Though these types should be auto generated it makes modifications error prone.

ryancurrah:
@daniel-codefresh this is ready for review.

ryancurrah:
Hey @daniel-codefresh I hope you had a great holidays and happy new year. I am wondering if you would be able to review this?

ryancurrah:
Is anyone available to review this @whynowy?

eventsources/sources/bitbucketserver/start.go
```diff
@@ -191,9 +214,37 @@ func (el *EventListener) StartListening(ctx context.Context, dispatch func([]byt
...
+
+	bitbucketURL, err := url.Parse(bitbucketserverEventSource.BitbucketServerBaseURL)
+	if err != nil {
+		return fmt.Errorf("parsing bitbucketserver url. err: %w", err)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	bitbucketClient := newBitbucketServerClient(ctx, bitbucketConfig, bitbucketToken)
```
whynowy, line 237: The `ctx` will be cancelled when this function exits, will the `bitbucketConfig` still be able to used for delete the webhook?

ryancurrah: Good catch. I created a second client that will be used by `PostInactivate`. I also ensured that the `customClient` uses the `ctx`.

pkg/apis/eventsource/v1alpha1/webhook_context.go
```diff
@@ -32,6 +32,9 @@ type WebhookContext struct {
 	// Default value: 1048576 (1MB).
 	// +optional
 	MaxPayloadSize *int64 `json:"maxPayloadSize,omitempty" protobuf:"bytes,9,opt,name=maxPayloadSize"`
+	// CheckInterval is a duration in which to wait before checking that the webhooks exist, e.g. 1s, 30m, 2h... (defaults to 1m)
+	// +optional
+	CheckInterval string `json:"checkInterval" protobuf:"bytes,10,opt,name=checkInterval"`
```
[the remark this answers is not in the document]
ryancurrah: I tried it but I get this error when running `make codegen`.

```
The swagger spec at "api/openapi-spec/swagger.json" is invalid against swagger specification 2.0.
See errors below:
- some references could not be resolved in spec. First found: object has no key "io.k8s.apimachinery.pkg.apis.meta.v1.Duration"
```

I was copying an existing duration example from here, as I did not see any uses of a `Duration` type, probably for this reason maybe?
https://github.com/argoproj/argo-events/blob/6414f3b95588caaba5e6a3ddf72310972a2d47a6/pkg/apis/eventsource/v1alpha1/types.go#L1134

ryancurrah (commented):
Changes pushed and responded to comments. Thanks for reviewing.

ryancurrah:
@whynowy this is ready for review again.

pkg/apis/eventsource/v1alpha1/webhook_context.go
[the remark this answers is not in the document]
whynowy: We can fix this later.

Just realized it's added to the generic `WebhookContext`, which is used by many event sources. This is not right, please move to bigbuckets specific types.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -93,6 +98,40 @@ func (router *Router) HandleRoute(writer http.ResponseWriter, request *http.Requ
 		return
 	}
 
+	// When SkipBranchRefsChangedOnOpenPR is enabled and webhook event type is repo:refs_changed,
```
whynowy, line 101: Why is there so specific logic hardcoded?

If there's a requirement like you mentioned in the PR description, instead of doing this, you can use filter feature in the Sensor take the events you are interested.

eventsources/sources/bitbucketserver/start.go
whynowy: In this case, I don't think this is a good approach, maybe just extract the way to get a client to a function?

pkg/apis/eventsource/v1alpha1/webhook_context.go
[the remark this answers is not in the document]
ryancurrah: Ack, I will move this.

eventsources/sources/bitbucketserver/start.go
ryancurrah: I met with Pipekit folks and because of the API call required it does not seem possible. With auth and an adding a self-signed ca certificate. This addition goes a long way for Argo to replace the multibranch pipeline in Jenkins. If you want I can do this in a separate supporting/pleading my case a little better.

eventsources/sources/bitbucketserver/start.go
ryancurrah: Your instructions aren't clear here. I will take a second look though.

ryancurrah: Looking at this again, I feel this is the right approach using a separate instance of the bitbucket client for deleting which does not use the canceled context and adding the bitbucket clients to the `Router` struct which can be accessed by the different functions is a common approach to using clients.

eventsources/sources/bitbucketserver/start.go
```diff
@@ -125,53 +164,37 @@ func (router *Router) PostInactivate() error {
...
 	logger := route.Logger
 
-	if bitbucketserverEventSource.DeleteHookOnFinish && len(router.hookIDs) > 0 {
-		logger.Info("deleting webhooks from bitbucket")
-
-		bitbucketToken, err := common.GetSecretFromVolume(bitbucketserverEventSource.AccessToken)
-		if err != nil {
-			return fmt.Errorf("failed to get bitbucketserver token. err: %w", err)
-		}
+	if !bitbucketserverEventSource.DeleteHookOnFinish && len(router.hookIDs) == 0 {
```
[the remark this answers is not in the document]
ryancurrah: This is done. ✅

whynowy (approved):
Please fix the conflict.

ryancurrah:
> Please fix the conflict.

Done and also squashed commits for a clean merge.

ryancurrah:
@whynowy hoping this gets merged before another conflict in go.mod.

whynowy:
> > Please fix the conflict.
> 
> Done and also squashed commits for a clean merge.

Sorry, you have to do it again... Ping me on slack after you do it.

tooptoop4:
when dis bad boy gonna be released?

Issue #2928, opened 2023-12-08
Retry not occurring when trigger fails


bencompton:
**Describe the bug**

When retry trigger configuration is used and the trigger fails (e.g., HTTP trigger receives 500 response), the trigger is not retried.

**To Reproduce**

Create a HTTP trigger and force the HTTP endpoint to return an error response (e.g., HTTP 500).

```
    - template:
        name: failing-trigger
        http:
          # Expect to fail
          url: https://failing-http-response.com/
          method: GET
      atLeastOnce: true
      retryStrategy:
        steps: 3
```

**Additional context**

Looking at the code [here](https://github.com/intuit-data-os/argo-events/blob/9def50653c0bee565619b084c80ac21d0d12443f/sensors/listener.go#L335), it doesn't appear that the result of the trigger is taken into account, only that it completes.

ThuWangzw:
Same problem and I think it was caused by [this code](https://github.com/intuit-data-os/argo-events/blob/9def50653c0bee565619b084c80ac21d0d12443f/sensors/listener.go#L401):
```go
	if err := common.DoWithRetry(retryStrategy, func() error {
		var e error
		newObj, e = triggerImpl.Execute(ctx, eventsMapping, updatedObj)
		return e
	}); err != nil {
		return fmt.Errorf("failed to execute trigger, %w", err)
	}
	logger.Debug("trigger resource successfully executed")

	logger.Debug("applying trigger policy")
	if err := triggerImpl.ApplyPolicy(ctx, newObj); err != nil {
		return err
	}
```
the sensor called the http trigger's `Execute` method but didn't check its status code. So the `DoWithRetry` function just execute the `triggerImpl.Execute` for once. I think the `triggerImpl.ApplyPolicy` should be moved into the `DoWithRetry` function

PR #2956, opened 2023-12-28
fix: fix trigger retryStrategy

ThuWangzw:
Checklist:

* [] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).

Fix the issue [#2928](https://github.com/argoproj/argo-events/issues/2928) that the trigger wasn't correctly retried as defined in the retryStrategy.

### DIFF
--- a/sensors/listener.go
+++ b/sensors/listener.go
@@ -361,7 +361,7 @@ func (sensorCtx *SensorContext) triggerActions(ctx context.Context, sensor *v1al
 			if err != nil {
 				// Log the error, and let it continue
 				logger := logging.FromContext(ctx)
-				logger.Warnw("Failed to execute a trigger", zap.Error(err), zap.String(logging.LabelTriggerName, trigger.Template.Name))
+				logger.Errorw("Failed to execute a trigger", zap.Error(err), zap.String(logging.LabelTriggerName, trigger.Template.Name))
 			}
 		}()
 		return nil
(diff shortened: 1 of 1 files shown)


sensors/listener.go
```diff
@@ -211,7 +211,16 @@ func (sensorCtx *SensorContext) listenEvents(ctx context.Context) error {
 			}
 
 			actionFunc := func(events map[string]cloudevents.Event) {
-				sensorCtx.triggerActions(ctx, sensor, events, trigger)
+				retryStrategy := trigger.RetryStrategy
+				if retryStrategy == nil {
+					retryStrategy = &apicommon.Backoff{Steps: 1}
```
terrytangyuan, line 216: We should probably make this configurable

sensors/listener.go
```diff
@@ -345,13 +354,21 @@ func (sensorCtx *SensorContext) triggerActions(ctx context.Context, sensor *v1al
...
-		sensorCtx.triggerWithRateLimit(ctx, sensor, trigger, eventsMapping, depNames, eventIDs)
+		return sensorCtx.triggerWithRateLimit(ctx, sensor, trigger, eventsMapping, depNames, eventIDs)
 	} else {
-		go sensorCtx.triggerWithRateLimit(ctx, sensor, trigger, eventsMapping, depNames, eventIDs)
+		go func() {
+			err := sensorCtx.triggerWithRateLimit(ctx, sensor, trigger, eventsMapping, depNames, eventIDs)
+			if err != nil {
+				// Log the error, and let it continue
+				logger := logging.FromContext(ctx)
+				logger.Warnw("Failed to execute a trigger", zap.Error(err), zap.String(logging.LabelTriggerName, trigger.Template.Name))
```
terrytangyuan, line 364: logger.error instead of warn?

sensors/listener.go
```diff
@@ -362,8 +379,10 @@ func (sensorCtx *SensorContext) triggerWithRateLimit(ctx context.Context, sensor
 		log.Errorw("Failed to execute a trigger", zap.Error(err), zap.String(logging.LabelTriggerName, trigger.Template.Name),
 			zap.Any("triggeredBy", depNames), zap.Any("triggeredByEvents", eventIDs))
 		sensorCtx.metrics.ActionFailed(sensor.Name, trigger.Template.Name)
+		return err
 	} else {
 		sensorCtx.metrics.ActionTriggered(sensor.Name, trigger.Template.Name)
+		return nil
```
terrytangyuan, line 385: We can remove the else block and unindent these lines

sensors/listener.go
whynowy: There is a default backoff retry strategy.

sensors/listener.go
ThuWangzw: fixed 😄

whynowy (approved):
Thanks!

Issue #2958, opened 2023-12-29
bug: kafka sensors stop working quietly


AmitMendl:
**Describe the bug**
Sensors connected to a kafka eventbus stop working some while after being deployed.

**To Reproduce**
1. Create a kafka eventbus
2. Create a basic HTTP webhook 
3. Create a kafka sensor (doesn't matter what it triggers)
4. Wait a couple of weeks
5. Sensor will stop receiving events

**Expected behavior**
The sensor should receive the events, or at least the pod

**Environment (please complete the following information):**
 - Kubernetes: v1.28
 - Argo Events: v1.8.0
 - Streamzi: v0.39.0

**Screenshots**
I unfortunelty cannot supply any screenshots or logs, since this happened in an offline environment, I will try to describe the logs.
 - eventsource logs indicated that the event was created
 - akhq into kafka showed that events were created in the kafka topic, but at some point in the middle of the topic they aren't being consumed.
 - logs on the sensor pod show the start of a kafka transaction but no the end of it.

**Additional context**
 - This does not happen instantly, and usually follows dozens of days that the sensor works fine
 - When investigating this, I have verified that the event is being created, and I could see the message being added to eventbus's topic
 - this happens seemingly at random, some sensors will fail after 2 weeks, some after a month, some are yet to fail.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

AmitMendl:
After looking at the source code I suspect that this is caused by the producer's Errors() channel not being read, I created a pull requested that I believe that fixes this, https://github.com/argoproj/argo-events/pull/2959

James-Derune:
We have also experienced this behavior. Sensors stop writing logs and no longer process events from Kafka, despite the events being created properly there (there is an offset lag)

We run using OpenShift and this happens both on version 4.10 and 4.12 of OCP (k8s versions are 1.23.5 and 1.25.14).

piby180:
I still experience this with the latest version (1.9.2). Sensors stop processing messages after a couple of days. They have to be restarted in order to make them work again

PR #2959, opened 2023-12-29
fix: added Errors() read from kafkaSensor's producer

AmitMendl:
Checklist:

Fix issue https://github.com/argoproj/argo-events/issues/2958, added procuder.Errors() read and log to stop kafka sensor's producer from deadlocking.


whynowy (approved):
thanks!

Issue #2672, opened 2023-06-22
Sensor - systematic "Event ... discarded due to filtering" warning log


nicolas-vivot:
**Describe the bug**
Not sure it is a bug or on purpose.

In the scenario where we have a sensor with multiple triggers, each having their own filter(s) but all watching for the same event source, when an event is received and processed successfully by one of the triggers, the other ones systematically log as a warning a message like "Event [...] discarded due to filtering"

This is a little annoying since yes at the trigger level it has been discarded, but the incoming event was actually correctly processed by another trigger.
I would actually expect such a warning to be logged only if none of the triggers could process the event it subscribes to - or said differently if all triggers discard the event.

I am using Datadog + some alert monitors based on that log message to be alerted when some incoming payload could not be processed because none of the trigger / filter could accept it.
This was working fine when the sensor had just one trigger, but i recently added more to be able to trigger different workflow based on the event payload.

This behavior is really miss-leading since it is logged as a warning.
I would like to know first if there is anything wrong that i did, or a better way to achieve my use case that would allow not to trigger that kind of warning log.
But i would like to continue to have a way to detect that an incoming event could not be processed / was ignored.
Does Argo Event provide something at the sensor level, in the filters or triggers to actually have a default behavior when none of the triggers could apply ?

**To Reproduce**
Here is an example of sensor with 3 different triggers / filters, each to trigger a specific workflow

```
apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: example-sensor
  labels:
    app: example-sensor
spec:
  replicas: 1
  # ... any other definitions in between
  dependencies:
    - name: workflow-case-1
      eventSourceName: some-source
      eventName: some-event
      filters:
        data:
          - path: body.workflowName
            type: string
            value:
              - "WORKFLOW_CASE_1"
    - name: workflow-case-2
      eventSourceName: some-source
      eventName: some-event
      filters:
        data:
          - path: body.workflowName
            type: string
            value:
              - "WORKFLOW_CASE_2"
    - name: workflow-case-3
      eventSourceName: some-source
      eventName: some-event
      filters:
        data:
          - path: body.workflowName
            type: string
            value:
              - "WORKFLOW_CASE_3"


  triggers:

    - template:
        conditions: workflow-case-1
        name: workflow-1
        argoWorkflow:
          operation: submit
          # ... skip workflow skeleton as it is out of scope

    - template:
        conditions: workflow-case-2
        name: workflow-2
[...]

PR #2964, opened 2024-01-04
fix: Reduce Sensor log verbosity for filtered events

bencompton:
Changing log level from `WARN` to `DEBUG` when events do not match a Sensor filter, as suggested by @whynowy in this [issue comment](https://github.com/argoproj/argo-events/issues/2672#issuecomment-1630131794). This represented ~400GB of logs per day in my cluster at scale, and logging the entire payload for discarded events is only useful for occasional debugging purposes.


jmeridth (approved):

whynowy (approved):
Thanks!

PR #2977, opened 2024-01-13
Feat/recursive file events

AmitMendl:
Checklist:

Add FileEventSource support for recursive file events


AmitMendl:
I cant figure out why but for some reason `make codegen` produced results that fail the codegen test, i'd use codegen locally, and it'd add blank lines to the api docs' md files for some reason I cant quite figure out, I am assuming its something problamatic with my local setup since the html files pass the test, so I removed the blank lines manually

Makefile
```diff
@@ -16,8 +16,8 @@ EXECUTABLES = curl docker gzip go
 #  docker image publishing options
 DOCKER_PUSH?=false
 IMAGE_NAMESPACE?=quay.io/argoproj
-VERSION?=latest
-BASE_VERSION:=latest
+VERSION?=v1.9.0
```
whynowy, line 19: Should not change these.

eventsources/sources/file/start.go
```diff
@@ -138,6 +158,32 @@ func (el *EventListener) listenEvents(ctx context.Context, dispatch func([]byte,
 				// watcher stopped watching file events
 				return fmt.Errorf("fs watcher stopped for %s", el.GetEventName())
 			}
+			if (fileEventSource.WatchPathConfig.Recursive) && (event.Has(fsnotify.Create)) {
+				fileinfo, err := os.Stat(event.Name)
+				if err != nil {
+					log.Errorw("failed getting filestat", zap.Error(err))
+				}
+				if fileinfo.IsDir() {
+					err = watcher.Add(event.Name)
```
whynowy, line 167: 1. If there is a dir create with sub directories (folder A contains B, B contains C), how many events are supposed to get? will it have duplicated processing?
2. No eviction mechanism for deleting?

Issue #2993, opened 2024-01-25
TestMetricsWithWebhook fails


BabisK:
**Describe the bug**
When executing the e2e test `TestMetricsWithWebhook` a assertion fails. The reason is that the test expects the URL https://not-existing.com/ to not respond. But it seems that this URL now responds with 404.

```
2024-01-25T20:54:40.342Z	INFO	argo-events.sensor	sensors/listener.go:433	Successfully processed trigger 'http-trigger'	{"sensorName": "e2e-test-metrics", "triggerName": "http-trigger", "triggerType": "HTTP", "triggeredBy": ["test-dep"], "triggeredByEvents": ["62666466616533632d306231382d343331392d623135362d323966353734303239343662"]}
error: context deadline exceeded    then.go:79: expected sensor pod log contains 'Failed to execute a trigger'
    then.go:129: Terminating port-forward for POD e2e-test-metrics-sensor-lnwt8-7546655b5f-c94fj
    panic.go:522: Deleting sensor e2e-test-metrics
    then.go:129: Terminating port-forward for POD e2e-test-metrics-webhook-eventsource-t7m6h-85f8779d8-dlpxh
    panic.go:522: Deleting event source e2e-test-metrics-webhook
    --- FAIL: TestFunctionalSuite/TestMetricsWithWebhook (21.43s)
```

**To Reproduce**
1. Setup the test environment with k3d
2. `EventBusDriver=stan make TestMetricsWithWebhook`

**Expected behavior**
The URL should not respond and the test would secceed.

**Environment (please complete the following information):**
 - Kubernetes: [k3d]
 - Argo Events: [master]

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2994, opened 2024-01-25
Replace url https://not-existing.com/ with https://invalid.com/ (#2993)

BabisK:
This url is reserved by IANA and is expected to always fail resolution.

Checklist:


whynowy (approved):
Thanks!

PR #2996, opened 2024-01-26
feat: add waitTime config field to azure queue storage event source

eduardodbr:
Checklist:

Azure queue storage eventsource is making the `DequeueMessage` API call in a loop, which might create a lot of requests per second if the message queue is empty. Since the azure SDK does not allow for waiting for a message when the queue is empty like AWS SQS does, this PR adds a new config field that allow to configure how much time it should wait (sleep) between empty results from the queue. It defaults to 3 seconds.


eduardodbr:
CI failing because of #2993

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -1299,6 +1299,10 @@ type AzureQueueStorageEventSource struct {
 	// If set to true the decoding is done before the evaluation of JSONBody
 	// +optional
 	DecodeMessage bool `json:"decodeMessage,omitempty" protobuf:"bytes,8,opt,name=decodeMessage"`
+	// WaitTime is the duration (in seconds) for which the event source waits between empty results from the queue.
+	// The default value is 3 seconds.
+	// +optional
+	WaitTime *int32 `json:"waitTime,omitempty" protobuf:"varint,9,opt,name=waitTime"`
```
whynowy, line 1305: `waitTimeInSeconds` might be better.

eduardodbr: done

whynowy (approved):
LGTM

PR #3004, opened 2024-02-02
fix: do not overwrite the entrypoint of nats-server-config-reloader

yulin-li:
Checklist:

To support `distroless` based `nats-server-config-reloader` docker image

The `natsio/nats-server-config-reloader` image should be >= `0.9.0` now


yulin-li:
> You need to sign off on your commits. Running `git commit -s --amend` should do it, then do a `git push --force-with-lease`.
> 
> ```
>  -s, --signoff, --no-signoff
>            Add a Signed-off-by trailer by the committer at the end of the commit log message. The meaning of a signoff depends on the project to which you’re
>            committing. For example, it may certify that the committer has the rights to submit the work under the project’s license or agrees to some contributor
>            representation, such as a Developer Certificate of Origin. (See http://developercertificate.org for the one used by the Linux kernel and Git projects.)
>            Consult the documentation or leadership of the project to which you’re contributing to understand how the signoffs are used in that project.
> 
>            The --no-signoff option can be used to countermand an earlier --signoff option on the command line.
> ```

Thanks for reminder, I signed the commit off

manifests/namespace-install.yaml
```diff
@@ -254,42 +254,42 @@ data:
...
         - version: 2.9.12
           natsImage: nats:2.9.12
           metricsExporterImage: natsio/prometheus-nats-exporter:0.9.1
-          configReloaderImage: natsio/nats-server-config-reloader:0.7.0
+          configReloaderImage: natsio/nats-server-config-reloader:0.14.1
           startCommand: /nats-server
         - version: 2.9.16
           natsImage: nats:2.9.16
           metricsExporterImage: natsio/prometheus-nats-exporter:0.9.1
-          configReloaderImage: natsio/nats-server-config-reloader:0.7.0
```
whynowy, line 292: Instead updating old versions, maybe add the latest nats version with the new config reloader image? This config is always supposed to be maintained by the users.

yulin-li: sure, `latest` makes sense

whynowy: What I meant was, still add the latest nats image `2.10.x`, and also update the other images like `config-reloader` with new version, but only for `2.10.x` config, and remain the histories unchanged.

Makefile
```diff
@@ -151,7 +151,7 @@ docs/assets/diagram.png: go-diagrams/diagram.dot
 .PHONY: start
 start: image
 	kubectl apply -f test/manifests/argo-events-ns.yaml
-	kubectl kustomize test/manifests | sed 's@quay.io/argoproj/@$(IMAGE_NAMESPACE)/@' | sed 's/:$(BASE_VERSION)/:$(VERSION)/' | kubectl -n argo-events apply -l app.kubernetes.io/part-of=argo-events --prune=false --force -f -
+	kubectl kustomize test/manifests | sed 's@quay.io/argoproj/@$(IMAGE_NAMESPACE)/@' | sed 's/argo-events:$(BASE_VERSION)/argo-events:$(VERSION)/' | kubectl -n argo-events apply -l app.kubernetes.io/part-of=argo-events --prune=false --force -f -
```
whynowy, line 154: I understand this will make it more accurate, just curious, did you see any issue caused by not having `argo-events` in the script?

yulin-li: yes, the `natsio/nats-server-config-reloader:latest` was replaced to `natsio/nats-server-config-reloader:<id>` in the tests, then the tests failed.

manifests/namespace-install.yaml
yulin-li: thanks, updated.

whynowy (approved):
Thanks

PR #3113, opened 2024-04-22
Use authorized key parsing instead of reading binary certificates

florian-feicht:
Checklist:

The current implementation of the public key parsing a binary public key is needed (https://pkg.go.dev/golang.org/x/crypto/ssh#ParsePublicKey).
I would propose to switch to https://pkg.go.dev/golang.org/x/crypto/ssh#ParseAuthorizedKey as it supports a common string format, which is a lot easier to handle in secrets.

Example in  go playground:
https://go.dev/play/p/doCqwWh6jpl


pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -303,7 +303,7 @@ type SFTPEventSource struct {
 	Username *corev1.SecretKeySelector `json:"username,omitempty" protobuf:"bytes,3,opt,name=username"`
 	// Password required for authentication if any.
 	Password *corev1.SecretKeySelector `json:"password,omitempty" protobuf:"bytes,4,opt,name=password"`
-	// SSHKeySecret refers to the secret that contains SSH key
+	// SSHKeySecret refers to the secret that contains SSH key. Key needs to contain private key and public key.
```
whynowy, line 306: It changes to require both private and public keys?

florian-feicht: Actually the implementation required private and public key already. Only adjusted the comment to make it clear.

PR #3131, opened 2024-05-13
Optional use QueueSubscribe for NATS events source

### LINKED ISSUE #2936: Use `QueueSubscribe` for NATS events source?
Right now NATS event source uses `subscribe` and because of that the has to use the [Active/Passive strategy](https://argoproj.github.io/argo-events/eventsources/ha/#active-passive) for HA. 

Could we not switch to using [Queue Subscriptions](https://docs.nats.io/using-nats/developer/receiving/queues) giving us the ability to have Active/Active HA?

---

sysr9:
resolves #2936 

Checklist:

adds queue parameter to NATS EventSource to implement queue-based architecture


praveenperera:
Wouldn't it make more sense to make this the default, and switch the active-active failover?

praveenperera:
@sysr9 makes sense, if we are using the QueueGroup how can we make it use active-active HA mode? Or should that be a different PR?

pkg/apis/eventsource/v1alpha1/types.go
```diff
@@ -631,6 +631,10 @@ type NATSEventsSource struct {
 	// Filter
 	// +optional
 	Filter *EventSourceFilter `json:"filter,omitempty" protobuf:"bytes,8,opt,name=filter"`
+	// Queue is the name of the queue group to subscribe as if specified. Uses QueueSubscribe
+	// logic to subscribe as queue group. If the queue is empty, uses default Subscribe logic.
+	// +optional
+	Queue string `json:"queue" protobuf:"bytes,9,opt,name=queue"`
```
whynowy, line 637: Use `*string`, all the rest looks good to me, thanks!

whynowy (approved):
LGTM

PR #3147, opened 2024-05-28
Add more EventSource event logs for failed to publish events

aaron-arellano:
Checklist:

Adds a debug log in the EventSource which logs the incoming event body with the CloudEvents event ID for better correlation in the EventSource. Also adds the CloudEvents EventID to the `Failed to publish an event` log so we can actually know what event failed to publish to the EventBus. This ID was not available in that log before and when a failure occurred where the event failed to publish to the EventBus we were unable to pinpoint exactly what event failed to publish. This PR is meant to add more transparency to event failures in the EventSource so we have the ability to see what failed, similar to how we can debug the event in the Trigger.


whynowy (approved):
LGTM, thanks for adding your org to the user list!

PR #3173, opened 2024-06-19
fix: update kubernetes labels to comply with recommended labels

### LINKED ISSUE #3172: Kustomizations don't comply with kubernetes recommended labels
**Describe the bug**
The argo-events manifests use the old style of labels of `app` which is now preferred to be set to `app.kubernetes.io/name` per the [Kubernetes label recommendations](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).

 - Kubernetes: v1.26.14
 - Argo: v2.9.4
 - Argo Events: v1.9.3

**Additional context**

---

therealdwright:
This commit updates all docs, manifests and tests to comply with the
official [Kubernetes Recommended Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

Signed-off-by: Daniel Wright <dan@dwright.io>

Closes #3172

Checklist:

### DIFF
--- a/docs/eventsources/setup/amqp.md
+++ b/docs/eventsources/setup/amqp.md
@@ -44,28 +44,28 @@ The structure of an event dispatched by the event-source over the eventbus looks
         kind: Service
         metadata:
           labels:
-            component: rabbitmq
+            app.kubernetes.io/component: rabbitmq
           name: rabbitmq-service
         spec:
           ports:
             - port: 5672
           selector:
-            app: taskQueue
-            component: rabbitmq
+            app.kubernetes.io/name: taskQueue
+            app.kubernetes.io/component: rabbitmq
         ---
         apiVersion: v1
         kind: ReplicationController
         metadata:
           labels:
-            component: rabbitmq
+            app.kubernetes.io/component: rabbitmq
           name: rabbitmq-controller
         spec:
           replicas: 1
           template:
             metadata:
               labels:
-                app: taskQueue
-                component: rabbitmq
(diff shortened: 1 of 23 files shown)


therealdwright:
> could you test and comment what will be the behaviors and anything needs to do?

Yeah - we have forked the manifests are running this version of the labels and everything works. You are correct that you will either need to run some variant of `kubectl replace` or `kubectl apply --force`

whynowy:
I'm hesitating about this change - definitely it's reasonable, even though it requires a force update. 

I'm thinking that we also should add recommended labels to the orchestrated resources - the deployments generated for event sources and sensors.

therealdwright:
> I'm thinking that we also should add recommended labels to the orchestrated resources - the deployments generated for event sources and sensors.

I will have to spend a bit more time understanding the code if this is something you really want me to do as part of this PR.

Really the goal here was to make the world a better place as the `app` and `k8s-app` are pretty legacy at this stage, we maintain our own manifests to avoid using these legacy labels.

whynowy:
> > I'm thinking that we also should add recommended labels to the orchestrated resources - the deployments generated for event sources and sensors.
> 
> I will have to spend a bit more time understanding the code if this is something you really want me to do as part of this PR.
> 
> Really the goal here was to make the world a better place as the `app` and `k8s-app` are pretty legacy at this stage, we maintain our own manifests to avoid using these legacy labels.

Appreciate if you would like to spend time on it, let me know!

Issue #1241, opened 2021-06-07
Support IAM authentication to AWS Event Sources 


saranyaeu2987:
**Is your feature request related to a problem? Please describe.**
Currently AWS event sources like S3, SNS etc support authentication using access key and secret only.
example: https://github.com/argoproj/argo-events/blob/master/examples/event-sources/minio.yaml

Please support IAM authentication using **useSDKCreds** like in Argo workflows

**Describe the solution you'd like**
Support  IAM authentication using **useSDKCreds**   like in argo workflows https://argoproj.github.io/argo-workflows/fields/?query=useSDKCreds

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
@saranyaeu2987 - if access key and secret are not specified in the spec, SNS event source use the default approach to create a session to AWS, which covers the IAM one. I assume this request is only for minio.

tooptoop4:
this is about IAM role support?

PR #3192, opened 2024-07-04
Support IAM authentication to AWS Event Sources #1241

nikzayn:
Checklist:

### DIFF
--- a/examples/event-sources/minio.yaml
+++ b/examples/event-sources/minio.yaml
@@ -18,6 +18,8 @@ spec:
         - s3:ObjectRemoved:Delete
       # type of the connection
       insecure: true
+      # refers to IAM authentication
+      useSDKCreds: true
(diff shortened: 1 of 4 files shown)


whynowy:
Thanks for adding your firm to USERS! Curious what `1241` means?

nikzayn:
> Thanks for adding your firm to USERS! Curious about what `1241` means?

My bad I have updated the title.

examples/event-sources/minio.yaml
```diff
@@ -18,6 +18,8 @@ spec:
         - s3:ObjectRemoved:Delete
       # type of the connection
       insecure: true
+      # refers to IAM authentication
+      useSDKCreds: true
```
whynowy, line 22: I don't think we support this at this moment.

Issue #3065, opened 2024-03-12
Using git as a source for k8s trigger: git executable not found


bkanuka:
Using `git` as a source for `k8s` trigger fails with error: "failed to fetch artifact, failed after retries: failed to fetch. err: exec: \"git\": executable file not found"

I'm using the following Sensor:
```
apiVersion: argoproj.io/v1alpha1
kind: Sensor
metadata:
  name: github-auto-build
spec:
  template:
    container:
      env:
        - name: DEBUG_LOG
          value: "true"
      volumeMounts:
        - mountPath: /git
          name: git
    volumes:
      - name: git
        emptyDir: {}
    serviceAccountName: operate-workflow-sa
  dependencies:
    - name: github
      eventSourceName: github
      eventName: github-climateengine
      filters:
        data:
          - path: headers.X-Github-Event
            type: string
            value:
              - push
  triggers:
    - template:
        name: workflow-trigger
        k8s:
          operation: create
          source:
            git:
              url: "https://github.com/argoproj/argo-workflows.git"
              cloneDirectory: "/git/argoproj"
              namespace: argo-events
              filePath: "examples/hello-world.yaml"
              branch: "master"
```

The template fails with the message:

```
2024-03-12T21:09:14.490Z ERROR argo-events.sensor sensors/listener.go:379 Failed to execute a trigger {"sensorName": "github-auto-build", "error": "failed to fetch artifact, failed after retries: failed to fetch. err: exec: \"git\": executable file not found in $PATH", "triggerName": "build-workflow", "triggeredBy": ["github"], "triggeredByEvents": ["34383866653034612d383330372d346161312d383264302d633730643666363239613730"]}
```

I can confirm that the Sensor is running a Pod with image `quay.io/argoproj/argo-events:v1.9.1`

**Environment:**
 - Kubernetes: v1.26.13
 - Argo Workflows: v3.5.5
 - Argo Events: v1.9.1

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

bkanuka:
It looks like `git` was added to an old image in this PR:  https://github.com/argoproj/argo-events/pull/521
It's possible when moving to a single image, `git` was left out.

bkanuka:
I can confirm that this is simply a matter of the `git` binary not being on the argo-events image.   I used the following Dockerfile as the Sensor image and was able to use git sources successfully!

[...]

PR #3244, opened 2024-08-16
fix: add git to argo-events image

tdenmon:
This commit adds `git` to the argo-events Docker image, which is required in order to use Git as a source for a Sensor.

Fixes [#3065](https://github.com/argoproj/argo-events/issues/3065)

Checklist:


tdenmon (commented):
Copying the relevant linked libraries is not my ideal solution here, but as `git` is dynamically linked in the apk package it's either this or cloning down the `git` source code and manually building it statically linked.

This is the far faster solution. I'm open to suggestions/pushback here though.

whynowy:
The `git` binary was even before my time. But I don't think we need it in the image - we use `go-git` lib, and it worked before. It went to a bad state due to some unknown changes. Need investigation.

tdenmon:
> The `git` binary was even before my time. But I don't think we need it in the image - we use `go-git` lib, and it worked before. It went to a bad state due to some unknown changes. Need investigation.

Could be related to [go-git #838](https://github.com/go-git/go-git/issues/838)? There's a [refactor PR](https://github.com/go-git/go-git/pull/1111) to explicitly drop any dependency on an external git binary.

whynowy:
> > The `git` binary was even before my time. But I don't think we need it in the image - we use `go-git` lib, and it worked before. It went to a bad state due to some unknown changes. Need investigation.
> 
> Could be related to [go-git #838](https://github.com/go-git/go-git/issues/838)? There's a [refactor PR](https://github.com/go-git/go-git/pull/1111) to explicitly drop any dependency on an external git binary.

Probably...

Issue #3242, opened 2024-08-16
EventSource - Service annotations


pburgisser:
**Is your feature request related to a problem? Please describe.**
On GKE with Application LoadBalancer, when exposing a webhook a HTTP healthcheck is automatically created and points to the service port (for example 12000) on path /. The checks fail as the probe is on /health. We can override the path to user by creating a backendconfig. However, we need to reference that backendconfig in the service's annotations.

[Documentation](https://github.com/argoproj/argo-events/blob/master/api/event-source.md#argoproj.io/v1alpha1.Service) of the EventSource shows that we can't inject custom annotations
 
**Describe the solution you'd like**
In the `service`'s part of the EventSource, we should be able to add annotations.

**Describe alternatives you've considered**
Creating (and "duplicating") a service that points to the webhook pod with the annotations

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

MenD32:
I have also experienced this issue, a quick fix is to define the BackendConfig in the GCP console but it'd be very nice to have it work out-of-the-box in k8s

MenD32:
i opened a PR since this change can be pretty easy to implement

PR #3257, opened 2024-08-24
feat: eventsource service metadata

MenD32:
Checklist:

added  apicommon.Metadata as a field to the Service in EventSourceSpec, to allow custom configuration of labels and annotations. 

inspired by issue #3242


MenD32:
i am not sure why kafka E2E fails, since the code changes shouldn't affect it at all. if someone has a clue of why this would happen, I'm open to suggestions

controllers/eventsource/resource.go
```diff
@@ -397,6 +398,8 @@ func buildService(args *AdaptorArgs) (*corev1.Service, error) {
 			Selector:  args.Labels,
 		},
 	}
+	svc.ObjectMeta.SetAnnotations(args.EventSource.Spec.Service.Metadata.Annotations)
```
whynowy, line 401: nil check.

controllers/eventsource/resource.go
```diff
@@ -397,6 +398,8 @@ func buildService(args *AdaptorArgs) (*corev1.Service, error) {
...
+	svc.ObjectMeta.SetLabels(args.EventSource.Spec.Service.Metadata.Labels)
```
whynowy, line 402: nil check.

controllers/eventsource/resource.go
MenD32: This is an empty dict by default, using kubebuilder comments, I will add a comment in the function denoting that

controllers/eventsource/resource.go
MenD32: This is an empty dict by default, using kubebuilder comments, I will add a comment in the function denoting that

controllers/eventsource/resource.go
MenD32: I added an if statement because there is a change that the yaml will be nil explicitly

controllers/eventsource/resource.go
MenD32: I added an if statement because there is a change that the yaml will be nil explicitly

MenD32:
i added a 3 second sleep in the test, since the test was inconsistent with Kafka, even though the changes are completely irrelevant

controllers/eventsource/resource.go
```diff
@@ -397,6 +398,16 @@ func buildService(args *AdaptorArgs) (*corev1.Service, error) {
 			Selector:  args.Labels,
 		},
 	}
+
+	if args.EventSource.Spec.Service.Metadata != nil {
+		if args.EventSource.Spec.Service.Metadata.Labels != nil {
+			svc.ObjectMeta.SetLabels(args.EventSource.Spec.Service.Metadata.Labels)
```
whynowy, line 404: Instead of `SetLabels` and `SetAnnotations`, can we do for loops and copy the labels and annotations over? `Set` means original labels and annotations will be replaced.

controllers/eventsource/resource.go
```diff
@@ -397,6 +397,22 @@ func buildService(args *AdaptorArgs) (*corev1.Service, error) {
...
+
+	labels := mergeLabels(args.EventSource.Labels, args.Labels)
+	annotations := make(map[string]string)
+
+	if args.EventSource.Spec.Service.Metadata != nil {
+		if args.EventSource.Spec.Service.Metadata.Labels != nil {
+			labels = mergeLabels(args.EventSource.Spec.Service.Metadata.Labels, labels)
+		}
+		if args.EventSource.Spec.Service.Metadata.Annotations != nil {
+			annotations = mergeLabels(args.EventSource.Spec.Service.Metadata.Annotations, annotations)
```
whynowy, line 409: The function `mergeLabels` is weird, but we can refactor it later.

whynowy (approved):

whynowy:
@MenD32 - can you resolve the conflict?

MenD32:
Yeah

test/e2e/functional_test.go
```diff
@@ -189,6 +189,8 @@ func (s *FunctionalSuite) TestMetricsWithWebhook() {
 		Contains("argo_events_event_processing_duration_milliseconds").
 		Contains("argo_events_events_processing_failed_total")
 
+	time.Sleep(3 * time.Second)
```
whynowy, line 192: Do we need this?

api/event-source.html
whynowy, line 1: We don't have this file any more.

api/event-source.md
whynowy, line 1: same for this file.

test/e2e/functional_test.go
MenD32: i had some weird behavior with the kafka e2e test, and for some reason it helped, but i can remove it.

whynowy:
@MenD32 - unfortunately you have to resolve the conflicts again, sorry for the inconvenience.

ryancurrah:
@MenD32 why would you want to allow custom configuration of labels and annotations on a Service?

PR #3282, opened 2024-09-13
fix:eventsource crash on cast #3238

### LINKED ISSUE #3238: Panic on unchecked cast: `interface conversion: interface {} is cache.DeletedFinalStateUnknown, not *unstructured.Unstructured`
**Describe the bug**
We have set up an argo eventsource in our kubernetes cluster. Regularly (but unpredictably), the eventsources crashes in a panic. The panic message is a consistent problem, namely that a cast of an interface is done without checking.

Here is an example of a panic stack trace:

```panic: interface conversion: interface {} is cache.DeletedFinalStateUnknown, not *unstructured.Unstructured

goroutine 159 [running]:
github.com/argoproj/argo-events/eventsources/sources/resource.passFilters(0xffffffffffffffff?, 0xffffffffffffffff?, {0x1?, 0x0?, 0x5b89900?}, 0x1?)
	/home/runner/work/argo-events/argo-events/eventsources/sources/resource/start.go:290 +0xa53
github.com/argoproj/argo-events/eventsources/sources/resource.(*EventListener).StartListening.func2(0xc001a59b60)
	/home/runner/work/argo-events/argo-events/eventsources/sources/resource/start.go:131 +0x6b
github.com/argoproj/argo-events/eventsources/sources/resource.(*EventListener).StartListening.func3()
	/home/runner/work/argo-events/argo-events/eventsources/sources/resource/start.go:178 +0x122
created by github.com/argoproj/argo-events/eventsources/sources/resource.(*EventListener).StartListening in goroutine 93
	/home/runner/work/argo-events/argo-events/eventsources/sources/resource/start.go:173 +0x8e8
```
Checking the start.go:290 we can see an unchecked cast. An `*unstructured.Unstructured` is expected but in this case a `cache.DeletedFinalStateUnknown` is provided.

**To Reproduce**
Unfortunately, this is a thing that happens _sometimes_ and we have a difficult time finding out when the specified panic happens. 
In general, passing an `*InformerEvent` with an Obj that is not `*unstructured.Unstructured`, will fail the unchecked cast.

**Expected behavior**
If the InformerEvent Obj is not an `*unstructured.Unstructured`, it should be ignored in the passFilters function. In general, the eventsource should not crash for any reason, in this case when an unexpected type is encountered.

 - Kubernetes: v1.29.4
 - Argocli: v3.5.6
 - Argo Events: v1.9.2

---

Daimanta:
Checklist:

Fixes #3238

This fix prevents argo-events from crashing when an unexpected, that is not a `*unstructured.Unstructured`, is encountered. The solution is to filter out those objects as they cannot be parsed and will not have information that can be used in the rest of process.

In the old situation argo-events would crash on receiving the unexpected object.
In the new situation argo-events will ignore the unexpected object.

In case that the object is indeed a `*unstructured.Unstructured`, the program will behave as it did before.

### DIFF
--- a/USERS.md
+++ b/USERS.md
@@ -36,6 +36,7 @@ Organizations below are **officially** using Argo Events. Please send a PR with 
 1. [Mobimeo GmbH](https://mobimeo.com/en/home/)
 1. [OneCause](https://www.onecause.com/)
 1. [Pinnacle Reliability](https://pinnaclereliability.com/)
+1. [PDOK](https://pdok.nl)
 1. [Phrase](https://www.phrase.com/)
 1. [Produvar](https://www.produvar.com/)
 1. [ProPoint Solutions](https://supersalon.com)
--- a/eventsources/sources/resource/start.go
+++ b/eventsources/sources/resource/start.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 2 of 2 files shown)


whynowy (approved):
Thank you!

PR #3240, opened 2024-08-15
feat(sensor): disable leader election for jetstream eventbus

mweibel:
Permit HA sensor when JetStream eventbus is used for sensors.

Checklist:

* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).

PR #3303, opened 2024-09-27
feat(sensor): disable leader election for jetstream eventbus

ryancurrah:
Supersedes https://github.com/argoproj/argo-events/pull/3240

Checklist:

### DIFF
--- a/docs/sensors/ha.md
+++ b/docs/sensors/ha.md
@@ -12,10 +12,10 @@ behaviors!**
 ## Kubernetes Leader Election
 
 By default, Argo Events will use NATS for the HA leader election except when
-using a Kafka Eventbus, in which case a leader election is not required as a
-Sensor that uses a Kafka EventBus is capable of horizontally scaling. If using
-a different EventBus you can opt-in to a Kubernetes native leader election by
-specifying the following annotation.
+using a Kafka or JetStream Eventbus, in which case a leader election is not 
+required as a Sensor that uses a Kafka/JetStream EventBus is capable of 
+horizontally scaling. If using a different EventBus you can opt-in to a Kubernetes 
+native leader election by specifying the following annotation.
 ```yaml
 annotations:
   events.argoproj.io/leader-election: k8s
--- a/pkg/sensors/listener.go
+++ b/pkg/sensors/listener.go
@@ -58,9 +58,9 @@ func (sensorCtx *SensorContext) Start(ctx context.Context) error {
 	replicas := int(sensorCtx.sensor.Spec.GetReplicas())
 	leasename := fmt.Sprintf("sensor-%s", sensorCtx.sensor.Name)
 
-	// sensor for kafka eventbus can be scaled horizontally,
+	// sensor for kafka or jetstream eventbus can be scaled horizontally,
 	// therefore does not require an elector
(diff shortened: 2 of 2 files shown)


whynowy:
What is the reason leader election is not needed when using jetstream? Assume the trigger relies on events from multiple sources, with multiple sensor instances, it will be an issue right?

ryancurrah:
Leader election is not needed when using JetStream because, like Kafka consumer groups, it supports horizontal scaling through queue subscriptions, which distribute events across multiple sensor instances, ensuring that only one instance processes each event. Similar to how Kafka consumer groups ensure that each message is processed by only one consumer within a group, JetStream queue subscriptions achieve the same event distribution. Though it is required to ensure that queue subscriptions are enabled; otherwise, all sensor instances will receive and process the same events (If sensor leader election is turned off), leading to duplicate event processing.

Before disabling leader election for JetStream, I guess my PR should verify that the queue subscription is enabled to ensure events are handled by only one sensor instance.

Queue subscriptions were enabled in this commit https://github.com/argoproj/argo-events/commit/546129637287e7a4586333b840dbea55050c43b6.

ryancurrah:
Leader election is not needed when using JetStream because, like Kafka consumer groups, it supports horizontal scaling through queue subscriptions, which distribute events across multiple sensor instances, ensuring that only one instance processes each event. Similar to Kafka consumer groups, where each message is processed by only one consumer within a group, JetStream queue subscriptions achieve the same event distribution. It is important to ensure that queue subscriptions are enabled; otherwise, if sensor leader election is turned off, all sensor instances will receive and process the same events, leading to duplicate event processing.

Before disabling leader election for JetStream, my PR should verify that queue subscriptions are enabled to ensure that events are handled by only one sensor instance.

Queue subscription support was added in this commit: https://github.com/argoproj/argo-events/commit/546129637287e7a4586333b840dbea55050c43b6.

whynowy:
> Leader election is not needed when using JetStream because, like Kafka consumer groups, it supports horizontal scaling through queue subscriptions, which distribute events across multiple sensor instances, ensuring that only one instance processes each event. Similar to Kafka consumer groups, where each message is processed by only one consumer within a group, JetStream queue subscriptions achieve the same event distribution. It is important to ensure that queue subscriptions are enabled; otherwise, if sensor leader election is turned off, all sensor instances will receive and process the same events, leading to duplicate event processing.
> 
> Before disabling leader election for JetStream, my PR should verify that queue subscriptions are enabled to ensure that events are handled by only one sensor instance.
> 
> Queue subscription support was added in this commit: [5461296](https://github.com/argoproj/argo-events/commit/546129637287e7a4586333b840dbea55050c43b6).

This is not my concern. My concern is about the following scenario:

Sensor `s1` has dependencies on `es1 && es2`, with multiple sensor instances, those events can go to different pods, which will not be able to meet the `&&` condition.

ryancurrah:
> This is not my concern. My concern is about the following scenario:
> 
> Sensor s1 has dependencies on es1 && es2, with multiple sensor instances, those events can go to different pods, which will not be able to meet the && condition.

Hrm how does it work for Kafka? I assume it would be the same.

mweibel:
so this was closed but my earlier PR related to this (#3240) is still open. Any chance this could go further?

Issue #3327, opened 2024-10-14
GitArtifact hardcodes master as the default branch, and there's no mention of this in the documentation


ElectricWarr:
Using a Sensor to trigger creation of some k8s objects pulled directly from a git repo, I expected `GitArtifact` to take the default branch if `source.git.branch` parameter was not set, but instead the trigger fails with the following error:

```
{"level":"error","ts":1728911591.502107,"logger":"argo-events.sensor","caller":"sensors/listener.go:364","msg":"Failed to execute a trigger","sensorName":"mySensor","error":"failed to fetch artifact, failed after retries: failed to checkout. err: reference not found","triggerName":"workflow-trigger-1","stacktrace":"github.com/argoproj/argo-events/sensors.(*SensorContext).triggerActions.func1\n\t/home/runner/work/argo-events/argo-events/sensors/listener.go:364"}
```

`master` seems to be hardcoded as the default [here](https://github.com/argoproj/argo-events/blob/2b8f2fff10ab84e254fdd80511e8c4a60de8503c/pkg/sensors/artifacts/git.go#L39). The docs for `GitArtifact` are [here](https://argoproj.github.io/argo-events/APIs/#argoproj.io/v1alpha1.GitArtifact).

Many _many_ projects, especially new repos, use `main` as their default branch now, and the error is a little opaque to the average user, so it would be nice if it could include the specific reference that wasn't found.

If targeting the default branch isn't possible then I suggest the documentation should include something about how some combination of branch/tag/ref is recommended, and what the behaviour is by default if none are specified.

Edit: I'm aware that default branches are not strictly a git feature, but all the git tools I've ever used behave the same way - when asked to clone a repo from a URL, they always clone the default branch unless explicitly told otherwise.

ElectricWarr:
For reference the source block on my k8s trigger template looks like this:
```yaml
git:
  cloneDirectory: /git/argoproj
  creds:
    password:
      key: password
      name: my-secret
    username:
      key: username
      name: my-secret
  filePath: pipeline.yaml
  url: redacted
```

ElectricWarr:
On a shallow read it looks like where `branch`, `tag` and `ref` are _all_ unset, `r.Clone()` could be used in place of the current `r.Fetch()`/`w.Checkout()` path.

As far as I can see this would be fully backwards-compatible, the only affected users would have to be depending on `master` being the default for repos where `master` wasn't actually the default branch of their target remote. I'd expect very few of those exist since it wasn't documented!

PR #3328, opened 2024-10-14
Update APIs.md

ElectricWarr:
Included strange behaviour of `GitArtifact`'s `branch` parameter relating to #3327


whynowy:
APIs.md is auto generated.

Update the doc at https://github.com/argoproj/argo-events/blob/master/pkg/apis/events/v1alpha1/sensor_types.go#L960 and then run `make codegen`.

ElectricWarr:
Damn, I did wonder what was going on there, sorry! I'll pull down the repo and do it properly with whatever prereqs make needs etc. when I've got a spare moment

PR #3331, opened 2024-10-17
Update link to proto file in build-your-own-trigger.md

atropos112:
Updating the link to proto file, it was moved and the link in the docs is currently invalid.


whynowy (approved):
Thank you!

PR #3371, opened 2024-11-23
Upgrade Argo CLI version to v3.6.0 due to CVE-2024-27304

tonodor:
Just a small update to fix high rated CVE in Argo CLI


whynowy (approved):
Thanks!

PR #3388, opened 2024-12-09
fix: wrap JS_KEY in quotes in nats-cluster.conf and nats.conf

### LINKED ISSUE #3231: nats config fails when JS_KEY value starts with an integer
**Describe the bug**
`eventbus-default-js` pods fail to start if the randomly generated secret/eventbus-default-js-server "encryption" value starts with an integer. The error that is reported is:

> nats-server: variable reference for 'JS_KEY' on line 17 could not be parsed: Parse error on line 1: 'Expected a top-level value to end with a new line, comment or EOF, but got 'N' instead.'

**To Reproduce**
1. edit secret value for "encryption" to be a base64 encoded value of a key that starts with an integer. Example
`printf "384KNjNxieWe" | base64`
Mzg0S05qTnhpZVdl
2. delete eventbus pods so they restart
3. See error

**Expected behavior**
Pods should start

**Screenshots**

 - Kubernetes: 1.29.6
 - Argo Events: v1.8.0

**Additional context**
Editing the cm/eventbus-default-js-config and adding double quotes around the $JS_KEY environment variable appears to fix the issue and allow the nats server to read the value as a string properly.

```
jetstream {
     key: "$JS_KEY"
```
https://github.com/argoproj/argo-events/blob/7b44edd366a35bd187edcbb86597ea67d50f47bc/controllers/eventbus/installer/assets/jetstream/nats.conf#L17

https://github.com/argoproj/argo-events/blob/7b44edd366a35bd187edcbb86597ea67d50f47bc/controllers/eventbus/installer/assets/jetstream/nats-cluster.conf#L17

---

ItsKev:
Fixes an issue where the EventBus fails to start if the JS_KEY environment variable begins with a number, causing the following error on startup:

> nats-server: variable reference for 'JS_KEY' on line 17 could not be parsed: Parse error on line 1: 'Expected a top-level value to end with a new line, comment or EOF, but got '1' instead.'

Fixes #3231

### DIFF
--- a/pkg/reconciler/eventbus/installer/assets/jetstream/nats-cluster.conf
+++ b/pkg/reconciler/eventbus/installer/assets/jetstream/nats-cluster.conf
@@ -14,7 +14,7 @@ server_name: $POD_NAME
 #                                 #
 ###################################
 jetstream {
-  key: $JS_KEY
+  key: "$JS_KEY"
   store_dir: "/data/jetstream/store"
   {{.Settings}}
 }
--- a/pkg/reconciler/eventbus/installer/assets/jetstream/nats.conf
+++ b/pkg/reconciler/eventbus/installer/assets/jetstream/nats.conf
@@ -14,7 +14,7 @@ server_name: $POD_NAME
 #                                 #
 ###################################
 jetstream {
-  key: $JS_KEY
+  key: "$JS_KEY"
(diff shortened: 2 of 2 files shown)


whynowy (approved):
LGTM, thank you!

PR #3414, opened 2025-01-06
move fuzz tests upstream from cncf-fuzzing

AdamKorcz:
Checklist:

This moves over several of argo-events fuzz tests from [cncf-fuzzing](https://github.com/cncf/cncf-fuzzing/tree/main/projects/argo). These fuzz tests have been running for several years continuously on OSS-Fuzz, but due to recent refactoring in argo-events they broke. I have fixed the fuzz tests and rewritten them to native golang fuzzers (that use std libs fuzzing engine). Having them upstream will make maintenance much easier.

### DIFF
--- a/pkg/eventsources/sources/awssns/fuzz_test.go
+++ b/pkg/eventsources/sources/awssns/fuzz_test.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2025 The Argoproj Authors.
+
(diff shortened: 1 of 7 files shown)


AdamKorcz:
The failing unit test looks unrelated.

whynowy (approved):
thank you!

PR #3433, opened 2025-01-15
Move fuzzers upstream from cncf-fuzzing

AdamKorcz:
Checklist:

Moves more fuzzers in similar fashion as https://github.com/argoproj/argo-events/pull/3414


whynowy (approved):
Thanks

PR #3434, opened 2025-01-16
feat: add workload identity support to azure event hub eventsource

### LINKED ISSUE #3207: Support Azure Managed Identity for Azure Event Hubs event-source
Azure EventHubs event source currently supports key-based authentication only. When deploying Azure EventHubs with local authentication disabled, which is a security best practice, there is no way to configure the event source to use Microsoft Entra Workload Identity. This event source should support service principal authentication similar to [Azure ServiceBus event source](https://github.com/argoproj/argo-events/tree/master/eventsources/sources/azureservicebus)

The [Azure EventHubs event source](https://github.com/argoproj/argo-events/tree/master/eventsources/sources/azureeventshub) should be refactored to work similarly to [Azure ServiceBus event source](https://github.com/argoproj/argo-events/tree/master/eventsources/sources/azureservicebus).

Alternative is to leave as-is and not support workload identity.

**Additional context**
As part of this refactoring, the SDK for event hubs needs to be updated from deprecated [azure-event-hubs-go](https://pkg.go.dev/github.com/Azure/azure-event-hubs-go/v3@v3.6.2) to current [azure-sdk-for-go/sdk/messaging/azeventhubs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs). 

Also wanted to note that I'd be happy to contribute a PR for this.

---

pauldotyu:
This PR enhances the Azure Event Hub event source to resolve #3207.

It replaces the deprecated SDK [https://pkg.go.dev/github.com/Azure/azure-event-hubs-go/v3@v3.6.2](https://pkg.go.dev/github.com/Azure/azure-event-hubs-go/v3@v3.6.2) with [https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs) and makes `SharedAccessKeyName` and `SharedAccessKey` fields optional in the `AzureEventsHubEventSource` spec. When both these fields are empty, authentication to Azure Event Hub will be handled via Microsoft Entra ID Workload Identity.

To use Microsoft Entra ID Workload Identity on AKS, you will need the following:
- An AKS cluster with OIDC endpoint and workload identity enabled
- A user assigned managed identity with federated credential
- A service account with required annotation
- EventSource template spec with required pod label and serviceaccount

More info at https://learn.microsoft.com/azure/aks/workload-identity-overview?tabs=go


pkg/eventsources/sources/azureeventshub/start.go
```diff
@@ -61,87 +63,106 @@ func (el *EventListener) GetEventSourceType() aev1.EventSourceType {
...
+
+					if event.PartitionKey != nil {
+						eventData.PartitionKey = *event.PartitionKey
+					}
+
+					eventBytes, err := json.Marshal(eventData)
+					if err != nil {
+						el.Metrics.EventProcessingFailed(el.GetEventSourceName(), el.GetEventName())
+						log.Errorw("failed to marshal the event data", "eventSource", el.GetEventName(), "messageID", *event.MessageID, zap.Error(err))
+						return
```
whynowy, line 148: This `return` will exit the goroutine for the corresponding partition, is it expected?

docs/developer_guide.md
```diff
@@ -33,11 +33,104 @@ eval $(minikube docker-env)
...
+### 4. Changing Types
 
 If you're making a change to the `pkg/apis` package, please ensure you re-run
 following command for code regeneration.
 
 ```
 make codegen
 ```
+
+### Deploying and testing your changes on a cluster
```
whynowy, line 45: Thanks for adding all of these. Might be worth of mentioning local clusters like `k3d`, `kind` and `minikube` are supported for testing by running `make start`/`make image`, before giving an option of pushing the test image to any image registry (and mentioning ttl.sh is a convenient one). WDYT?

whynowy (commented):
Thanks and sorry for the late response!

pkg/eventsources/sources/azureeventshub/start.go
pauldotyu: Ah, good call! I suppose I can use `continue` instead of `return` here so that the goroutine does not exit abruptly. I can change that.

whynowy (approved):
Thank you!

PR #3441, opened 2025-01-20
fix: compare errors with string to avoid log spew in jetstream sensor

sercant:
This PR fixes an issue in the error comparison to avoid log spew. Instead of comparing the errors directly, compare the string versions.

Checklist:


sercant:
Hello @whynowy !

We are submitting this change because on our setup, when one of the NATS Jetstream is changing instance (spot), argo-events eventsource emits millions of logs. It would be great if you could take a look at it when you have time. 🙏

whynowy (approved):
LGTM

Issue #3432, opened 2025-01-12
Tutorial: Trigger Template Parameterization not working


haiwu:
**Describe the bug**
Just follow the exact steps in Argo Events tutorial section 'Trigger Template Parameterization' at https://argoproj.github.io/argo-events/tutorials/02-parameterization/, and always getting 'hello world' output, instead of expected output of 'amazing!!'

**To Reproduce**
Just follow the exact steps in Argo Events tutorial section 'Trigger Template Parameterization' at https://argoproj.github.io/argo-events/tutorials/02-parameterization/

**Expected behavior**
Should have got expected output of 'amazing!!', but got 'hello world' output

Someone else already mentioned the same here: https://github.com/argoproj/argo-events/discussions/3161. 

If argo events no longer supports this use case, please remove this misleading section in the tutorial.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #3457, opened 2025-02-04
docs(sensor): move parameters to template level

joepke:
Fix the trigger parameterization example from https://argoproj.github.io/argo-events/tutorials/02-parameterization/, as mentioned in #3432 and #3161.

The outer `parameters` section should be at template level, as defined in the Trigger spec (https://argoproj.github.io/argo-events/APIs/#argoproj.io/v1alpha1.Trigger). 

The other trigger parameterization example is configured correctly (https://github.com/argoproj/argo-events/blob/master/examples/sensors/complete-trigger-parameterization.yaml).


whynowy (approved):
Thank you!

PR #3480, opened 2025-02-18
port-forward on svc/webhook-eventsource-svc 12000:12000 instead of pod.

tbernacchi:
Port-forward on svc/webhook-eventsource-svc instead of pod.

Checklist:


whynowy (approved):
This is way easier, thanks

Issue #3506, opened 2025-03-06
Use new expr-lang builtins instead of pkg functions


crenshaw-dev:
**Describe the solution you'd like**

expr-lang has included int, float, and string functions since 1.14.0. Argo Workflows currently uses expr 1.16.9. Instead of using those functions as implemented in argoproj/pkg, we should use the built-ins.

We should also inline the implementation of jsonpath, since it is trivial and doesn't make sense to externalize to the pkg repo.

This simplifies maintenance, because we don't have to worry about the pkg repo lifecycle as much.

**Additional context**

More info here: https://github.com/argoproj/argo-cd/issues/22175

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #3507, opened 2025-03-06
chore: use expr built-in functions instead of pkg (#3506)

crenshaw-dev:
Fixes https://github.com/argoproj/argo-events/issues/3506

### Motivation

expr has had built-in `int`, `float`, and `string` functions since [1.14.0](https://github.com/expr-lang/expr/releases/tag/v1.14.0) (Events is currently using 1.15.5). 

The `jsonpath` function is trivial and doesn't make sense to maintain in a separate library.

More info here: https://github.com/argoproj/argo-cd/issues/22175

Equivalent Argo Workflows PR: https://github.com/argoproj/argo-workflows/pull/14254

### Verification

Compared the [expr-lang builtins](https://github.com/expr-lang/expr/blob/77e876bb7cda24c5aeedbfd1a0efa1465fb503bb/builtin/lib.go) with the [pkg functions](https://github.com/argoproj/pkg/blob/cbc37dc61de58f5431fddc2fc377b56a69e94de2/expr/function.go) to confirm they're functionally identical.


whynowy (approved):
LGTM

whynowy:
@crenshaw-dev - can you fix the conflict?

PR #2951, opened 2023-12-19
feat: add alibabacloud mns as eventsource

shuangkun:
Checklist:

* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).

Add AlibabaCloud MNS as eventsource for argo-events.

PR #3550, opened 2025-04-09
feat: add alibabacloud mns eventsource

aln8:
Checklist:

* [x ] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).


aln8:
It's a fix for https://github.com/argoproj/argo-events/pull/2951, not sure why that's closed.

pkg/eventsources/sources/alibabacloudmns/start.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	queue := ali_mns.NewMNSQueue(el.MNSEventSource.Queue, *mnsClient, 50)
+
+	respChan := make(chan ali_mns.MessageReceiveResponse)
+	errChan := make(chan error)
+	received := make(chan struct{})
+	for {
+		// goroutine for mns client queue to block wait
+		go func() {
+			queue.ReceiveMessage(respChan, errChan, 30)
```
whynowy, line 59: Is this function used to receive one message? Does it need to use a goroutine to run it?

pkg/eventsources/sources/alibabacloudmns/start.go
```diff
@@ -0,0 +1,126 @@
...
+			received <- struct{}{}
+		}()
+
+		// goroutine to wait/handle response channel data
+		go func() {
```
whynowy, line 64: If the assumption above is right, this does not need a goroutine as well.

pkg/eventsources/sources/alibabacloudmns/start.go
aln8: Yes, but is block wait here, so if no new message come, `ctx.Done()` could never be selected.

I put this in goroutine and add a received channel to let StartListening quit first.

pkg/eventsources/sources/alibabacloudmns/start.go
aln8: The alibaba mns lib has poor design I think, for my understanding, `respChan` is buffer free channel, so it need read and write in different goroutine.

pkg/eventsources/sources/alibabacloudmns/start.go
aln8: @whynowy

whynowy: I assume the 3rd parameter is the timeout?

I think something like below can be done:

```golang
// Start a goroutine to receive messages
go func() {
  for {
    select {
      case <-ctx.Done():
        close(respChan)
        close(errChan)
      default:
        // Receive a message with smaller timeout
    }
  }
}()

for {
  select {
    case resp, ok := <-respChan:
      if !ok { // Channel closed
        /////////
      }
      xxxx
    case err := <-errChan:
      xxxxxx
  }
}
```

aln8:
@whynowy I update the code as your commented, pls do another review, thx.

whynowy (approved):
LGTM, Thanks @aln8 !

whynowy:
@aln8 - could you also create a doc for it with a followup PR, and add it to https://github.com/argoproj/argo-events/blob/master/mkdocs.yml#L84?

PR #3579, opened 2025-05-02
feat: schema registry avro support (#1503)

peknu:
Add the ability to consume avro messages and use schemaRegistry to deserialize them into JSON format in a Kafka Event Source.

Checklist:


whynowy (approved):
LGTM, thanks

PR #3580, opened 2025-05-02
feat(kafka): enabling tls connection without client authentication

brunodomenici:
# Kafka TLS

As we can see in #1277 , we can't enable TLS without setting `insecureSkipVerify` to true or setting client authentication.

I propose a `TLS.Enable` with default to false (to keep compatibility with existing configurations). With that, we can set only this to true to activate TLS without needing to set other configuration in this object.

Checklist:


whynowy:
Thanks @brunodomenici .

What is the difference between this new property and `insecureSkipVerify`?

shearn89:
To add a use case here - we deploy AWS MSK, which defaults to `security.protocol=SSL` for comms even when the cluster requires no auth. We have to currently set `insecureSkipVerify = true` to connect at all, even though we're not using mTLS. It would be great to be able to use TLS to the broker (but **not** mTLS), which (if I understand correctly) this PR allows us to do.

Mcfloy:
Hello, I'll mention @daniel-codefresh (👋 ) to see if we can have another feedback on this pull request, and let us know if there's something we can improve.

This pull request (and #3600) would allow the @adeo organization to finally be able to use Argo Events, for now we are waiting for feedback or approval to merge so that we can move forward. (we're a bit jealous 🥲 to not be able to play with it)

whynowy:
> To add a use case here - we deploy AWS MSK, which defaults to `security.protocol=SSL` for comms even when the cluster requires no auth. We have to currently set `insecureSkipVerify = true` to connect at all, even though we're not using mTLS. It would be great to be able to use TLS to the broker (but **not** mTLS), which (if I understand correctly) this PR allows us to do.

Got it, thanks!

whynowy:
@brunodomenici - can you re-run codegen? I will approve it.

brunodomenici:
> @brunodomenici - can you re-run codegen? I will approve it.

Hello @whynowy . Done. Also, I updated my branch from `master`

whynowy (approved):

PR #3605, opened 2025-05-20
bump golangci-lint to v2

tylerauerbeck:
Updates golangci-lint to v2 using `golangci-lint migrate` from v2 cli and fixes associated lints


whynowy (approved):
Thanks!

PR #3640, opened 2025-06-19
fix: Fix Sensor Validation for admission webhook

ko-be:
We have been running into an issue when deleting sensors, when the eventbus has already been deleted.  When we try to remove the finalizer from a sensor we're attempting to force delete, the admission webhook denies the request with:

```
error: sensors.argoproj.io "flex-managed-sensor" could not be patched: admission webhook "webhook.argo-events.argoproj.io" denied the request: failed to get EventBus eventBusName=default; err=eventbus.argoproj.io "default" not found
```

This PR should fix this bug.

While creating additional test cases to catch this bug, also noticed that the current code panics when eventBusClient is nil, so added a guard against that as well.

---

Checklist:


whynowy (approved):
Thanks

PR #3655, opened 2025-07-03
Add a dynamic headers block for http trigger

### LINKED ISSUE #2978: Put value from event data in header for HTTP Trigger
Currently `headers` field is for static values.
We need possibility to put value from event data in header when using HTTP trigger.

Given we have event like this:
```
    {
        "context": {
          "type": "type_of_event_source",
          "specversion": "cloud_events_version",
          "source": "name_of_the_event_source",
          "id": "unique_event_id",
          "time": "event_time",
          "datacontenttype": "type_of_data",
          "subject": "name_of_the_configuration_within_event_source"
        },
        "data": {
          "header": {/* the headers from the request received by the event-source from the external entity */},
          "body": { "userId":"2" },
        }
    }
```

I wanna have something like this:
```
          headers:
            MY_HEADER_1: {{ contextKey: id }}
            USER_ID: {{ dataKey: body.userId }}
```

**Additional context**
This will be useful when you need to have guarantees that events generated for user with `id` number `1` will be received only by one instance of other service. We can calculate hash based on `USER_ID` and route requests to specific instance.

---

shearn89:
# What?

Add in a `dynamicHeaders` block to the HTTP trigger

# Why?

Fixes #2978 

I want to be able to pass in headers from the event payload, rather than hard code them. This PR should create a new parameter for the trigger that uses the same spec as the `payload`, called `dynamicHeaders`.

Intended usage:

```yaml
kind: Sensor
metadata:
  name: my-sensor
spec:
  eventBusName: jetstream
  dependencies:
    - name: my-event
      eventSourceName: github
      eventName: webhook
  triggers:
    - template:
        name: my-trigger
        http:
          dynamicHeaders:
            - src:
                contextKey: eventContext
                dependencyName: my-event
                dataKey: "headers.Content-Type"
              dest: "Content-Type"
            - src:
                contextKey: eventContext
                dependencyName: my-event
                dataKey: "headers.X-Github-Event"
              dest: "X-Github-Event"
          payload: {} #etc
```

Tested in my own cluster - created a sensor that fires back to another webhook, which gets logged.

Using the above snipped, I can see these headers in the log:

```
{
  "header": {
    "Accept-Encoding": [
      "gzip"
    ],
    "Content-Length": [
      "135"
    ],
    "Content-Type": [
      "[\"application/json\"]"
    ],
    "User-Agent": [
      "Go-http-client/2.0"
    ],
    "X-Forwarded-Port": [
      "443"
    ],
    "X-Forwarded-Proto": [
      "https"
    ],
    "X-Github-Event": [
      "[\"push\"]"
    ]
  },
  "body": {
    "ref": "refs/heads/main",
    "repository": {
      "html_url": "https://github.com/my-repo-url",
      "default_branch": "main"
    }
  }
}
```

...importantly - the `X-Github-Event` is being passed through.

Checklist:


shearn89:
I'm not sure if the failing test is due to this PR - looks to be in the Kafka trigger, which I've not touched... 🤔 

Update: yes, it's because the Kafka test is getting events out of the expected order, so just a flaky test.

shearn89:
@whynowy - any chance of a review this week? Would be great to get this merged as then I can roll out Events more widely across our org.

whynowy (approved):
Thanks, LGTM

PR #3685, opened 2025-08-08
Fixes #3684 Http trigger host

### LINKED ISSUE #3684: Feature Request: Allow overriding Host header in HTTPTrigger
## Summary

Add an optional `host` field to HTTPTrigger that, when set, overrides the `http.Request.Host` value used for the outgoing HTTP request.

## Current Behavior

- HTTPTrigger builds an `http.Request` using net/http defaults.
- The Host header is implicitly set from the request URL (`req.URL.Host`).
- There is no way to explicitly set a different Host value in the trigger spec.

## Problem Statement

In some deployment scenarios, the Host header must differ from the request URL for correct routing or TLS behavior.

Examples may include virtual hosting / reverse proxy routing – Many ingress controllers and API gateways route by Host.

Without this capability, certain HTTP-based triggers cannot be configured correctly without external workarounds.

## Proposed Change

1. Schema Update

Add an optional host field to HTTPTrigger in the CRD:

```yaml
triggers:
  - template:
      http:
        url: http://10.0.0.5/api
        host: api.example.com
```

2. Runtime Behavior

In HTTP trigger execution:

```go
if trigger.Host != "" {
    request.Host = trigger.Host
}
```

This will override the Host header.

## Example Use Case

A user has a Kubernetes ingress configured for `api.example.com`, but for internal reasons the trigger must point to `http://10.0.0.5/....`

Without a `host` override, the ingress rejects the request because `Host` does not match the expected hostname.

fabricebrito:
Checklist:

* [ X] My organization is added to [USERS.md](https://github.com/argoproj/argo-events/blob/master/USERS.md).

Add an optional host field to HTTPTrigger that, when set, overrides the http.Request.Host value used for the outgoing HTTP request.

Fixes #3684


fabricebrito:
Closing in favour of a PR with signed commits

Issue #3695, opened 2025-08-11
Message distribution across partitions when using Kafka EventBus


qlimenoque:
**Describe the bug**
Argo Events is using the trigger name as the Kafka message key for all messages produced by a sensor

Since all sensor replicas have the same trigger name, all messages end up with the same key

In Kafka, messages with the same key are routed to the same partition, which makes Sensor's pod utilize only 1 partition instead of distributing across all available partitions

This makes Sensor's horizontal scaling unworkable, because only one consumer can only work with one partition

**To Reproduce**
1. Create **EventSource** that produces messages to Kafka EventBus
2. Send 5 messages to trigger
3. See that all of them go to the single partition with the same key as TriggerName

**Expected behavior**
Sensors should produce messages to Kafka topics in different partitions
For example, if I have 6 partitions, I would like to have the messages spread across all of them
In this case, I can scale up the Sensor up to 6 replicas and process each partition in parallel

**Screenshots**
<img width="3446" height="1768" alt="Image" src="https://github.com/user-attachments/assets/f59fe900-9781-4de4-b58b-dba59a59a7cf" />

**Environment (please complete the following information):**
 - Kubernetes: v.1.33.2
 - Argo Events: v1.9.6

**Additional context**

One of the solutions might be Producer configuration, where a user can set Partitioner in Kafka EventBus CRD. List of available Partitioner is described in Sarama Library [partitioner.go](https://github.com/IBM/sarama/blob/main/partitioner.go)

The other one (more complex) solution is to add entropy to message keys via adding UUID, date, etc

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #3696, opened 2025-08-11
feat: make KafkaBus partitioner configurable

qlimenoque:
This PR resolves https://github.com/argoproj/argo-events/issues/3695

1. Adds an optional field `spec.kafka.partitioner` to the **EventBus** CRD to select the Kafka producer partitioning mode: `random`, `hash`, `roundrobin`, or `manual`. 

2. Changes default partitioner mode from `hash` to `random`


whynowy:
@qlimenoque - thanks for raising the PR. One question - what is the impact of changing default partitioner mode from `hash` to `random` for existing users?

qlimenoque:
@whynowy Hi!

Speaking in the context of Argo systems, as far as I know, this change should only improve topic consumption by really "enabling" parallelism and scaling of sensors (consumers)

The only thing this could potentially affect is if other custom consumers read the topics, and there are some settings there to read from specific partitions, but that's just my guess

We can either 
1. Set the default hash value, allowing users to choose the mode themselves

or

2. Make random mode the default, but highlight this change in the release notes

What do you think?

whynowy:
> @whynowy Hi!
> 
> Speaking in the context of Argo systems, as far as I know, this change should only improve topic consumption by really "enabling" parallelism and scaling of sensors (consumers)
> 
> The only thing this could potentially affect is if other custom consumers read the topics, and there are some settings there to read from specific partitions, but that's just my guess
> 
> We can either
> 
> 1. Set the default hash value, allowing users to choose the mode themselves
> 
> or
> 
> 2. Make random mode the default, but highlight this change in the release notes
> 
> What do you think?

Thanks for the information! I'm okay with the change if it does not have any impact on the argo-events features.

Can you re-run codgen to fix the conflicts?

qlimenoque:
@whynowy done ✅

whynowy (approved):

PR #3776, opened 2025-10-19
fix(http): correct basicAuth password err handling

thevilledev:
Replaced incorrect check on password retrieval with proper err check in HTTPTrigger.Execute. This ensures failures fetching the password secret surface as errors. Added unit test to verify execution fails early when password secret retrieval fails, and that no HTTP requests are made.

Checklist:


whynowy (approved):
Thanks for fixing.

Issue #3771, opened 2025-10-15
Transform script result cannot be an empty array


MenD32:
**Describe the bug**
The current way the `transform` key is defined, the resulting lua object cannot be parsed to have an empty array.

**To Reproduce**
1. Create a transform script like so:
```
event.body.env = {};
```

use it in a dependency:
```yaml
- src:
    dependencyName: benchmark
    dataKey: body.env
    useRawData: true
  dest: spec.templates.0.container.env
``` 

the created object will be invalid since env expects the object to be a list, not an array

**Expected behavior**
I'd expect it to be possible for argo events to parse empty arrays as well as empty dicts from lua transform

**Environment (please complete the following information):**
 - Kubernetes: [e.g. v1.32]
 - Argo Events: [e.g. v2.4.7]

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

yugannkt:
I noticed this issue is open. I would like to work on it, if possible. Please assign it to me. @whynowy

PR #3796, opened 2025-11-06
feat: Supports empty Lua arrays via metatable flag

yugannkt:
**Problem**
Previously, when using a Lua transform in Argo Events, setting a field to an empty table ({}) always resulted in an empty object (`{}`) in JSON, not an empty array (`[]`). This caused issues for fields (like env) that require an array.

**Solution**
Added support for marking a Lua table as an array using a metatable:
`setmetatable({}, { __is_array = true })`
Updated the Go conversion logic to check for the `__is_array` flag and serialize the table as an array, even if empty.
Added a test to verify this behavior.
Result
Users can now reliably produce empty arrays from Lua transforms, ensuring compatibility with Kubernetes specs that require arrays.

Closes https://github.com/argoproj/argo-events/issues/3771

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,7 +101,7 @@ jobs:
         include:
           - driver: stan
           - driver: jetstream
-          #- driver: kafka
+          - driver: kafka
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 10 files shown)


whynowy (approved):
LGTM, thank you!

PR #3797, opened 2025-11-07
feat: allow duplicate dependencies for kafka

eduardodbr:
Currently [docs](https://argoproj.github.io/argo-events/sensors/more-about-sensors-and-triggers/#duplicate-dependencies) say that NATS streaming doesn’t support duplicate dependencies but NATS JetStream does. However, kafka event bus also doesn’t support it.
This PR aims to allow duplicate dependencies when using kafka event bus mirroring the nats jetstream behavior.

How to reproduce:

Create 1 event source of any type, example:

```yaml
kind: EventSource
metadata:
  name: calendar-kafka
spec:
  calendar:
    calendar:
      interval: 10s
      metadata:
        foo: "1.0"
        baz: "bar"
```

Create a sensor with two dependencies from the same event source name and event name:

```yaml
kind: Sensor
metadata:
  name: calendar-kafka
spec:
  dependencies:
    - name: calendar_foo
      eventSourceName: calendar-kafka
      eventName: calendar
      filters:
        data:
          - path: metadata.foo
            type: string
            value:
              - "1.0"

    - name: calendar_baz
      eventSourceName: calendar-kafka
      eventName: calendar
      filters:
        data:
          - path: metadata.baz
            type: string
            value:
              - "bar"
```

Changes:

- Changed the `dependsOn` function to return all dependencies from the same event source and event name. Previously it would only return one.
- Send the dependency name as kafka header “dependencyName” to the trigger and action topics. 
- Change the `Update` function receive the dependency name as input.
- Change the `Action` function to receive the dependency name as input.


isubasinghe (approved):

whynowy (approved):
Thanks!

PR #3809, opened 2025-11-12
fix: data filter != comparator should require NO matches (NOT IN behavior)

### LINKED ISSUE #3808: bug: DataFilter != comparator with multiple values doesn't work as NOT IN
## Summary

The `!=` comparator in `DataFilter` with multiple values exhibits unintuitive behavior that makes it practically unusable. Instead of implementing true \"NOT IN\" semantics (value must not match ALL items in array), it currently implements \"not equal to ANY\" (passes on first mismatch).

## Current Behavior

When using `comparator: "!="` with multiple values, the filter **passes if the event value is different from ANY value in the array**, which means it will pass for almost any input.

**Example:**
```yaml
filters:
  data:
    - path: body.brandId
      type: string
      value: ["id1", "id2", "id3"]
      comparator: "!="
```

**Event: `{"brandId": "id1"}`**
- Current result: **PASSES** ✓ (because `"id1" != "id2"` is true)
- Expected result: **FAILS** ✗ (because `"id1"` IS in the array)

## Expected Behavior

The `!=` comparator should implement true \"NOT IN\" semantics:
- **PASS**: if value does not match ANY item in the array (value NOT IN array)
- **FAIL**: if value matches any item in the array (value IN array)

## Root Cause

In `pkg/sensors/dependencies/filter.go`, the implementation exits early on the first mismatch:

```go
// Lines ~361-399 (string type)
for _, value := range f.Value {
    // ... regex matching ...
    switch f.Comparator {
    case v1alpha1.NotEqualTo:
        if !match {
            matchResult = true  // First mismatch
        }
    }
    
    if matchResult {
        continue filterData  // Exits loop early - WRONG!
    }
}
return false  // Only reached if ALL values matched
```

The same issue exists for number types (lines ~306-359) and bool types (lines ~278-304).

## Test Coverage Gap

The existing test suite in `filter_data_test.go` has **no test cases** for `!=` comparator with multiple values - all `!=` tests use single-value arrays.

## Reproduction Test Cases

```go
{
    name: "string filter NotEqualTo with multiple values - value IN array",
    args: args{
        data: []v1alpha1.DataFilter{
            {
                Path:       "k",
                Type:       v1alpha1.JSONTypeString,
                Value:      []string{"id1", "id2", "id3"},
                Comparator: "!=",
            },
        },
        operator: v1alpha1.EmptyLogicalOperator,
        event: &v1alpha1.Event{
            Data: []byte(`{"k": "id1"}`),
        },
    },
    expectedResult: false,  // Should FAIL
    expectErr:      false,
}
```

Running this test with current code: **PASSES** (bug)
Running with fixed code: **FAILS** (correct)

## Impact

- The `!=` comparator is essentially broken for practical use with multiple values
- Users expecting \"NOT IN\" behavior will get incorrect results
- No workaround except using expr filters with explicit AND logic

## Proposed Fix

Check ALL values before deciding:
```go
for _, value := range f.Value {
    if match {
        matchFound = true
        break
    }
}

// For != comparator, pass only if NO match was found
if f.Comparator == v1alpha1.NotEqualTo {
    if !matchFound {
        continue filterData  // PASS
    } else {
        return false  // FAIL
    }
}
```

## Environment

- All versions affected (present in current master branch)
- Affects string, number, and bool data types

I'll prepare a PR with the fix and comprehensive test coverage.

uptownhr:
## Summary

Fixes #3808

This PR fixes the `!=` comparator in `DataFilter` to implement true "NOT IN" semantics when used with multiple values.

## Problem

The `!=` comparator was exhibiting unintuitive behavior:
- **Before**: Passed on first mismatch (value != ANY item in array)
- **After**: Passes only if NO matches found (value NOT IN array)

### Example of the Bug

With `comparator: "!="` and `value: ["id1", "id2", "id3"]`:

**Event:** `{"brandId": "id1"}`
- ❌ **Old behavior**: PASSES (because `"id1" != "id2"` is true)
- ✅ **New behavior**: FAILS (because `"id1"` IS in the array)

### Code Changes
- **pkg/sensors/dependencies/filter.go**:
  - Added special handling for `NotEqualTo` comparator in `JSONTypeNumber` case
  - Added special handling for `NotEqualTo` comparator in `JSONTypeString` case
  - Both now check ALL values before deciding (true NOT IN semantics)
  - Other comparators maintain existing behavior

### Test Coverage
- **pkg/sensors/dependencies/filter_data_test.go**:
  - Added 4 new test cases proving the fix:
    1. String filter with value IN array (should FAIL)
    2. String filter with value NOT IN array (should PASS)
    3. Number filter with value IN array (should FAIL)
    4. Number filter with value NOT IN array (should PASS)

## Test Results

```bash
$ go test ./pkg/sensors/dependencies/...
PASS
ok  	github.com/argoproj/argo-events/pkg/sensors/dependencies	0.010s
```

All existing tests pass with no regressions.

## Breaking Change Note

This is technically a breaking change as it modifies the behavior of the `!=` comparator with multiple values. However:
1. The previous behavior was essentially broken/unusable
2. No test coverage existed for this scenario (suggesting low/no usage)
3. The new behavior matches intuitive expectations

## Implementation Details

The fix checks if comparator is `NotEqualTo` and if so:
1. Scans through ALL values in the array
2. Tracks if ANY match is found
3. Only passes if NO match was found (true NOT IN)

For all other comparators, the existing logic is preserved (pass on first match).


whynowy:
Thanks for raising the PR. Couple of questions:

1. Shall we introduce a new comparator for `not in`, like `!in`, `!()`?
2. Can this problem be solved by using other filters like `expr` or `script`?

uptownhr:
> Thanks for raising the PR. Couple of questions:
> 
> 1. Shall we introduce a new comparator for `not in`, like `!in`, `!()`?
> 2. Can this problem be solved by using other filters like `expr` or `script`?

1. good question - I'm thinking given `=` already does `in`, it makes more sense to have `!=` be `not in`. 
2. i am currently using `expr` to get around this - but it would have been nice if `!=` worked as I imagined.

PR #3812, opened 2025-11-14
chore: Add an example sensor to trigger an existing WorkflowTemplate

openJT:
I really wish I had an example of this while integrating argo-events with workflows. While searching for this I could only find outdated information and did not understand why such an example was not in the doc. In my research I saw people complaining about it, unmerged PR with outdated information and AI giving me non working examples based on this bad information.  It was suggested that I contribute this example so here it is!

Feel free to suggest modifications, better wording/naming etc.

Thanks.


whynowy:
Thanks, could you fix the conflicts?

openJT:
Done

PR #3823, opened 2025-11-19
chore: add retest workflow

eduardodbr:
Add GitHub action that is triggered with a comment "/retest" from a Member of the argo project. Similar to [argo-workflows](https://github.com/argoproj/argo-workflows/blob/main/.github/workflows/retest.yaml)

The Action re-runs failed jobs.

Also updated one test to try and make it less flaky.


whynowy (approved):
thanks

PR #3884, opened 2026-01-17
Fix/amqp integer queue args

### LINKED ISSUE #3876: bug: amqp EventSource converts argument values to double from integer
**Describe the bug**

When providing a [QueueDeclareConfig](https://argoproj.github.io/argo-events/APIs/#argoproj.io/v1alpha1.AMQPQueueDeclareConfig) with `arguments` to an amqp EventSource, values that are provided as `integer` will be converted to `double`. This causes an error in the creation of an amqp queue.

Resulting in an error:

`failed after retries: failed to get the delivery for the event source gpu-refresh, failed to declare queue: Exception (406) Reason: \"PRECONDITION_FAILED - invalid arg 'x-expires' for queue 'my-queue-name' in vhost 'dev': \\"expected integer, got double\\"\`

Not proficient at Go, but I think the bug comes from this method:

https://github.com/argoproj/argo-events/blob/9fe771b09ae6e6f6791e2e385dcc60f31a598a2a/pkg/eventsources/sources/amqp/start.go#L294-L305

**To Reproduce**

Configure an amqp EventSource like below:

```yml
kind: EventSource
metadata:
  name: amqp
spec:
  amqp:
    example:
      # amqp server url
      url: amqp://rabbitmq-service.argo-events:5672
      # jsonBody specifies that all event body payload coming from this
      # source will be JSON
      jsonBody: true
      # name of the exchange.
      exchangeName: test
      # exchange type. more info at https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchanges
      exchangeType: fanout
      # routing key
      routingKey: hello
      # optional queue settings
      # if not provided, default values will be used
      queueDeclare:
        name: "my-queue-name"
        arguments: |-
          x-expires: 86400000
```

**Expected behavior**
The expected behavior is that a queue is created, with the correct `x-expires` configured. In reality, no queue is created and an error is returned after a while (not sure exactly when, some retrying is happening in the background.).

**Environment:**
 - Argo Events: v1.9.6

**Additional context**
I have also tried other variables, like `x-message-ttl`, resulting in the same error.

---

yugannkt:
## Description

Fixes #3876 

AMQP EventSource fails to create queues when `queueDeclare.arguments` contain integer values like `x-expires` or `x-message-ttl`, causing RabbitMQ to reject with:
```
PRECONDITION_FAILED - invalid arg 'x-expires' for queue 'my-queue-name' in vhost 'dev': 
expected integer, got double
```

## Root Cause

The `parseYamlTable()` function was using `sigs.k8s.io/yaml`, which internally converts YAML → JSON → Go types. This conversion forces all numeric scalars to `float64` when unmarshalling into `interface{}`-typed values (like `amqplib.Table`).

## Solution

Switched to `gopkg.in/yaml.v3` for native YAML parsing, which preserves integer scalar types as `int`/`int64` instead of converting them to `float64`.

### DIFF
--- a/pkg/eventsources/sources/amqp/start_test.go
+++ b/pkg/eventsources/sources/amqp/start_test.go
@@ -19,24 +19,8 @@ package amqp
 import (
 	"testing"
 
-	amqplib "github.com/rabbitmq/amqp091-go"
 	"github.com/stretchr/testify/assert"
-	"sigs.k8s.io/yaml"
 )
-
-func TestParseYamlTable_Issue3876RootCause(t *testing.T) {
-	// This test demonstrates the original bug mechanism described in
-	// https://github.com/argoproj/argo-events/issues/3876:
-	// sigs.k8s.io/yaml performs YAML->JSON conversion and json.Unmarshal into
-	// interface{}-typed values, which commonly yields float64 for numbers.
-	var table amqplib.Table
-	err := yaml.Unmarshal([]byte("x-expires: 86400000\n"), &table)
-	assert.NoError(t, err)
-	val, ok := table["x-expires"]
-	assert.True(t, ok)
-	_, isFloat64 := val.(float64)
-	assert.True(t, isFloat64, "expected float64 via sigs.k8s.io/yaml, got %T (%v)", val, val)
-}
 
 func TestParseYamlTable(t *testing.T) {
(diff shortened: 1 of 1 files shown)


whynowy (approved):
thanks for the fix!

PR #3906, opened 2026-02-05
Initialize Prometheus Counters to 0 when service(s) start

aaronmell:
Per the https://prometheus.io/docs/instrumenting/writing_clientlibs/ counters and guages MUST start at 0

If counters and guages do not start at 0, the first time a counter or guage changes its value after a resart, prometheus will ignore the change. The causes functions like rate() and increase() to report the wrong values

Checklist:

### DIFF
--- a/USERS.md
+++ b/USERS.md
@@ -35,6 +35,7 @@ Organizations below are **officially** using Argo Events. Please send a PR with 
 1. [FikaWorks](https://fika.works/)
 1. [Kraken Technologies Ltd.](https://kraken.tech/)
 1. [Loam](https://www.getloam.com/)
+1. [Lumin Digital](https://lumindigital.com)
 1. [MariaDB](https://mariadb.com/)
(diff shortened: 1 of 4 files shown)


whynowy:
Is this true?

> f counters and guages do not start at 0, the first time a counter or guage changes its value after a resart, prometheus will ignore the change. The causes functions like rate() and increase() to report the wrong values

aaronmell:
> Is this true?
> 
> > f counters and guages do not start at 0, the first time a counter or guage changes its value after a resart, prometheus will ignore the change. The causes functions like rate() and increase() to report the wrong values

Yes, thats why I opened the pull request. Our alerting wasn't working right for some events we have that fire very infrequently. its mentioned in the docs for prometheus as well

aaronmell:
@whynowy Anything else you need for this?

whynowy:
I'm curious when this could happen, a container restart?

aaronmell:
Yes, anytime a pod restart happens

whynowy (approved):

PR #3983, opened 2026-04-06
feat: configurable consumer batch timeout + Kafka sensor latency optimizations (enable real-time processing)

### LINKED ISSUE #3984: Configurable consumer batch timeout for Kafka EventBus sensors

When using Kafka as the EventBus, the sensor consumer batches messages with a hardcoded 1-second timeout (`Batch(100, 1*time.Second, ...)`). This introduces significant latency between an event being published and a trigger firing — up to 1 second per internal topic hop (event → trigger → action), resulting in 3-5 seconds of total latency even for a single event at low volume.

In our environment (3-broker Strimzi cluster across 3 GCP zones, mTLS, `min.insync.replicas=2`), we observed ~4.0 seconds between `eventsource.publish` and `sensor.trigger` through distributed tracing. The batch timer accounted for ~2.5 seconds of that total, with the remaining time spent on Kafka transactional commits.

This latency is acceptable for high-throughput batch workloads but problematic for latency-sensitive use cases like webhook-driven workflows, real-time notifications, and interactive event-driven pipelines.

Add a configurable `consumerBatchMaxWait` field to the Kafka EventBus spec that controls the maximum time the sensor consumer waits to fill a batch before processing. The field should:

1. Accept a Go duration string (e.g., `1s`, `500ms`, `100ms`) or `0` to disable batching entirely
2. Default to `1s` (preserving current behavior when not set)
3. When set to `0`, process messages individually in real-time without batching
4. Be configurable at the EventBus level (default for all sensors) and overridable per Sensor via a `eventBusConsumerBatchMaxWait` field in the Sensor spec

Example EventBus configuration:

```yaml
kind: EventBus
metadata:
  name: default
spec:
  kafka:
    url: kafka:9092
    consumerBatchMaxWait: "100ms"
```

Example Sensor-level override:

```yaml
kind: Sensor
metadata:
  name: latency-sensitive-sensor
spec:
  eventBusConsumerBatchMaxWait: "0"
```

- **Hardcode a lower timeout (e.g., 100ms):** This would improve latency for everyone but reduces throughput optimization for high-volume use cases. A configurable approach lets users choose the right trade-off.
- **Remove batching entirely:** While this provides the best latency, it eliminates the transaction amortization benefit. Making it opt-in via `"0"` is safer.
- **Use JetStream instead of Kafka:** JetStream already processes messages one at a time (no batching), but switching event buses is not always feasible for teams that depend on Kafka's ecosystem, horizontal scaling, and exactly-once semantics.

**Additional context**

Benchmarks from a 3-broker Strimzi Kafka cluster (Kafka 4.1.1, 3 GCP zones, mTLS, `min.insync.replicas=2`, `transaction.state.log.replication.factor=3`):

| Configuration | Event-to-trigger latency |
|---|---|
| Default (`1s` batch) | ~3500ms |

Image

---

kaio6fellipe:
## Summary

Adds configurable batch timeout and latency optimizations to the Kafka EventBus sensor, reducing event-to-action latency from ~3.5s to <50ms.

**Batch timeout configuration:**
- **EventBus-level config** (`spec.kafka.consumerBatchMaxWait`): sets the default for all sensors using this EventBus
- **Sensor-level override** (`spec.eventBusConsumerBatchMaxWait`): overrides the EventBus default for a specific sensor
- **Default behavior preserved**: when not set, batches with a 1-second timeout (identical to current upstream behavior)
- **Real-time mode**: set to `"0"` to disable batching and process messages individually as they arrive

**Latency optimizations:**
- **Direct action invocation for single-dep triggers**: single-dependency triggers (OneAndDone) invoke the action callback directly instead of routing through the Kafka action topic, eliminating an entire Kafka transaction round-trip
- **Direct action invocation for satisfied multi-dep triggers**: when a multi-dependency trigger's dependencies are all satisfied, the action is invoked directly instead of routing through the action topic
- **Reduced sarama transaction retry backoff**: default retry backoff reduced from 100ms to 10ms (configurable via CRD), cutting per-transaction latency from ~500ms to <50ms when `CONCURRENT_TRANSACTIONS` retries occur
- **Consistent partition key routing**: trigger/action messages from the same source event are routed to the same partition using a composite key, reducing `AddPartitionsToTxn` calls from N to 1 per transaction
- **New CRD field `transactionRetryBackoff`**: operators can tune the transaction retry backoff per EventBus without rebuilding the image

Closes #3984

### Configuration examples

EventBus level (applies to all sensors):

```yaml
kind: EventBus
metadata:
  name: default
spec:
  kafka:
    url: kafka:9092
    consumerBatchMaxWait: "100ms"
    transactionRetryBackoff: "10ms"  # default; use "100ms" for cross-region
```

Sensor-level override:

```yaml
kind: Sensor
metadata:
  name: latency-sensitive-sensor
spec:
  eventBusConsumerBatchMaxWait: "0"  # real-time for this sensor only
```

### Why this matters

The Kafka sensor routes events through internal topics (event -> trigger -> action), with each hop going through `Batch(100, 1s)`. At low volume, every batch waits the full 1-second timeout before emitting a single message. Combined with Kafka transactional commit overhead (`WaitForAll` acks), each hop adds significant latency.

This PR addresses two independent latency sources:

1. **Batch wait latency (~1s per hop)**: eliminated by setting `consumerBatchMaxWait: "0"`, which processes messages individually via `consumeClaimRealtime` instead of the `Batch()` function
2. **Transaction overhead (~500ms per hop from `CONCURRENT_TRANSACTIONS` retries)**: eliminated by bypassing the action topic for direct action invocation and reducing the sarama retry backoff from 100ms to 10ms

| Configuration | Event-to-action latency | Use case |
|---|---|---|
| Default (`1s` batch) | ~3,500ms | High-throughput batch workloads |
| `0` (real-time, no batch) | **<50ms** | Webhook-driven, latency-sensitive workflows |

#### Default behavior

Image

#### Batch disabled

image

### Implementation details

**Batch timeout:**
- Added `ConsumerBatchMaxWait` field to `KafkaBus` struct (EventBus CRD)
- Added `EventBusConsumerBatchMaxWait` field to `SensorSpec` (Sensor CRD)
- Sensor-level value takes precedence over EventBus-level; both default to `1s`
- `ConsumeClaim` branches between `consumeClaimBatched` (uses `Batch()` with configured timeout) and `consumeClaimRealtime` (processes messages one at a time) based on the resolved duration

**Latency optimizations:**
- Single-dependency triggers (`OneAndDone()`) invoke `trigger.Action()` directly, returning the callback for post-commit execution — no messages produced to the action topic
- Multi-dependency triggers invoke `trigger.Action()` directly when `trigger.Update()` returns satisfied events, skipping the action topic hop
- Transaction retry backoff reduced from sarama's default 100ms to 10ms (`Producer.Transaction.Retry.Backoff`), with max retries set to 100
- Added `TransactionRetryBackoff` field to `KafkaBus` struct for operator-level tuning
- All trigger/action messages use a consistent partition key (`topic-partition-offset`) to route to the same partition, reducing `AddPartitionsToTxn` calls

**Topics:** While the action topic is still created and consumed, it is bypassed in the common flow for both single-dep and satisfied multi-dep triggers. The action topic remains as a fallback.

### Why this is safe

- **Backward compatible**: default behavior is unchanged when the fields are not set
- **No semantic changes**: transaction/checkpoint/offset logic is identical for both batched and real-time processing
- **`atLeastOnce` preserved**: direct action invocation executes before the offset is committed — if the sensor crashes, the message is reconsumed and the action fires again
- **Multi-dep crash safety**: if the sensor crashes after the trigger topic commit but before action execution, the trigger topic message is reconsumed and re-evaluated

## Test plan

- [x] Verified default behavior (no config set) matches current upstream (batches with 1s)
- [x] Verified `consumerBatchMaxWait: "0"` processes messages in real-time without batching
- [x] Verified sensor-level override takes precedence over EventBus-level
- [x] Verified `atLeastOnce` semantics preserved in both modes
- [x] Verified single-dep triggers invoke actions directly (no action topic messages, no Kafka transaction)
- [x] Verified event-to-action latency <50ms with `consumerBatchMaxWait: "0"` on single-node and multi-node Kafka
- [x] Verified `transactionRetryBackoff` CRD field is read and applied by the sensor
- [x] Verified distributed traces propagate correctly from eventsource -> sensor.trigger -> target service
- [ ] Test with multi-dependency (AND) triggers to verify trigger topic routing + direct action invocation after satisfaction
- [ ] Load test with high message volume to verify batched mode throughput is preserved


pkg/apis/events/v1alpha1/kafka_eventbus.go
```diff
@@ -23,4 +23,18 @@ type KafkaBus struct {
 	// Supported values: random, hash, roundrobin, manual. Defaults to random.
 	// +optional
 	Partitioner string `json:"partitioner,omitempty" protobuf:"bytes,7,opt,name=partitioner"`
+	// ConsumerBatchMaxWait sets the maximum time the Kafka sensor consumer
+	// will wait to fill a batch of messages before processing them.
+	// Accepts a Go duration string (e.g., "1s", "100ms", "0" to disable batching).
+	// When set to "0", messages are processed individually in real-time
+	// without batching. Defaults to "1s" if not specified.
+	// +optional
+	ConsumerBatchMaxWait string `json:"consumerBatchMaxWait,omitempty" protobuf:"bytes,8,opt,name=consumerBatchMaxWait"`
```
whynowy, line 32: Use `*metav1.Duration`

pkg/apis/events/v1alpha1/kafka_eventbus.go
```diff
@@ -23,4 +23,18 @@ type KafkaBus struct {
...
+	// TransactionRetryBackoff sets the backoff duration between Kafka transaction
+	// retry attempts when CONCURRENT_TRANSACTIONS errors occur. Lower values
+	// reduce latency at the cost of slightly higher CPU from more frequent retries.
+	// Accepts a Go duration string (e.g., "10ms", "100ms"). Defaults to "10ms".
+	// Use "100ms" for cross-region deployments with high coordinator latency.
+	// +optional
+	TransactionRetryBackoff string `json:"transactionRetryBackoff,omitempty" protobuf:"bytes,9,opt,name=transactionRetryBackoff"`
```
whynowy, line 39: Same here.

PR #3999, opened 2026-04-13
docs: enhance Kafka EventBus topics documentation

### LINKED ISSUE #3016: Enhance topics documentation for Kafka eventbus
Currently from description here https://argoproj.github.io/argo-events/eventbus/kafka/#topics it's hard to understand on why each topic is needed and how Argo Events will use it.

Currently EventBus based on Kafka looks like a black box. And in doer to build robust system we need more context.

Need more info from maintainers about:
1. How each topic `event`, `trigger`, `action` is used by Argo Events
2. If we have messages with different payload, should we create several EventBus or one EventBus is enough
3. Recommendations about what partitions we need to set for topics if we expect high amount of messages. Should we set number of partitions for `event`, `trigger`, `action` topics or only for `events`.

Also whatever you think will be useful to understand on how we can work with it.

---

alliasgher:
## Summary

Expand the Kafka EventBus topics documentation to explain **how** each topic is used internally, provide partitioning guidance, and document when to use single vs. multiple EventBus resources.

### New section: "How Each Topic Is Used"
- **Event topic** — shared channel where EventSources produce and Sensors consume CloudEvents, keyed by `{source}.{subject}`
- **Trigger topic** — internal Sensor topic for transaction coordination and exactly-once trigger evaluation
- **Action topic** — internal Sensor topic for recording completed actions and preventing duplicate execution

### Enhanced topic naming tables
- Added "used by" column showing which components interact with each topic

### New section: "Partitioning Recommendations"
- Event topic: partition count >= Sensor replicas for max parallelism
- Trigger/action topics: 1-3 partitions generally sufficient

### New section: "Single vs. Multiple EventBus"
- Single EventBus works for most cases (events keyed by source.subject)
- Separate EventBus when: isolating teams, different throughput/retention, different security configs

## Motivation

As described in #3016, the Kafka EventBus documentation described *what* topics exist but not *why* or *how* they are used, making it hard to build robust systems. The topic roles were derived from the source code (`pkg/eventbus/kafka/`).

Fixes #3016

### DIFF
--- a/docs/eventbus/kafka.md
+++ b/docs/eventbus/kafka.md
@@ -105,25 +105,83 @@ otherwise it is your responsibility to create these topics. If a topic does
 not exist and cannot be automatically created, the EventSource and/or Sensor
 will exit with an error.
 
-If you want to take advantage of the horizontal scaling enabled by the Kafka
-EventBus be sure to create topics with more than one partition.
+### How Each Topic Is Used
+
+Argo Events uses three Kafka topics per Sensor to implement reliable event
+delivery:
+
+- **Event topic** — The shared communication channel between EventSources and
+  Sensors. EventSources **produce** CloudEvents messages to this topic, keyed
+  by `{source}.{subject}`. Sensors **consume** from it using a consumer group
+  to receive the events that match their dependency definitions. All
+  EventSources and Sensors that share the same EventBus write to and read from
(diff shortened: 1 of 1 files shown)


whynowy (approved):
Thank you!

PR #4001, opened 2026-04-13
fix: make TLS optional for jetstreamExotic connections

### LINKED ISSUE #2938: jetstreamExotic requiring secure connection (no opt-out)
**Describe the bug**
Argo Events requires secure connection to use jetstreamExotic, but it shouldn't - not because secure connections shouldn't be used, but because there are cases when they're not required, like on dev environments (which is my case) and when running inside a service mesh that provides mTLS.

**To Reproduce**
1. Run a NATS server with JetStream enabled, but without any secure connections enabled
2. Spin up Argo Events
3. Create an EventBus pointing to the NATS server, using jetstreamExotic
4. Check the logs

**Expected behavior**
I should be able to connect the EventBus with the external NATS server without problems.

**Screenshots**
(None)

 - Kubernetes: v1.27.3
 - Argo: v3.5.1
 - Argo Events: latest+ec96ecd

**Additional context**
Here's where the secure connection is forced: https://github.com/argoproj/argo-events/blob/master/eventbus/jetstream/base/jetstream.go#L70

---

alliasgher:
## Summary

Make TLS optional for `jetstreamExotic` EventBus connections. Previously, TLS was always forced even when no TLS configuration was provided.

## Root Cause

In `pkg/eventbus/jetstream/base/jetstream.go`, the `MakeConnection()` function called `nats.Secure()` in **both** branches:

```go
if stream.tls != nil {
    // explicit TLS config — correct
    opts = append(opts, nats.Secure(tlsConfig))
} else {
    // no TLS config — still forces TLS with InsecureSkipVerify!
    opts = append(opts, nats.Secure(&tls.Config{InsecureSkipVerify: true}))
}
```

This means even when a user doesn't configure TLS at all, the NATS client attempts a TLS handshake, which fails against NATS servers that don't have TLS enabled.

## Fix

When `tls` is nil, simply don't add `nats.Secure()` to the connection options. This lets the NATS client connect using plain TCP, which is the expected behavior for:
- Development environments
- Service mesh setups (e.g., Istio) where mTLS is handled at the transport layer
- NATS servers that don't have TLS configured

## Changes

- `pkg/eventbus/jetstream/base/jetstream.go`: Remove forced `nats.Secure()` call when no TLS config is provided. Remove unused `crypto/tls` import.

Fixes #2938

### DIFF
--- a/docs/eventbus/jetstream.md
+++ b/docs/eventbus/jetstream.md
@@ -71,7 +71,9 @@ Jetstream has the concept of a Stream, and Subjects (i.e. topics) which are used
 
 ### Exotic
 
-To use an existing JetStream service, follow the example below.
+To use an existing JetStream service instead of having Argo Events manage one,
+use `jetstreamExotic`. This is useful when you already have a NATS JetStream
+cluster and want Argo Events to connect to it as a client.
 
 ```yaml
 apiVersion: argoproj.io/v1alpha1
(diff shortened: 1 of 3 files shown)


alliasgher:
Closing this PR — my analysis was incomplete.

The `InsecureSkipVerify` fallback I removed is **intentional** for the **managed** JetStream case (non-exotic):
- The managed JetStream installer (`pkg/reconciler/eventbus/installer/jetstream.go`) generates self-signed certs for the JetStream server but does NOT set the `BusConfig.JetStream.TLS` field.
- The managed JetStream pod runs with TLS enabled.
- When the client connects, the NATS server requires TLS upgrade. Without `InsecureSkipVerify`, the client refuses the self-signed cert.

The E2E test failure on this PR (`E2E Tests (jetstream)`) confirms this — `nats: tls error: tls: failed to verify certificate: x509: certificate signed by unknown authority`.

A proper fix for #2938 requires either:

1. **Adding an explicit opt-out field** to `JetStreamConfig` (e.g., `DisableTLS bool`) — requires API/protobuf regeneration. The exotic installer would set this when the user wants no TLS.
2. **Wiring the self-signed CA cert into the managed BusConfig** so the client uses proper TLS verification instead of `InsecureSkipVerify`. This is more correct architecturally but is a bigger change.

I'll leave it to the maintainers to decide on the preferred approach.

Issue #1372, opened 2021-10-11
Support GitHub app authentication for GitHub EventSource


joakimhew:
**Is your feature request related to a problem? Please describe.**
A common issue when setting up any sort of connection to GitHub organisations is having to use personal access tokens which might not make sense to the organisation. A workaround for this is to create an GitHub app for the organisation and instead use that to authenticate. This is something argocd already [supports](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#github-app-repositories)

Would it make sense to add support for this in the GitHub `EventSource`?

**Describe the solution you'd like**
I would imagine the `EventSource`, besides using `apiToken`, would allow for something like:

```yaml
githubApp:
  privateKey: 
     name: k8s-app-private-key
     key: secret
  installationId: <>
  appId: <>
```

**Describe alternatives you've considered**
Another alternative could be to support using ssh keys instead of API tokens which would allow us to use non personal ([deploy keys](https://docs.github.com/en/developers/overview/managing-deploy-keys)). This is also supported by [argocd](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#ssh-repositories)

**Additional context**
More on [GitHub apps](https://docs.github.com/en/developers/apps)

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

daniel-codefresh:
@whynowy Heres the related PR: #1438

PR #4018, opened 2026-04-21
feat: support GitHub App authentication for Sensor git triggers

### LINKED ISSUE #3904: Support GitHub app authentication for Sensors
Add GitHub Apps authentication support for Sensors  which use github as source, similar to the existing implementation for EventSources ([#1372](https://github.com/argoproj/argo-events/issues/1372))

We are using git source in sensor triggers to dispatch argo workflow like this and currently can use only token
```
- template:
     conditions: example
     name: example
     argoWorkflow:
       operation: submit
       source:
         git:
           filePath: workflows/example-workflow.yaml
           url: https://github.com/workflows-example-repo
           branch: main
           creds:
             username:
               name: example-secret
               key: username
             password:
               name: example-secret
               key: token

```

yugannkt:
Closes #3904

Checklist:

Adds githubApp field to GitArtifact in Sensor spec, enabling GitHub App authentication for git-based trigger sources. This mirrors the existing GitHub App support in EventSources (#1372).


whynowy (approved):
thanks!

PR #4028, opened 2026-04-25
docs: mention warn in LOG_LEVEL FAQ entry

alliasgher:
The warn log level itself landed in #4019, so this PR is now docs-only: it updates the FAQ's `LOG_LEVEL` note to mention `warn` and when it is useful (high-frequency triggers where `info` is too noisy).


whynowy:
already done in https://github.com/argoproj/argo-events/pull/4019?

alliasgher:
You're right, #4019 covers the code change. Rebased and trimmed this to the remaining FAQ docs update in 9367df3.

PR #4038, opened 2026-05-07
feat(sensor): allow cross-namespace EventBus reference via eventBusNamespace

### LINKED ISSUE #3048: Add cluster-wide EvetnBus
We are currently using a single Kafka as event bus. We have to define the same `default` `EventBus` resource in all our `Namespaces` where we are using ArgoEvents, with a `Secret` to store the credentials, which is not convenient.

I'd like to have a `ClusterEventBus` resource that I, as platform administrator, could deploy for all my users, that they could refer is their `EventSource` and `Sensor` resources.

I currently use a Kyverno ClusterPolicy to clone a reference `Secret` and `EventBus` resources in all `Namespaces`.

---

Capiru:
## What this PR does

Adds an optional `eventBusNamespace` field to `SensorSpec` so a **Sensor** can subscribe to an `EventBus` that lives in a **different namespace**. This lets platform teams centralise their `EventBus` (and its Secrets) in a shared namespace while user teams deploy `Sensor`s in their own namespaces — addressing the operational pain described in #3048.

## API change

```yaml
kind: Sensor
metadata:
  name: my-sensor
  namespace: app-ns          # Sensor lives here
spec:
  eventBusName: default
  eventBusNamespace: bus-ns  # EventBus lives here
  # … dependencies, triggers …
```

`eventBusNamespace` is optional; when omitted the behaviour is identical to today (lookup in the Sensor's own namespace).

## Bug fix bundled in this PR

`pkg/reconciler/sensor/resource.go` was hard-coding `EVENTBUS_SUBJECT=eventbus-<sensor.Namespace>`. When the EventBus is in a different namespace, publisher and subscriber would silently talk past each other. The subject now correctly derives from `eventBus.Namespace`. This is a **no-op for the same-namespace case**.

## Changes

| File | Reason |
|------|--------|
| `pkg/apis/events/v1alpha1/sensor_types.go` | New `EventBusNamespace` field (proto tag 9) + `GetEventBusNamespace` helper |
| `pkg/reconciler/sensor/controller.go` | Use helper for EventBus `Get` lookup |
| `pkg/reconciler/sensor/resource.go` | Derive `EVENTBUS_SUBJECT` from `eventBus.Namespace` |
| `pkg/webhook/validator/sensor.go` | Store full `ArgoprojV1alpha1Interface`; resolve EventBus in target namespace |
| `pkg/webhook/validator/validator.go` | Pass `aeClient` to `NewSensorValidator` |
| Tests (3 files) | Cross-namespace happy path + negative cases |
| `docs/eventbus/eventbus.md` | New "Cross-namespace EventBus reference" section |
| `docs/managed-namespace.md` | Cross-link + RBAC note |

## Limitations (documented, intentional, out of scope for this PR)

- **Secrets must be pre-replicated** — any Secret referenced by the EventBus (access token, TLS certs) must also exist in the Sensor's namespace. Tools like Kyverno can automate this; the controller does not copy Secrets across namespaces.
- **Namespace-scoped installs** — the cluster-scoped install already has the necessary RBAC. Namespace-scoped installs must include the EventBus namespace in `--managed-namespace`.
- **EventSource cross-namespace** — not implemented here; a follow-up PR can mirror this change for `EventSourceSpec`.
- **`ClusterEventBus`** — the broader ask from #3048; this PR does not introduce a new CRD.

## Testing

- All existing unit tests pass (`pkg/reconciler/sensor/...`, `pkg/webhook/validator/...`, `pkg/apis/...`).
- New cases:
  - Controller reconciles successfully when EventBus is in a different namespace.
  - Controller fails when `eventBusNamespace` points to a namespace without an EventBus (proves the field is honoured).
  - `EVENTBUS_SUBJECT` env var matches `eventBus.Namespace` for both same- and cross-namespace cases.
  - Webhook admits a Sensor whose `eventBusNamespace` points to an existing EventBus in another namespace.
  - Webhook denies when the EventBus is missing in the referenced namespace.

## Backward compatibility

Existing Sensor manifests (no `eventBusNamespace`) work unchanged. The `EVENTBUS_SUBJECT` value is unchanged for same-namespace deployments because the fallback resolves to `sensor.Namespace`, which equals `eventBus.Namespace` in that case.

Closes #3048 (partial — Sensor side)


whynowy:
1. Direct cross namespace access is not an acceptable solution;
2. It's not going to work with the implementation even if it's acceptable - the secrets across namespaces won't be abled to mounted as volumes.

thibaultserti:
Hello, I would also benefit from such feature.
What would be an acceptable implementation for this?

PR #4110, opened 2026-07-15
fix:spec.template.container.imagePullPolicy is ignored by Sensor controller

### LINKED ISSUE #4050: spec.template.container.imagePullPolicy is ignored by Sensor controller
**Describe the bug**
The imagePullPolicy specified under spec.template.container in a Sensor resource is ignored.

**To Reproduce**
Sensor YAML:
```
kind: Sensor
metadata:
  name: router-reboot-sensor
  namespace: argo
spec:
  eventBusName: default
  template:
    container:
      imagePullPolicy: IfNotPresent
```
the generated Sensor Pod is always created with:
```
apiVersion: v1
kind: Pod
metadata:
  labels:
    controller: sensor-controller
spec:
  containers:
  - name: main
    image: quay.io/argoproj/argo-events:v1.9.10
    imagePullPolicy: Always
  restartPolicy: Always
```
**Expected behavior**
The generated Pod should honor the imagePullPolicy defined in the Sensor template:

**Screenshots**

 - Kubernetes: [1.35.4]
 - Argo Events: [2.4.21]

DhaneshRaghavan:
# Summary
spec.template.container.imagePullPolicy is ignored by Sensor controller as reported in https://github.com/argoproj/argo-events/issues/4050

- Update `Container.ApplyToContainer()` to apply `ImagePullPolicy` when it is explicitly set in the Sensor template.
- Add a regression test to verify that the generated Deployment honors `spec.template.container.imagePullPolicy`.

fixes https://github.com/argoproj/argo-events/issues/4050

Signed-off-by: dhanesh.raghavan@outlook.com


DhaneshRaghavan:
@whynowy - please review this PR.

pkg/apis/events/v1alpha1/template.go
```diff
@@ -88,7 +88,7 @@ func (c *Container) ApplyToContainer(cc *corev1.Container) {
 	if cc.SecurityContext == nil {
 		cc.SecurityContext = c.SecurityContext
 	}
-	if cc.ImagePullPolicy == "" {
+	if c.ImagePullPolicy != "" {
```
whynowy, line 91: Shall we also update the logic for security context?

DhaneshRaghavan: @whynowy - with all due respect, I'm not seeing the reason why we should. Keeping it as it is seems to be good for me. I might be missing something, but welcome with more context.

PR #4111, opened 2026-07-15
Bump embedded JetStream encryption key to 32 bytes for FIPS compatibility

### LINKED ISSUE #4109: Embedded JetStream EventBus crashes under OpenSSL FIPS because generated encryption key is too short
**Describe the bug**

An embedded JetStream EventBus fails to start when the NATS server uses an OpenSSL FIPS provider that enforces the minimum HMAC key strength.

Argo Events [generates the JetStream encryption key using](https://github.com/argoproj/argo-events/blob/db2215e0db7b082e704c3fe58f3fa4a0ac92689a/pkg/reconciler/eventbus/installer/jetstream.go#L542):

```go
encryptionKey := sharedutil.RandomString(12)
```

Because the generated value is ASCII, this produces a 12-byte HMAC key. OpenSSL's FIPS provider requires an HMAC key of at least 112 bits. When NATS derives its JetStream encryption keys using HMAC-SHA256, the FIPS provider rejects the generated key and the server panics with `hmac_setkey:invalid key length`.

This prevents the Argo Events-managed JetStream EventBus from running with a FIPS-enabled NATS build.

**To Reproduce**

1. Deploy Argo Events v1.9.11.
2. Configure an embedded JetStream EventBus version using a NATS build backed by an OpenSSL FIPS provider. The configuration used here was:

   ```yaml
   configs:
     jetstream:
       versions:
         - version: 2.14.3
           natsImage: cgr.dev/nats-fips:2.14.3
           metricsExporterImage: cgr.dev/prometheus-nats-exporter-fips:0.20.1
           configReloaderImage: cgr.dev/nats-server-config-reloader-fips:0.23.0
           startCommand: /nats-server
   ```

3. Create an embedded JetStream EventBus:

   ```yaml
   kind: EventBus
   metadata:
     name: default
     namespace: argo-events
   spec:
     jetstream:
       version: 2.14.3
       replicas: 3
       persistence:
         volumeSize: 1Gi
   ```

4. Confirm that the controller-generated encryption key is 12 bytes:

   ```shell
   kubectl -n argo-events get secret eventbus-default-js-server \
     -o jsonpath='{.data.encryption}' | base64 -d | wc -c
   ```

   Output:

   ```text
   12
   ```

5. Inspect the `main` container logs for an EventBus pod.

   The NATS server begins starting JetStream and then crashes:

   ```text
   [INF] Starting JetStream
   [INF] Encryption: ChaCha20-Poly1305
   [INF] Starting JetStream cluster
   [INF] Creating JetStream metadata controller

   panic: EVP_MAC_init
       openssl error(s):
       error:1C800069:Provider routines:hmac_setkey:invalid key length:
       providers/implementations/macs/hmac_prov.c:173:

   vendor/github.com/microsoft/go-crypto-openssl/openssl.newHMAC3(...)
   vendor/github.com/microsoft/go-crypto-openssl/openssl.NewHMAC(...)
   crypto/hmac.New(...)
   github.com/nats-io/nats-server/v2/server.(*Server).jsKeyGen.func2(...)
   github.com/nats-io/nats-server/v2/server.(*fileStore).genEncryptionKeys(...)
   ```

**Expected behavior**

Argo Events should generate a JetStream encryption key that satisfies current FIPS HMAC key-strength requirements, allowing an embedded JetStream EventBus to start successfully with a FIPS-enabled NATS build.

NATS recommends an encryption key of at least 32 bytes, so changing the generated key length from 12 to 32 would resolve the immediate failure and follow the NATS recommendation:

```diff
- encryptionKey := sharedutil.RandomString(12)
+ encryptionKey := sharedutil.RandomString(32)
```

A regression test should also verify that newly generated JetStream encryption keys are at least 32 bytes.

**Screenshots**

Not applicable. Relevant container output is included above.

- Kubernetes: k3d / v1.35.6+k3s1
- Argo: Not applicable; Argo Workflows is not involved in the failure
- Argo Events: v1.9.11
- NATS Server: v2.14.3, FIPS build
- NATS image: `cgr.dev/nats-fips:2.14.3`
- Argo Events image: `cgr.dev/argo-events-fips:v1.9.11`
- Crypto backend: OpenSSL 3.x FIPS provider through `go-crypto-openssl`

**Additional context**

The key is generated here:

https://github.com/argoproj/argo-events/blob/v1.9.11/pkg/reconciler/eventbus/installer/jetstream.go#L540-L545

NATS uses the configured JetStream key as the HMAC-SHA256 key when deriving storage encryption keys:

https://github.com/nats-io/nats-server/blob/v2.14.3/server/jetstream.go#L240-L252

OpenSSL's FIPS provider requires at least 112 bits of HMAC key strength:

https://github.com/openssl/openssl/blob/openssl-3.4.0/providers/common/securitycheck.c#L78-L88

NATS recommends an encryption key of at least 32 bytes:

https://docs.nats.io/running-a-nats-service/configuration

The current Argo Events `master` branch still uses `RandomString(12)` at the time this issue was filed.

Existing EventBus secrets should not be silently rotated because existing JetStream data may be encrypted with the old key. The proposed generation change should apply to newly created secrets; migration or validation of existing short keys may need separate handling.

---

AruneshDwivedi:
## What / Why
The embedded JetStream EventBus encryption key was generated with `sharedutil.RandomString(12)`, yielding a 12-byte HMAC key. OpenSSL FIPS providers enforce a minimum HMAC key strength of 112 bits, so NATS panics with `hmac_setkey: invalid key length` when JetStream starts under a FIPS-enabled NATS build (e.g. `cgr.dev/nats-fips:2.14.3`), and the EventBus never comes up.

Closes #4109

## Fix
Generate a 32-byte key. 32 bytes exceeds the 112-bit FIPS minimum and matches NATS' own recommendation for the JetStream encryption key. Existing secrets are not rotated (their data stays encrypted with the old key).

## Verification
- Added a regression assertion in `TestJetStreamCreateObjects` (create auth secrets) that the server secret's encryption key is at least 32 bytes.
- `go test ./pkg/reconciler/eventbus/installer/ -run TestJetStreamCreateObjects` passes.

Signed-off-by: Kushal Dwivedi


whynowy:
Is it backward compatible?

pkg/reconciler/eventbus/installer/jetstream.go
```diff
@@ -545,7 +545,9 @@ func (r *jetStreamInstaller) createSecrets(ctx context.Context) error {
 
 	if !oldClientObjExisting || !oldServerObjExisting {
 		// Generate server-auth.conf file
-		encryptionKey := sharedutil.RandomString(12)
+		// 32 bytes satisfies NATS' recommended minimum and exceeds the
+		// 112-bit HMAC key strength required by OpenSSL FIPS providers.
+		encryptionKey := sharedutil.RandomString(32)
```
whynowy, line 550: This only happens for new create, it should be backward compatible.

whynowy (approved):

walsm232:
@AruneshDwivedi @whynowy thanks for getting this merged.

One follow-up from operators who already have persistent embedded JetStream EventBuses: create-only generation is the safe default (existing PVC data stays readable with the old key), but it also means existing buses are left on a 12-byte encryption key. Those buses still cannot start under a FIPS NATS build so today the only operator-supported way to move them onto FIPS images is to delete the `EventBus` (which uninstalls the PVCs) and recreate it I think, unless I am misunderstanding this.

NATS already documents an in-place master-key rotation path: restart once with the new key in `encryption_key` / `key` and the old key in `prev_encryption_key` / `prev_key`, let the server re-wrap the per-stream keys, then drop `prev_key` and restart again which looks like it would work. See:

- https://docs.nats.io/learn/security/encryption
- https://docs.nats.io/reference/config/jetstream/prev_encryption_key

I wonder if it would have been useful if the change had also driven that rotation. For example, generate a 32-byte key, inject the existing value as `prev_key` for one restart, then remove it, so existing EventBuses can move to FIPS NATS without deleting the bus and losing the JetStream store. Argo Events currently has no hook for this: `JS_KEY` looks to be hardcoded, there is no `prev_key` in the JetStream config template I don't think, and `createSecrets` returns immediately when both secrets already exist.

Happy to open a follow-up issue if that is the better place to track it (or potentially a PR for this).

PR #4114, opened 2026-07-16
fix: make JetStream TLS optional when unset

### LINKED ISSUE #2938: jetstreamExotic requiring secure connection (no opt-out)

amarkdotdev:
Fixes #2938

When `JetStreamConfig.TLS` was nil we still called `nats.Secure` with `InsecureSkipVerify`, so plain TCP NATS (dev / mesh-provided mTLS) could not connect.

If TLS is unset, connect over plain TCP. If you need TLS, set `tls` on the JetStream config (including `insecureSkipVerify` when you want that).

## Testing

- `go test ./pkg/eventbus/jetstream/base/ -count=1`


whynowy:
LGTM, have you tested to see if it's backward compatible for native EventBus?

amarkdotdev:
@whynowy Yes — native/managed EventBus stays TLS-compatible; unset TLS stays plain TCP.

**Native EventBus (spec.jetstream):** managed JetStream always serves TLS with a self-signed cert. This PR makes the installer advertise `tls.insecureSkipVerify` on `BusConfig`, so clients still connect with `nats.Secure` + skip-verify (same effective behavior as before, when Secure was forced unconditionally). Covered by unit test on Install → `Status.Config.JetStream.TLS.InsecureSkipVerify == true`.

**When TLS is unset** (exotic / plain TCP): client path no longer adds `nats.Secure`; connection is plain TCP as before this bug. Covered by `TestMakeConnectionTLSOptional` (nil TLS → no TLS handshake error wording) and exotic Install asserting `TLS == nil`.

Ran: `go test ./pkg/eventbus/jetstream/base/ ./pkg/reconciler/eventbus/installer/ -count=1`

whynowy (approved):
LGTM

PR #4121, opened 2026-07-23
fix: skip leader election for single-replica recreate-strategy event sources

### LINKED ISSUE #4112: Kafka EventSource with 1 Replica performs Leader Election and Crashloops
**Describe the bug**
When running an EventSource (e.g., using a Kafka EventBus) configured with a single replica (replicas: 1 or defaulted), the pod still executes Kubernetes leader election. Because the leader election timeouts in the codebase are extremely aggressive (5-second lease, 2-second renew deadline), any minor API server latency causes the leader election to fail. Because there is only 1 replica, the failure to renew the lease results in immediate pod crashes and continuous restart loops (CrashLoopBackOff), even though High Availability (HA) was never intended or configured. In our understanding this contradicts the official documentation: "HA can be achieved by setting spec.replicas to a number greater than 1... If following EventSource types have spec.replicas > 1, Active-Passive strategy is used..." [Argo Events HA Docs](https://argoproj.github.io/argo-events/eventsources/ha)

**Expected behavior**
For an EventSource with Replicas = 1 we would expect the leader election to be turned off.

 - Kubernetes: 1.29
 - Argo Events: v1.9.7

**Additional context**
In the "newElector" method the "clusterSize" variable is used for every other elector but the kubernetes elector:
https://github.com/argoproj/argo-events/blob/master/pkg/shared/leaderelection/leaderelection.go#L40C96-L40C96

It could also be possible to not call the newElector method if the number of replicas 1 one here: https://github.com/argoproj/argo-events/blob/master/pkg/eventsources/eventing.go#L441C5-L441C5

---

pujitha24:
Motivation:
EventSources that use a "recreate" deployment strategy (e.g. those backed
by a Kafka EventBus) always ran Kubernetes-based leader election in
EventSourceAdaptor.Start, regardless of spec.replicas. The k8s leader
election timeouts are intentionally aggressive (5s lease, 2s renew
deadline), so any transient API server latency can cause the single
replica to fail to renew its own lease. When that happens, the
OnStoppedLeading callback calls log.Fatalf("leader lost..."), which can
crash-loop the pod under such latency — even though with only one
replica there is no other instance to contend with and no HA scenario
is configured. Per the project's own HA docs, leader election is only
meant to matter when spec.replicas > 1.

This does not change behavior for replicas > 1 (still elects as before)
or for non-recreate-strategy event sources (already bypassed election).
It only removes the unnecessary, latency-sensitive election path when a
single replica is running.

Approach:
Extract the decision into a small helper,
needsLeaderElection(isRecreateType bool, replicas int) bool, and use it
in EventSourceAdaptor.Start to decide whether to go through
leaderelection.NewElector/RunOrDie at all. Since GetReplicas() floors at
1, this only changes behavior when replicas == 1 (the default), and the
event source now calls e.run directly, the same code path already used
by non-recreate types and by the eventual OnStartedLeading callback.

Validation:
  go build ./...
  go test ./pkg/eventsources/...
Both pass. Added a table-driven unit test, TestNeedsLeaderElection, in
pkg/eventsources/eventing_test.go covering isRecreateType x replicas
combinations.

Fixes #4112

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>


whynowy (approved):
LGTM, thanks

PR #3495, opened 2025-02-25
fix: Kafka consumer fails to reconnect and sensor stops working

smunukutla-mycarrier:
PR to fix #3481

PR #4148, opened 2026-08-17
fix: return from ConsumeClaim when the Kafka batch channel closes

### LINKED ISSUE #4147: Kafka eventbus sensor: ConsumeClaim spins forever when the claim's message channel closes
## Summary

On a Kafka broker rolling restart, a sensor's Kafka-eventbus `ConsumeClaim` can spin forever on 100% CPU instead of returning so sarama can rebalance the claim. The sensor pod stays `Ready` throughout, so nothing outside the process detects the stall — recovery currently requires manually deleting the pod.

## Root cause

`pkg/eventbus/kafka/base/utils.go`'s `Batch()` runs its work in a goroutine that does `defer close(out)` when its input channel (the claim's `Messages()`) closes:

https://github.com/argoproj/argo-events/blob/db6b682c/pkg/eventbus/kafka/base/utils.go#L16-L51

`pkg/eventbus/kafka/sensor/kafka_handler.go`'s `ConsumeClaim` reads that output channel without the comma-ok form:

https://github.com/argoproj/argo-events/blob/db6b682c/pkg/eventbus/kafka/sensor/kafka_handler.go#L146-L152

```go
case msgs := <-batch:
    if len(msgs) == 0 {
        h.Logger.Warn("Kafka batch contains no messages")
        continue
    }
```

Once `batch` is closed, every receive returns a nil slice immediately (the zero value, not a block), so this arm fires in a tight loop: log a warning and `continue`. The `case <-session.Context().Done()` arm never fires because the consumer group session itself is still alive — only the claim's message channel closed. Sarama is therefore never given the chance to rebalance the partition away, and `Listen` never reconnects.

## Observed impact

In our cluster, a single Kafka broker rolling restart closed the partition consumer channel for several sensor claims simultaneously. Each affected claim then produced roughly 500k `"Kafka batch contains no messages"` warning logs every 5 minutes and consumed a full CPU core, indefinitely. The pods reported `Ready` the entire time, so no health check caught it, and the log volume was large enough to matter for log-ingestion quotas. The only recovery was `kubectl delete pod`.

## Related PRs

#3495 adds `ErrIncompleteResponse` handling inside `Listen`, but that covers a different failure path (an error returned from the claim's message channel) — it does not touch `ConsumeClaim`'s read of the batched channel, and does not affect the scenario described here.

## Proposed fix

Read `batch` with the comma-ok form and return from `ConsumeClaim` as soon as it closes, exactly as any other Kafka consumer group handler is expected to when its claim is revoked:

```go
case msgs, ok := <-batch:
    if !ok {
        // the claim's message channel closed, so sarama can rebalance
        h.Logger.Info("Kafka batch channel closed, returning from ConsumeClaim")
        return nil
    }
    if len(msgs) == 0 {
        h.Logger.Warn("Kafka batch contains no messages")
        continue
    }
```

I have a fix + regression test ready and will open a PR referencing this issue.

TimmersThomas:
Fixes #4147

## Problem

On a Kafka broker rolling restart, `base.Batch()`'s internal goroutine closes its output channel when the claim's `Messages()` channel closes. `ConsumeClaim` read that channel without the comma-ok form, so a closed channel produced a nil slice on every receive instead of blocking. The handler logged `"Kafka batch contains no messages"` and looped forever, burning a full CPU core, without ever reaching the `session.Context().Done()` case — the session itself was still alive, only the claim had been revoked. Sarama never got the chance to rebalance the partition, and the sensor pod stayed `Ready` throughout, so nothing detected the stall.

See #4147 for the full root-cause writeup and links to the relevant source.

## Fix

Read the batched channel with comma-ok and return `nil` as soon as it closes, so sarama treats this exactly like any other consumer handler exit and reassigns the claim.

## Testing

Added `pkg/eventbus/kafka/sensor/kafka_handler_test.go` with hand-rolled `sarama.ConsumerGroupSession` / `sarama.ConsumerGroupClaim` fakes (this package had no existing test file or sarama fakes to reuse). The test closes the claim's message channel while the session context stays alive, and asserts `ConsumeClaim` returns within a bounded timeout. It fails (times out) against the old code and passes against the fix.

```
go test -race -count=1 ./pkg/eventbus/kafka/sensor/...
go build ./...
```

Both pass locally.

#3495 is related but covers a different path (`ErrIncompleteResponse` inside `Listen`) — it does not touch `ConsumeClaim`'s read of the batched channel and does not affect the scenario here.

Checklist:

### DIFF
--- a/pkg/eventbus/kafka/sensor/kafka_handler.go
+++ b/pkg/eventbus/kafka/sensor/kafka_handler.go
@@ -145,7 +145,12 @@ func (h *KafkaHandler) ConsumeClaim(session sarama.ConsumerGroupSession, claim s
 
 	for {
 		select {
-		case msgs := <-batch:
+		case msgs, ok := <-batch:
+			if !ok {
+				// the claim's message channel closed, so sarama can rebalance
+				h.Logger.Info("Kafka batch channel closed, returning from ConsumeClaim")
+				return nil
+			}
 			if len(msgs) == 0 {
 				h.Logger.Warn("Kafka batch contains no messages")
 				continue
--- a/pkg/eventbus/kafka/sensor/kafka_handler_test.go
+++ b/pkg/eventbus/kafka/sensor/kafka_handler_test.go
@@ -1,0 +1,90 @@
+package kafka
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/IBM/sarama"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+)
+
+// fakeConsumerGroupSession is a minimal sarama.ConsumerGroupSession whose
+// Context() stays live for the lifetime of the test, mirroring a session
+// that has not been rebalanced away.
+type fakeConsumerGroupSession struct {
+	ctx context.Context
+}
+
+func (f *fakeConsumerGroupSession) Claims() map[string][]int32                  { return nil }
+func (f *fakeConsumerGroupSession) MemberID() string                            { return "fake-member" }
+func (f *fakeConsumerGroupSession) GenerationID() int32                         { return 1 }
+func (f *fakeConsumerGroupSession) MarkOffset(string, int32, int64, string)     {}
+func (f *fakeConsumerGroupSession) Commit()                                     {}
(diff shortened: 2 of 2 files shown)


whynowy (approved):
LGTM

PR #4158, opened 2026-08-24
fix(eventbus): close NATS connection when trigger connection init fails

### LINKED ISSUE #4106: Sensor leaks NATS connections when JetStream K/V initialization fails
## Describe the bug

`SensorJetstream.Connect` opens a NATS connection and passes it to `NewJetstreamTriggerConn`. If trigger initialization fails—for example, `JSContext.KeyValue(sensorName)` reports a timeout or an offline stream—`Connect` returns the error without closing the newly opened connection.

The Sensor reconnect daemon retries every five seconds, leaving another NATS client after each failed attempt. In a production incident the EventBus reached 65,536/65,536 connections; 65,263 were abandoned request clients whose subscriptions were `_INBOX...`. Sensors stopped consuming webhook events while Kubernetes readiness remained green.

This is related to #2376, which covers the same failed K/V path but addresses the typed-nil panic rather than connection cleanup.

## Affected versions

Confirmed in v1.9.9 through v1.9.11. The relevant path is unchanged on `master` as of 2026-07-13.

## Source

- v1.9.11 opens the connection and directly returns constructor failure: https://github.com/argoproj/argo-events/blob/v1.9.11/pkg/eventbus/jetstream/sensor/sensor_jetstream.go#L74-L80
- K/V lookup can fail after the connection opens: https://github.com/argoproj/argo-events/blob/v1.9.11/pkg/eventbus/jetstream/sensor/trigger_conn.go#L68-L76
- The reconnect daemon repeats every five seconds: https://github.com/argoproj/argo-events/blob/v1.9.11/pkg/sensors/listener.go#L318-L334

## Expected behavior

Every NATS connection created by `SensorJetstream.Connect` is either returned inside a successful `TriggerConnection` or closed before an error is returned.

## Suggested regression test

Start an embedded JetStream server, construct a `SensorJetstream` for a missing K/V bucket, record `server.NumClients()`, and call `Connect` repeatedly without `Initialize`. Each call should return the K/V error and a nil trigger connection. Assert eventually that `server.NumClients()` returns to the baseline. Before the fix it increases once per call.

## Minimal fix

```go
triggerConn, err := NewJetstreamTriggerConn(conn, stream.sensorName, triggerName, dependencyExpression, deps)
if err != nil {
    _ = conn.Close()
    return nil, err
}
return triggerConn, nil
```

This keeps ownership explicit: `Connect` owns `conn` until construction succeeds; afterward the returned `TriggerConnection` owns it.

## Operational impact

During prolonged K/V or JetStream failure, each Sensor trigger leaks one connection per five-second retry. Raising NATS `max_connections` only delays exhaustion.

HarnageaGabriel:
## Summary
- `SensorJetstream.Connect` opened a NATS connection via `MakeConnection()` and returned `NewJetstreamTriggerConn`'s error directly, without closing the connection on failure.
- Since the Sensor reconnect daemon retries every 5s (`pkg/sensors/listener.go`), a persistent K/V/JetStream init failure leaked one NATS connection per attempt — a production incident hit 65,536/65,536 connections exhausted.
- Fix closes `conn` on the failure path after `MakeConnection` succeeds; the returned `TriggerConnection` takes ownership only on success.

## Test plan
- [x] Added `TestSensorJetstreamConnectClosesConnectionOnTriggerConnectionError` (embedded JetStream server, missing K/V bucket) asserting `Connect` returns an error and the server's client count returns to baseline.
- [x] `go build ./...`
- [x] `go test ./pkg/eventbus/... -run . -v`
- [x] `go vet ./pkg/eventbus/jetstream/sensor/...`

Fixes #4106

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,6 +63,7 @@ require (
 	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/nats-io/graft v0.0.0-20220215174245-93d18541496f
+	github.com/nats-io/nats-server/v2 v2.11.15
 	github.com/nats-io/nats.go v1.53.1
 	github.com/nats-io/stan.go v0.10.4
 	github.com/nsqio/go-nsq v1.1.0
@@ -142,6 +143,7 @@ require (
 	github.com/alibabacloud-go/tea v1.2.2 // indirect
 	github.com/aliyun/credentials-go v1.3.10 // indirect
 	github.com/andybalholm/brotli v1.1.0 // indirect
+	github.com/antithesishq/antithesis-sdk-go v0.6.0-default-no-op // indirect
 	github.com/ardielle/ardielle-go v1.5.2 // indirect
 	github.com/awalterschulze/gographviz v0.0.0-20200901124122-0eecad45bd71 // indirect
 	github.com/aws/aws-sdk-go-v2 v1.36.3 // indirect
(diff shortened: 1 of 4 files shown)


whynowy (approved):
Thanks!

PR #4174, opened 2026-09-05
fix(sensors): treat an exactly elapsed interval as loggable in the log trigger

### LINKED ISSUE #4173: Log trigger uses a strict time comparison: interval 0 can skip events, and TestLogTrigger fails on Windows
**Describe the bug**

The log trigger decides whether to log with a strict comparison:

```go
func (t *LogTrigger) shouldLog(log *v1alpha1.LogTrigger) bool {
	return time.Now().After(t.LastLogTime.Add(log.GetInterval()))
}
```

https://github.com/argoproj/argo-events/blob/f2ca952/pkg/sensors/triggers/log/log.go#L56-L58

With `intervalSeconds` unset (0, which the docs describe as "log every event"), `shouldLog` returns `false` whenever `time.Now()` has not moved past `LastLogTime`, so the event is dropped without any log line. The same off-by-one applies when exactly `intervalSeconds` have elapsed: the interval is treated as not yet over.

How often that happens depends on the resolution of Go's monotonic clock. On Linux it is nanoseconds, so two consecutive reads are almost never equal and the bug stays hidden. On Windows the monotonic clock advances in coarse ticks, so two consecutive `time.Now()` calls are routinely equal and the existing `TestLogTrigger` fails deterministically at `log_test.go:23`.

**To Reproduce**

1. Check out `master` on a Windows machine and run the package tests (screenshot 1):

   ```
   go test -count=1 -run TestLogTrigger ./pkg/sensors/triggers/log/
   ```

   ```
   --- FAIL: TestLogTrigger (1.00s)
       log_test.go:23:
           Error Trace:    pkg/sensors/triggers/log/log_test.go:23
           Error:          Should be true
   FAIL
   ```

   Line 23 is `assert.True(t, l.shouldLog(&sv1.LogTrigger{}))`, i.e. interval 0 right after `Execute` set `LastLogTime`.

2. The clock behaviour behind it, measured with a small program on the same machine (screenshot 2, two runs):

   ```go
   same := 0
   for i := 0; i < 100000; i++ {
       last := time.Now()
       if !time.Now().After(last) {
           same++
       }
   }
   fmt.Printf("time.Now().After(previous time.Now()) was false %d / 100000 times\n", same)
   ```

   ```
   go go1.27.0 windows/amd64
   time.Now().After(previous time.Now()) was false 99999 / 100000 times
   ```

3. The same test passes on Linux (WSL2 on the same machine, go1.27.0 linux/amd64, 3/3 runs), which is why CI does not catch it (screenshot 3).

**Expected behavior**

An interval of 0 should log every event, and an interval that has exactly elapsed should count as elapsed. `shouldLog` should compare with `!Before(...)` (greater than or equal) instead of `After(...)`. `TestLogTrigger` should pass on every platform Go supports.

**Screenshots**

Screenshot 1: `master` on Windows, `TestLogTrigger` fails at `log_test.go:23`.

TestLogTrigger failing on Windows on master

Screenshot 2: monotonic clock resolution on the same Windows machine, 99999 of 100000 consecutive `time.Now()` pairs are equal (two runs).

time.Now() resolution measurement on Windows

Screenshot 3: the same test on `master` under Linux (WSL2), passes 3/3.

TestLogTrigger passing on Linux

 - Kubernetes: n/a (unit test)
 - Argo: n/a
 - Argo Events: master at f2ca952
 - Go: go1.27.0 windows/amd64 (fails), go1.27.0 linux/amd64 (passes)

**Additional context**

A one line change plus a clock independent test (set `LastLogTime` to `now` and assert interval 0 logs while interval 1s does not) fixes both cases. Happy to open a PR.

---

PhuocOng:
Fixes #4173.

`LogTrigger.shouldLog` uses `time.Now().After(LastLogTime.Add(interval))`, a strict comparison. When the interval is 0 (the default, meaning "log every event") and the clock has not advanced since the previous event, it returns `false` and the event is silently not logged. The same off-by-one applies when exactly `intervalSeconds` have elapsed.

On Linux this only shows up when two events land in the same nanosecond, which is why CI never sees it. On Windows, Go's monotonic clock advances in coarse ticks (a tight loop of 100000 `time.Now()` pairs on my machine yielded 0 pairs where the second read was `After` the first), so the existing `TestLogTrigger` fails deterministically at `log_test.go:23`:

```
--- FAIL: TestLogTrigger (1.00s)
    log_test.go:23: Should be true
```

The fix compares with `!Before(...)`, so an interval that has exactly elapsed counts as elapsed. `TestShouldLogSameInstant` pins the behavior without depending on the clock: `LastLogTime = now` must log with interval 0 and not log with interval 1s, and `LastLogTime = now - 1s` must log with interval 1s. It fails before the change and passes after; the existing `TestLogTrigger` now passes on Windows as well (3/3 runs).


whynowy (approved):
LGTM

PR #4177, opened 2026-09-07
docs: mark NATS Streaming (STAN) eventbus as deprecated

### LINKED ISSUE #4165: Remove EOL NATS Streaming (STAN) configuration from controller-config manifests
## Summary

The controller-config in the install manifests still ships the `nats` (NATS Streaming / STAN) EventBus version matrix and `natsStreamingImage` configuration. NATS Streaming reached end-of-life in June 2023, and the prometheus-nats-exporter removed STAN support entirely (nats-io/prometheus-nats-exporter#329), so newer exporter images are incompatible with the configuration the manifests still advertise.

## Proposal

Remove (or clearly mark as unsupported) the NATS Streaming sections from `manifests/base/controller-manager/controller-config.yaml` and related docs, leaving JetStream (and Kafka) as the supported EventBus backends. If outright removal is considered breaking for existing STAN EventBus users, a deprecation notice in the config plus removal in the next minor would work too.

## Context

Downstream, the argo-helm chart mirrors these manifests verbatim, and users there are asking for the deprecated STAN configuration to go away (argoproj/argo-helm#3303). The chart follows upstream, so the change needs to happen here first.

Bryandero98:
## Summary
Fixes #4165 — the docs still described NATS Streaming (STAN) as an active or future-tense deprecation, even though STAN reached end-of-life in June 2023 and `prometheus-nats-exporter` has since dropped STAN support entirely.

## Changes
- `docs/eventbus/stan.md`: added an explicit deprecation notice pointing to JetStream.
- `docs/eventbus/jetstream.md`: fixed "NATS Streaming (which will eventually be deprecated)" — STAN already reached EOL, this was stale future-tense wording.

## Why the manifest/config changes were reverted
An earlier version of this PR also removed the `eventBus.nats` block from `controller-config.yaml` (+ generated manifests) and the corresponding `ValidateConfig` check, on the assumption that no test exercised that path. CI caught that this assumption was wrong: the `E2E Tests (stan)` job creates a live STAN EventBus and waits for it to become ready, which requires `GetSTANVersion()` to resolve a version from `eventBus.nats`. Removing it broke STAN EventBus creation entirely, and the job timed out waiting for the EventBus.

The original issue itself anticipated this: "If outright removal is considered breaking for existing STAN EventBus users, a deprecation notice in the config plus removal in the next minor would work too." Since CI still exercises full STAN EventBus creation, hard removal is exactly that breaking change — so this PR now only touches docs, and the manifests/`config.go` are back to matching `upstream/master` exactly.

## Verification
- `go build ./pkg/reconciler/...` and `go test ./pkg/reconciler/...` pass.
- `git diff upstream/master -- manifests/ pkg/reconciler/config.go` is empty.
- CI: `E2E Tests (stan)` (the job that previously failed) re-run against this revision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Bryandero98:
Update: the initial version of this PR broke controller-manager startup entirely - CI caught it (all 3 E2E variants failed identically at the controller-manager rollout step, before any eventbus-specific logic even ran).

Root cause: `ValidateConfig` (`pkg/reconciler/config.go`) still required `supportedSTANVersions()` to be non-empty and calls `logger.Fatalw` on failure, so removing the `eventBus.nats` block made the controller crash-loop at startup on every install. Pushed a follow-up commit removing that check too - `GetSTANVersion` (the actual STAN installer lookup) already handles a nil/empty NATS config gracefully via its own "not found" error, which is what matters now that STAN is EOL.

Verified locally: `go build ./pkg/reconciler/...` and `go test ./pkg/reconciler/...` both clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Bryandero98:
Follow-up: the previous revision of this PR also removed the \`eventBus.nats\` block from the controller-config manifests and the corresponding \`ValidateConfig\` check. CI caught that this was a real regression — \`E2E Tests (stan)\` creates a live STAN EventBus and timed out waiting for it to become ready, since \`GetSTANVersion()\` could no longer resolve a version.

The original issue already suggested a softer alternative for exactly this case ("a deprecation notice in the config plus removal in the next minor"), so I've reverted the manifest/\`config.go\` changes back to match \`upstream/master\` exactly and kept only the docs deprecation notices (\`docs/eventbus/stan.md\`, \`docs/eventbus/jetstream.md\`). PR title/description updated accordingly.

whynowy (approved):
thanks

